PR #1, opened 2020-10-14
Experiment with plugins and theme bundled together.

iandunn:
Rough idea where the root of this repo corresponds to the `wp-content/plugins/pattern-directory` folder on production.

What pros/cons do you see? Do you want to stick with this approach, or go something else?

### DIFF
--- a/admin/admin-bootstrap.php
+++ b/admin/admin-bootstrap.php
@@ -1,0 +1,3 @@
+<?php
+
+namespace WordPressdotorg\Pattern_Directory\Admin;
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -1,0 +1,15 @@
+<?php
+/**
+ * Plugin Name: Pattern Directory
+ * Description: Creates a directory to manage block patterns.
+ * Plugin URI:  https://wordpress.org/patterns/
+ * Text Domain: wporg-plugins
+ */
+
+namespace WordPressdotorg\Pattern_Directory;
+
+require_once __DIR__ . '/admin/admin-bootstrap.php';
+//require_once __DIR__ . '/creator/creator-bootstrap.php';
+
+// Bundle the theme with the plugin, because the loose coupling in past directory projects has been frustrating.
+register_theme_directory( __DIR__ . '/themes' );
--- a/themes/pattern-directory/functions.php
+++ b/themes/pattern-directory/functions.php
@@ -1,0 +1,3 @@
+<?php
+
+namespace WordPressdotorg\Pattern_Directory\Theme;
--- a/themes/pattern-directory/index.php
+++ b/themes/pattern-directory/index.php
@@ -1,0 +1,1 @@
+<p>I'm a little theme stub, short and stout.</p>
--- a/themes/pattern-directory/style.css
+++ b/themes/pattern-directory/style.css
@@ -1,0 +1,10 @@
+@charset "UTF-8";
+
+/*
+ * Theme Name:  Pattern Directory
+ * Description: Displays block patterns to w.org visitors.
+ * Theme URI:   https://wordpress.org/patterns/
+ * Version:     ♾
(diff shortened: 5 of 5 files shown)


iandunn:
### Pros

* The theme and plugin(s) are more tightly coupled. In the past there've been some inconsistencies because the plugin directory plugin and theme were written by different people and weren't coordinated enough.

### Cons

* `register_theme_directory()` expects a folder of themes, not an individual theme folder. so we'll always have an additional unnecessary layer in the hierarchy.
* can't have `mu-plugins`
* `.wp-env` stuff, `/bin` folder, etc would live inside `plugins/pattern-directory` instead of root

I lean towards a more traditional approach, plus more intentional collaboration. 

I also think some of the issues w/ the plugin directory can be avoided by using a lean/YAGNI approach. e.g., instead of building a widget in the plugin, only for the theme to hardcode it into a template file, the plugin could do nothing, and the theme could just hardcode the content.

ryelle:
I think I'm missing a little context (even after reading the slack thread, maybe from your video call?), but I have to say I really liked [the learn setup](https://github.com/WordPress/learn/) — especially with composer/the create step installing required dependencies, and the other wp-env helpers. It made it very easy to jump back in after being away from the codebase & review a PR.

Is the idea that this repo would be deployed as one unit into `wporg/public_html/wp-content/plugins/pattern-directory`, and the theme on production would be loaded from `wporg/public_html/wp-content/plugins/pattern-directory/themes/pattern-directory`? That seems pretty unexpected (and might have issues if we're going to be child-themeing off wporg, especially with the current Grunt process).

I'm not totally sure what this different setup is meant to solve.

dd32:
> ### Cons
> * `register_theme_directory()` expects a folder of themes, not an individual theme folder. so we'll always have an additional unnecessary layer in the hierarchy.

To combat that, we can simply call `register_theme_directory( __DIR__ /* wp-content/plugins/pattern-directory/ */ )` and have the theme in `/theme`, WordPress will only scan all of the subfolders when it needs to find the list of themes (updates and when viewing the theme screen) so it won't cause any negative performance issues I don't think. That removes the need for the double-directory hierarchy.

> * can't have `mu-plugins`
> * `.wp-env` stuff, `/bin` folder, etc would live inside `plugins/pattern-directory` instead of root

I'm not sure those are actual cons here?
`mu-plugins` should not be needed, as all of the pattern directory should be loaded anyway, so forcing some to be enabled seems like it's fixing a problem that doesn't need solving.
`.wp-env` and `bin` are extra "noise" but the bin files would likely be needed in production anyway, and dotfiles are not too significant.

tellyworth:
Stepping back, here's the gist of our earlier discussion for @ryelle's benefit (and perhaps ours, when we come back to this later and can't remember why we did things this way :) -

Goals:
* A repo and environment where we can develop the entirety of the pattern directory: the back-end, API, creator tool, and theme all in one place.
* It should be easy to run and test locally.
* Also easy to test on a sandbox and deploy to staging/production.

We decided not to start with a fork of the plugin directory back-end. Instead, we'll start with a clean slate and move in just the pieces that we need.

dd32:
After reviewing it further, if the https://github.com/WordPress/five-for-the-future & https://github.com/WordPress/learn setups worked well for other folks, let's just go with that as the standard so everything not in the monorepo works in a similar way.

One of the reasons I suggested this route originally, was because of the problems we'd had in the past with the plugin and theme code living in completely separate paths, often resulting in two commits to meta to make the changes. It's not as bad when it's done in GitHub like this unlike with the Plugin Directory where the canonical source was the monorepo and it was hard to work on locally.

iandunn:
That sounds good to me 👍🏻

iandunn:
89d1b96 forks the Learn setup and adds some tweaks and stubs. I didn't test the Docker stuff since I don't use it. Feel free to adjust to fit your needs.

PR #5, opened 2020-10-27
Add basic Block Pattern Creator UI

ryelle:
This pulls in the plugin from https://github.com/ryelle/block-pattern-creator

I created a very basic CPT in the pattern-directory plugin, since I'm assuming that will be the source of that CPT. Happy to change anything about that (like the slug — realizing as I type this that core might want to use `wp-pattern` someday).

**To try it out:**

- Install dependencies in project root with `yarn`
- Build all the sub-projects: `yarn workspaces run build`
- Create some Block Patterns in wp-admin, `/wp-admin/post-new.php?post_type=wp-pattern`.
- View a published pattern, if you're logged in you should see the pattern editor.
- Play around & save your pattern, it should save.

⚠️  None of the sidebar controls are hooked up yet, they're all placeholders.

**Screenshot**

Screen Shot 2020-10-27 at 6 40 05 PM

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
+++ b/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
@@ -61,9 +61,11 @@ function enqueue_assets() {
 	wp_add_inline_script(
 		'wporg-pattern-creator-script',
 		sprintf(
-			'const wporgBlockPattern = {"settings": %1$s, "postId": %2$s}',
-			wp_json_encode( $settings ),
-			wp_json_encode( get_the_ID() )
+			'var wporgBlockPattern = JSON.parse( decodeURIComponent( \'%s\' ) );',
+			rawurlencode( wp_json_encode( array(
+				'settings' => $settings,
+				'postId'   => get_the_ID(),
+			) ) )
 		),
 		'before'
 	);
--- a/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
+++ b/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
@@ -1,0 +1,101 @@
+<?php
+/**
+ * Plugin Name: Block Pattern Creator
+ * Description: Create block patterns on the frontend of a site.
+ * Version: 1.0.0
+ * Requires at least: 5.5
+ * Author: WordPress Meta Team
+ * Text Domain: wporg-pattern-creator
+ * License: GPL v2 or later
+ * License URI: http://www.gnu.org/licenses/gpl-2.0.txt
+ */
+
+namespace WordPressdotorg\Pattern_Creator;
+
+/**
+ * Check the conditions of the page to determine if the editor should load.
+ *
+ * @return boolean
+ */
+function should_load_creator() {
+	return (
+		function_exists( 'gutenberg_experimental_global_styles_get_merged_origins' ) &&
+		\is_singular( 'wp-pattern' )
+	);
+}
+
+/**
+ * Register & load the assets.
+ *
+ * @throws \Error If the build files don't exist.
+ */
+function enqueue_assets() {
+	if ( ! should_load_creator() ) {
+		return;
+	}
+
+	$dir = dirname( __FILE__ );
+
+	$script_asset_path = "$dir/build/index.asset.php";
+	if ( ! file_exists( $script_asset_path ) ) {
+		throw new \Error( 'You need to run `yarn start` or `yarn build` for the Pattern Creator.' );
+	}
+
+	$script_asset = require( $script_asset_path );
+	wp_register_script(
+		'wporg-pattern-creator-script',
+		plugins_url( 'build/index.js', __FILE__ ),
+		$script_asset['dependencies'],
+		$script_asset['version'],
+		true
+	);
+
+	wp_set_script_translations( 'wporg-pattern-creator-script', 'wporg-pattern-creator' );
+
+	$merged = gutenberg_experimental_global_styles_get_merged_origins();
+	$settings = array(
+		'isRTL' => is_rtl(),
+		'__experimentalFeatures' => gutenberg_experimental_global_styles_get_editor_settings( $merged ),
+		'__experimentalGlobalStylesBaseStyles' => gutenberg_experimental_global_styles_get_core(),
+	);
+	wp_add_inline_script(
+		'wporg-pattern-creator-script',
+		sprintf(
+			'const wporgBlockPattern = {"settings": %1$s, "postId": %2$s}',
+			wp_json_encode( $settings ),
+			wp_json_encode( get_the_ID() )
+		),
+		'before'
+	);
+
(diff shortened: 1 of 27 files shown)


iandunn:
> core might want to use wp-pattern someday

I agree. #6 uses `wporg-pattern`.

public_html/wp-content/plugins/pattern-creator/src/style.css
```diff
@@ -0,0 +1,57 @@
+:root {
+	--default-font: sans-serif;
```
iandunn, line 2: Is there something that'll process this for IE11, or are you thinking we wouldn't support that?

public_html/wp-content/plugins/pattern-creator/pattern-creator.php
```diff
@@ -0,0 +1,101 @@
...
+		true
+	);
+
+	wp_set_script_translations( 'wporg-pattern-creator-script', 'wporg-pattern-creator' );
+
+	$merged = gutenberg_experimental_global_styles_get_merged_origins();
+	$settings = array(
+		'isRTL' => is_rtl(),
+		'__experimentalFeatures' => gutenberg_experimental_global_styles_get_editor_settings( $merged ),
+		'__experimentalGlobalStylesBaseStyles' => gutenberg_experimental_global_styles_get_core(),
```
iandunn, line 59: It might be safer to have `function_exists()` checks here too (or in `should_load_creator()`), to avoid fatals on prod when those functions are renamed.

public_html/wp-content/plugins/pattern-creator/pattern-creator.php
```diff
@@ -0,0 +1,101 @@
...
+	);
+	wp_add_inline_script(
+		'wporg-pattern-creator-script',
+		sprintf(
+			'const wporgBlockPattern = {"settings": %1$s, "postId": %2$s}',
+			wp_json_encode( $settings ),
+			wp_json_encode( get_the_ID() )
+		),
```
iandunn, line 69: I think something like this would be safer:

```php
sprintf(
	'var WordCampBlocks = JSON.parse( decodeURIComponent( \'%s\' ) );',
	rawurlencode( wp_json_encode( $data ) )
)
```

xref https://core.trac.wordpress.org/ticket/51159

iandunn:
I'm getting this when trying to run `start` or `build`:

```sh
> npm run build

> wporg-pattern-creator@1.0.0 build /Users/iandunn/vhosts/localhost/wporg-pattern.test/public_html/wp-content/plugins/pattern-creator
> wp-scripts build

/Users/iandunn/vhosts/localhost/wporg-pattern.test/public_html/wp-content/plugins/pattern-creator/node_modules/ignore-emit-webpack-plugin/index.js:64
                compilation.hooks.processAssets.tap({
                                                ^

TypeError: Cannot read property 'tap' of undefined
```

The common solutions seem like they'd need to be done in `wp-scripts`. It could be that the problem is there, or maybe  some undocumented dependencies related to  `wp-env` (which I'm not using)?

I'll take a look at the code itself, though.

public_html/wp-content/plugins/pattern-creator/src/store/actions.js
```diff
@@ -0,0 +1,23 @@
...
+}
+
+/**
+ * Save a block pattern.
+ */
+export function* saveBlockPattern() {
+	const patternId = yield select( MODULE_KEY, 'getEditingBlockPatternId' );
+
+	// @todo maybe check for errors?
+	yield dispatch( 'core', 'saveEditedEntityRecord', KIND, POST_TYPE, patternId );
```
iandunn, line 22: What do you think about using `async` / `await` instead of generator functions? That seems a lot simpler to me.

ryelle:
Ah, I should've added build steps to the testing instructions, sorry about that — in the root of the repo, run `yarn` to install dependencies (wp-scripts), then `yarn workspaces run build` to build everything (though right now that's just this project).

I _think_ the issue you're seeing it that you have an older version of wp-scripts globally installed, and it's incorrectly being picked up here.

iandunn:
Ah, you're right, thanks!

I ran `npm install` and `yarn workspaces run build`, but didn't think to run `yarn`. ( I don't have `wp-scripts` globally, but 🤷🏻‍♂️ )

public_html/wp-content/plugins/pattern-creator/readme.md
```diff
@@ -0,0 +1,22 @@
+# WordPress.org Block Pattern Creator
+
+This plugin adds an editor interface to the frontend of Block Pattern posts.
+
+## Requirements
+
+These should be available if you're using this environment.
+
+Plugins
+- Gutenberg
```
iandunn, line 10:
```suggestion
- Gutenberg (with Full Site Editing experiment enabled)
```

iandunn (approved):
LGTM. Left a few questions/suggestions, but no blockers.

public_html/wp-content/plugins/pattern-creator/readme.md
iandunn: Er, this isn't necessary after all.

public_html/wp-content/plugins/pattern-creator/src/style.css
ryelle: Not yet, but we can postcss easily enough.

public_html/wp-content/plugins/pattern-creator/pattern-creator.php
ryelle: I think it's safe to assume that these would be renamed when `gutenberg_experimental_global_styles_get_merged_origins` is, which I'm already checking in `should_load_creator`; in any case this section is somewhat copy-pasted to get it working, I don't see it staying like this for long.

iandunn: 👍🏻

public_html/wp-content/plugins/pattern-creator/src/style.css
iandunn: 👍🏻 

https://github.com/WordPress/pattern-directory/issues/7

public_html/wp-content/plugins/pattern-creator/src/store/actions.js
ryelle: The result from `select` isn't a promise, so yield and await are not interchangeable. We're not (only) waiting for a response, we want to generate multiple actions from this "action creator". So in the process of this function, we fire off the select action, which resolves into the patternId, then we keep going to fire off the save action. It's a bit overcomplicated for the current functionality, but we'll want to create other actions later (in case of errors, to save other meta, etc).

iandunn: Ah, ok.

ryelle:
I'm going to merge this with the `wp-pattern` CPT placeholder, but this can be updated to use the `wporg-pattern` from #6.

PR #6, opened 2020-10-28
Stub basic post type and API endpoint

### LINKED ISSUE #4: Stub pattern post type, meta, taxonomies
Necessary to iterate on #2

iandunn:
This creates a basic CPT, and tweaks Core's REST API endpoints to support listing and searching patterns.

Example API requests:

* Browse all: http://wporg-pattern.test/wp-json/wp/v2/wporg-pattern?_fields=title,content,meta,_links&_embed=wp:term
* Browse category: http://wporg-pattern.test/wp-json/wp/v2/wporg-pattern?_fields=title,content,meta,_links&_embed=wp:term&category=2
* Search: http://wporg-pattern.test/wp-json/wp/v2/search?search=head&subtype=wporg-pattern&_embed=self&_fields=_links

Fixes #4

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -1,0 +1,115 @@
+<?php
+
+namespace WordPressdotorg\Pattern_Directory\Pattern_Post_Type;
+
+const POST_TYPE = 'wporg-pattern';
+
+add_action( 'init', __NAMESPACE__ . '\register_post_type_data' );
+add_filter( 'rest_' . POST_TYPE . '_item_schema', __NAMESPACE__ . '\embed_extra_fields_in_search_endpoint' );
+add_action( 'enqueue_block_editor_assets', __NAMESPACE__ . '\enqueue_editor_assets' );
+
+
+/**
+ * Registers post types and associated taxonomies, meta data, etc.
+ */
+function register_post_type_data() {
+	register_post_type(
+		POST_TYPE,
+		array(
+			'label'        => 'Block Pattern',
+			'description'  => 'Stores publicly shared Block Patterns (predefined block layouts, ready to insert and tweak).',
+			'public'       => true,
+			'show_in_rest' => true,
+			'rewrite'      => array( 'slug' => 'pattern' ),
+			'supports'     => array( 'title', 'editor', 'author', 'custom-fields', ),
+		)
+	);
+
+	register_taxonomy(
+		'wporg-pattern-category',
+		POST_TYPE,
+		array(
+			'public'        => true,
+			'hierarchical'  => true,
+			'show_in_rest'  => true,
+			'rest_base'     => 'category',
+		)
+	);
+
+	register_post_meta(
+		POST_TYPE,
+		'wpop_description',
+		array(
+			'type'              => 'string',
+			'description'       => 'A description of the pattern',
+			'single'            => true,
+			'default'           => '',
+			'sanitize_callback' => 'sanitize_text_field',
+			'auth_callback'     => __NAMESPACE__ . '\can_edit_this_pattern',
+			'show_in_rest'      => true,
+		)
+	);
+
+	register_post_meta(
+		POST_TYPE,
+		'wpop_viewport_width',
+		array(
+			'type'              => 'number',
+			'description'       => 'The width of the pattern in the block inserter.',
+			'single'            => true,
+			'default'           => 0,
+			'sanitize_callback' => 'absint',
+			'auth_callback'     => __NAMESPACE__ . '\can_edit_this_pattern',
+			'show_in_rest'      => true,
+		)
+	);
+}
+
+/**
+ * Determines if the current user can edit the given pattern post.
+ *
+ * This is a callback for the `auth_{$object_type}_meta_{$meta_key}` filter, and it's used to authorize access to
+ * modifying post meta keys via the REST API.
+ *
+ * @param bool   $allowed
+ * @param string $meta_key
+ * @param int    $pattern_id
+ *
+ * @return bool
+ */
+function can_edit_this_pattern( $allowed, $meta_key, $pattern_id ) {
+	return current_user_can( 'edit_post', $pattern_id );
+}
+
+/**
+ * Add extra fields to the post type when it's embedded in the search results endpoint.
+ *
+ * @param array $schema
+ *
+ * @return array
+ */
+function embed_extra_fields_in_search_endpoint( $schema ) {
+	$schema['properties']['content']['context'][]                           = 'embed';
+	$schema['properties']['content']['properties']['rendered']['context'][] = 'embed';
+	$schema['properties']['meta']['context'][]                              = 'embed';
+	$schema['properties']['category']['context'][]                          = 'embed';
+
+	return $schema;
+}
+
+/**
+ * Enqueue scripts for the block editor.
+ */
+function enqueue_editor_assets() {
+	if ( POST_TYPE !== get_current_screen()->id ) {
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/pattern-directory/javascript/pattern-post-type.js
```diff
@@ -0,0 +1,70 @@
+/*
+ * There's no need for webpack etc, since we can assume wp-admin users will have modern browsers.
+ *
+ * JSX is the only thing that'd be nice to have, but it's not worth the tooling costs for just a few fields.
```
ryelle, line 4: Since we're already setting up the workspaces with a shared wp-scripts build process, I think it would be a better dev-experience to have this use it - we can let the build step handle pulling out dependencies, and will save us from needing to remember that one place uses a different syntax than others.

public_html/wp-content/plugins/pattern-directory/javascript/pattern-post-type.js
```diff
@@ -0,0 +1,70 @@
...
+const PatternPostType = () => {
+	const { useSelect, useDispatch } = wp.data;
+	const { createElement, Fragment, useState, useEffect } = wp.element;
+	const { TextControl, TextareaControl } = wp.components;
+	const { PluginDocumentSettingPanel } = wp.editPost;
+	const { editPost } = useDispatch( 'core/editor' );
+
+	const postMetaData = useSelect( select => select( 'core/editor' ).getEditedPostAttribute( 'meta' ) || {} );
+	const [ description, setDescription ] = useState( postMetaData.wpop_description );
+	const [ viewportWidth, setViewportWidth ] = useState( postMetaData.wpop_viewport_width );
```
ryelle, line 17: I made a `usePostMeta` hook [for WordCamp CPTs](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wc-post-types/js/src/components/hooks/use-post-meta.js) — not saying you need to change anything, but it could be helpful here.

public_html/wp-content/plugins/pattern-directory/javascript/pattern-post-type.js
```diff
@@ -0,0 +1,70 @@
...
+				},
+			} );
+		},
+		[ description, viewportWidth ]
+	);
+
+	const descriptionInput = createElement(
+		TextareaControl,
+		{
+			label: 'Description',
```
ryelle, line 36:
```suggestion
			key: 'description',
			label: 'Description',
```

public_html/wp-content/plugins/pattern-directory/javascript/pattern-post-type.js
```diff
@@ -0,0 +1,70 @@
...
+			value: description,
+			onChange: setDescription
+		}
+	);
+
+	const viewportWidthInput = createElement(
+		TextControl,
+		{
+			label: 'Viewport Width',
```
ryelle, line 46:
```suggestion
			key: 'viewportWidth',
			label: 'Viewport Width',
```

public_html/wp-content/plugins/pattern-directory/javascript/pattern-post-type.js
```diff
@@ -0,0 +1,70 @@
...
+			value: viewportWidth,
+			onChange: setViewportWidth,
+			type: 'number' // Replace this with a `NumberControl` once it's stable.
+		}
+	);
+
+	const container = createElement(
+		Fragment,
+		{},
+		[ descriptionInput, viewportWidthInput ]
```
ryelle, line 56: Since this is an array, it needs the key prop — I dropped that in as suggestions above.

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
```diff
@@ -0,0 +1,118 @@
...
+		POST_TYPE,
+		array(
+			'label'        => 'Block Pattern',
+			'description'  => 'Stores publicly shared Block Patterns (predefined block layouts, ready to insert and tweak).',
+			'public'       => true,
+			'show_in_rest' => true,
+			'rewrite'      => array( 'slug' => 'pattern' ),
+
+			'supports' => array(
+				'title', 'editor', 'revisions', 'author', 'excerpt', 'thumbnail', 'custom-fields',
```
ryelle: If we're adding a separate meta field for the description, what's the use for the excerpt?

ryelle (commented):
CPT looks good (once I locally fixed the console warning with the keys) 👍 

Since this isn't using wp-scripts, it's not running the linter, and there are lint issues in the JS. You should be able to copy [the pattern-creator package.json entirely,](https://github.com/WordPress/pattern-directory/blob/trunk/public_html/wp-content/plugins/pattern-creator/package.json) then add this project as a workspace [here,](https://github.com/WordPress/pattern-directory/blob/41bdf05aaa41d1e5ba7d803c430f1cb4955b3315/package.json#L19) and it'll automatically work. Well, the JS file will need to move to `src`, or you can pass `--entry=./javascript/pattern-post-type.js` to wp-scripts, it'll still build to `/build`. This can happen in this PR or a follow up.

iandunn:
> the console warning with the keys

What was that?

public_html/wp-content/plugins/pattern-directory/javascript/pattern-post-type.js
iandunn: I can see the appeal of everything using the same tooling, but it still feels to me like this doesn't _need_ any tooling, and that the devex is better without me.

The existing tooling might might it easier to set this up, but it could still be a pain. The biggest cost IMO is _using_ the tooling, though. It's slow, it's breaks often and is painful to fix, etc.

This is already working, too, so I don't see any benefit from redoing it now. If we want to add more fields or components later, then I can see that making the plugin complex enough where JSX would be worth the costs.

public_html/wp-content/plugins/pattern-directory/javascript/pattern-post-type.js
iandunn: 🤔 I struggle a lot with Gutenberg's data fetch/set architecture, and I think others do as well. At this point I think having to understand an additional abstraction on top of that would only make it harder.

public_html/wp-content/plugins/pattern-directory/javascript/pattern-post-type.js
iandunn: Thanks!

public_html/wp-content/plugins/pattern-directory/javascript/pattern-post-type.js
iandunn: Thanks, merged 👍🏻

I didn't get the console errors like I have in past situations, but 🤷🏻‍♂️

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
iandunn: 🤔 that's a great point, we probably don't need it. I'll remove it.

iandunn:
Oh, you _were_ getting that typical warnings about missing keys. Huh, that's weird that I didn't. Oh well 🤷🏻‍♂️

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
iandunn: I went ahead and removed `thumbnail` and `revisions` too, since we're not using them (at least not yet).

iandunn:
I'm gonna merge this since I'm leaving on AFK soon, but feel free to open an issue about JSX, or just make the change if you feel strongly.

PR #12, opened 2020-11-09
CPT Plugin: Add wp-scripts support 

ryelle:
Following up from #6 

> Since we're already setting up the workspaces with a shared wp-scripts build process, I think it would be a better dev-experience to have this use it - we can let the build step handle pulling out dependencies, and will save us from needing to remember that one place uses a different syntax than others.

To test:

- If you haven't yet, run `yarn` to install wp-scripts (or re-install, if you run into "Cannot read property 'tap' of undefined", a workaround for this was added in `wp-scripts@12.5.0`)
- Run `yarn workspaces run build` — both the pattern-creator and pattern-directory projects should build
- Check that the Pattern CPT (`/wp-admin/edit.php?post_type=wporg-pattern`) has a panel in the document sidebar with description and viewport width.
- Run `yarn workspaces run lint:js` — both the pattern-creator and pattern-directory projects should be linted
- Run `yarn workspaces run lint:css` — only the pattern-creator has CSS, the pattern-directory project should output "No CSS."

Those last two are also run as part of the github workflow.


ryelle:
~if the composer/workflow change above fixes the check I'll pull it out into trunk before merging this.~ ~nevermind, this is not as simple a solution as I expected.~ With #13 this is all fixed & the checks are working again.

StevenDufresne:
Looks good to me.

tellyworth:
Works for me.

Issue #14, opened 2020-11-09
Create pre-submission panel for pattern directory information


MichaelArestad:
We need a way for a user to create their pattern and input the necessary metadata for it (name, description, tags, etc). I suspect this should pretty much just be the Gutenberg editor so, to start, I mocked that up with a pre-submission flow unique to this use case.

[Try the prototype](https://www.figma.com/proto/bz4qs4jZCkqvQzPH2fmX6C/Pattern-Directory?node-id=139%3A4003&viewport=-2629%2C884%2C0.5&scaling=min-zoom).

![2020-11-09 16 27 35](https://user-images.githubusercontent.com/1123119/98609263-1cecd580-22aa-11eb-84c9-5a966c5c2784.gif)

The pre-submission sidebar gives the user inputs for a pattern name, description, tags (the usual tag interface), and some controls to help them create an ideal pattern preview.

## Screenshots

<img width="1439" alt="image" src="https://user-images.githubusercontent.com/1123119/98610744-8d492600-22ad-11eb-84b3-734d712ec60a.png">

![pre-submit](https://user-images.githubusercontent.com/1123119/98610778-a4881380-22ad-11eb-93e1-95c10ba3ff84.jpg)

## Outstanding questions

1. Should the editor have a different icon/header/something to indicate it's the pattern submission tool and not the regular editor on one of the user's sites?
2. Are there controls we should remove from the editor (like block installation)?
3. What does the user see next? I imagine either a success message on their Submitted Patterns page or a login flow that then leads them there.
4. What are the errors a user might encounter in creating a pattern?

ryelle:
In #21 I set up the `viewportWidth` as a dropdown of fixed widths (narrow: 800, wide: 1100, extra wide: 1400)  — I'm not sure there's much benefit in letting people control this to the pixel, but maybe the slider could snap to the 100s or 50s?

> Should the editor have a different icon/header/something to indicate it's the pattern submission tool and not the regular editor on one of the user's sites?

Is the full screenshot the entire page, or do you still expect to have the WordPress.org wrapper around the content like this (except not a hacky photoshop 🙂  )
![Screen Shot 2020-11-10 at 7 02 33 PM](https://user-images.githubusercontent.com/541093/98748283-534c5280-2387-11eb-8835-01d4e4df9a83.png)

> Are there controls we should remove from the editor (like block installation)?

[...]

Issue #9, opened 2020-11-06
Build the pattern submission tool.


tellyworth:
Patterns should be submitted from a purpose-built front-end form based on Gutenberg.

Tentative assumptions:
* Any logged-in user should be able to submit a pattern.
* It will be necessary to preview and edit a pattern during the submission process.
* Submitted patterns will need to pass some kind of validation, curation, moderation process before being published to the world.
* Users might expect to be able to see and use patterns they have submitted even if they're not yet published.
* Users will need to provide some metadata with their patterns, such as description and language.
* Copying or "forking" an existing pattern will probably be a common flow.
* Translations and i18n will need to be solved.

This is an umbrella task, specific issues should be created as needed.

MichaelArestad:
I created a few issues for the above:

Pattern submission tool: https://github.com/WordPress/pattern-directory/issues/14
Login flow: https://github.com/WordPress/pattern-directory/issues/15
Submitted patterns page: https://github.com/WordPress/pattern-directory/issues/16
Forking a pattern: https://github.com/WordPress/pattern-directory/issues/17
Pattern submission modal: https://github.com/WordPress/pattern-directory/issues/18

ryelle:
I think every task here is covered by an existing issue, and this "parent issue" can be closed because the project is being tracked in https://github.com/WordPress/pattern-directory/projects/1 — I prefer to keep issues to actionable tasks that can be closed, but @tellyworth if you'd rather have this issue you can reopen it.

PR #21, opened 2020-11-10
Pattern Creator: Hook up pattern meta UI

ryelle:
See #9 

The current pattern creator UI has some fields for description, viewport width, etc. This PR hooks those fields up with the backend meta, so they can be saved. This is still using the proof-of-concept UI, in a future PR we'll implement the design in https://github.com/WordPress/pattern-directory/issues/14

This hooks up the description & viewport to post meta, and the pattern category to the right taxonomy.

Screen Shot 2020-11-10 at 6 14 47 PM

To test

- Run `yarn workspace wporg-pattern-creator run build`
- Create some pattern categories, ex Header, Buttons, Columns.
- If needed, create some Block Patterns in wp-admin, `/wp-admin/post-new.php?post_type=wporg-pattern`.
- View a published pattern, if you're logged in you should see the pattern editor.
- Play around & save your pattern, it should save.
- Change some meta (title isn't hooked up, everything else should work)
- Reload the page, view it in wp-admin, or the API — the meta should save.

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/src/components/inspector/settings.js
+++ b/public_html/wp-content/plugins/pattern-creator/src/components/inspector/settings.js
@@ -3,13 +3,16 @@
  */
 import { PanelBody, PanelRow, SelectControl, TextControl, TextareaControl } from '@wordpress/components';
 
+/**
+ * Internal dependencies
+ */
+import usePostMeta from '../../store/hooks/use-post-meta';
+
 export default function Settings() {
 	const title = '';
-	const description = '';
-	const categories = '';
-	// const keywords = '';
-	// const viewportWidth = '';
 	const onChange = () => {};
+	const [ description, setDescription ] = usePostMeta( 'wpop_description', '' );
+	const [ viewportWidth, setViewportWidth ] = usePostMeta( 'wpop_viewport_width', '' );
 
 	return (
 		<>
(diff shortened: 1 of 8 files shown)


public_html/wp-content/plugins/pattern-creator/src/components/inspector/settings.js
```diff
@@ -18,15 +31,24 @@ export default function Settings() {
...
+					value={ term }
+					options={ termList }
+					onChange={ setTerms }
+				/>
+				<SelectControl
+					label="Preview Width"
+					value={ viewportWidth }
+					onChange={ setViewportWidth }
+					options={ [
+						{ value: 800, label: 'Normal' },
```
StevenDufresne, line 47: Should these come from `gutenberg` at some point?

StevenDufresne:
This looks good. It was working as expected. 

We may want to get some unit tests running soon, but i'm ok if this moves in without them since its early days.

public_html/wp-content/plugins/pattern-creator/src/components/inspector/settings.js
ryelle: They're just placeholder, this whole control will be revamped with the real design in #14

ryelle:
> We may want to get some unit tests running soon, but i'm ok if this moves in without them since its early days.

Yep, but I was worried this PR was getting big (and most of the stuff here is just overlaying core functions). We'll get a test framework & some tests started soon.

PR #32, opened 2020-12-22
Add a "create a pattern" flow

ryelle:
This is a very bare-bones attempt at a flow to create a new block pattern. It will let you create a new draft pattern, which you can edit and save.

**To test**

- Create a page with the slug `new-pattern` — this was the most straightforward way to set this up locally, but we could also use a page template
- Go to that page, you should have a page with an empty editor
- Enter some content, hit submit and publish
- View the patterns in wp-admin, your new pattern should show up
- Try as a logged out user, it should not let you create a pattern

⚠️  Things to follow up on (in new PRs)
- It only creates drafts - we need a validation step somewhere in here
- The Settings panel should be converted to use something like the PrePublish/PostPublish components, maybe? Something to give more feedback that the pattern was saved and is either being validated or is now available on wp.org.

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
+++ b/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
@@ -25,10 +25,10 @@ const QUERY_VAR = 'edit-pattern';
  */
 function should_load_creator() {
 	global $wp_query;
-	return $wp_query->is_singular( POST_TYPE ) &&
-		// @todo Permissions TBD, see https://github.com/WordPress/pattern-directory/issues/30.
-		current_user_can( 'edit_post', get_the_ID() ) &&
-		false !== $wp_query->get( QUERY_VAR, false );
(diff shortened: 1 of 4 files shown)


public_html/wp-content/plugins/pattern-creator/pattern-creator.php
```diff
@@ -53,6 +53,9 @@ function enqueue_assets() {
 		return;
 	}
 
+	/** Load in admin post functions for `get_default_post_to_edit`. */
+	require_once ABSPATH . 'wp-admin/includes/post.php';
```
ryelle, line 57: I'm not sure if this will have any adverse effects, loading this file on the frontend. It hasn't in my testing, but I'd like a gut-check.

iandunn: 🤔  

I don't see any logic being executed in it, just functions declared. I can't think of any side effects, unless some logic is added in the future, but that seems unlikely to me 👍🏻

PR #44, opened 2021-02-19
Use ElasticSearch/Jetpack to customize search

### LINKED ISSUE #28: Search description instead of post content
Normally Core searches `post_content`, but that probably won't make sense for most patterns, since that's storing the HTML and demo content. It'd make more sense to search the meta description field instead.

See #11

One way to do this might be to save the description into `post_excerpt`, instead of a meta field. Then it'd be searched for free. We'd still want to disable searching `post_content`, though. 

That'd also require changing the two APIs that proxy this API, which would be a pain. We might also want to search by other post meta per #11, so we might need that functionality anyway. We'll probably need to update them anyway, as the design evolves, so maybe those updates can be batched together after more are settled.

iandunn:
fixes #28 , WIP

xref https://github.com/tellyworth/plugin-directory/pull/1

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/search.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/search.php
@@ -116,6 +116,7 @@ function modify_es_query_args( $es_query_args, $wp_query ) {
 		$should_query[] = [
 			'term' => [
 				'meta.wpop_locale.value.raw' => $primary_locale,
+				'boost'                      => 2,
 			],
 		];
 
@@ -123,7 +124,7 @@ function modify_es_query_args( $es_query_args, $wp_query ) {
 			'term' => [
 				'meta.wpop_locale.value.raw' => 'en_US',
 				'boost'                      => 0.00001,
-				// todo ^ isn't working
+				// todo ^ isn't working. might not need the positive boost on $primary_locale once this works.
 			],
 		];
 	}
--- a/public_html/wp-content/plugins/pattern-directory/includes/search.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/search.php
@@ -2,8 +2,16 @@
 
 namespace WordPressdotorg\Pattern_Directory\Search;
 
+use WP_Query, Jetpack_WPES_Search_Query_Parser;
+use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\POST_TYPE;
 
+// Note: This depends on the Search module and custom index being enabled in `mu-plugins/main-network/site-patterns.php`.
+
+add_filter( 'jetpack_search_should_handle_query', __NAMESPACE__ . '\should_handle_query', 10, 2 );
+add_filter( 'jetpack_search_es_query_args', __NAMESPACE__ . '\modify_es_query_args', 10, 2 );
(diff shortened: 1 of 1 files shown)


iandunn:
so far this is fairly straightforward, but i'll know more once the access is setup on the jetpack/es side

fyi @dd32, @tellyworth

ryelle:
@iandunn What's the status of this PR/issue #28? Our initial launch deadline is Friday, since we're aiming to be stable by [5.8's release](https://make.wordpress.org/core/5-8/) next Tuesday (20th). Will we have ElasticSearch support by then?

iandunn:
Oh, I never saw that the website had a deadline tied to the Core release; it's unlikely this will be done by then. I'm AFK Thursday/Friday too.

I'm guessing using regular WP search will be performant with the small number of patterns we have now, but the quality of the results will be poor because of post content being included instead of descriptions.

It'd potentially be faster to [create a meta field with index terms](https://github.com/WordPress/pattern-directory/issues/28#issuecomment-736002596) as a temporary measure, but I'd lean towards just focusing on getting this done soon.

ryelle:
Okay, it sounds like that's acceptable for the initial launch, since the api.w.org only serves a subset of patterns. Moving #28 out of Initial Launch.

iandunn:
@tellyworth , @gibrown, @earnjam: This is mostly working, and ready for review if you have any thoughts.

I think the last bits are boosting the title over the description, and including the `en_US` fallback patterns. Those shouldn't impact the overall architecture, etc though. I'll keep working on those.

I've got e2e tests on my sandbox that I can commit to `meta.svn` once this is merged.

iandunn:
f23323e boosts the primary locale in some cases, but not enough to make them always appear before the `en_US` fallback. The latter is how the directory behaves for browsing, but maybe in a search context we want still want the relevancy to override locale in some cases?

If not, maybe a `sort` would be a better approach than boosting?

iandunn:
merged e2e tests: https://meta.trac.wordpress.org/log?rev=11187&stop_rev=11184

iandunn:
@earnjam: most of the w.org mu-plugins are still closed-source, but here's a copy of the one that `search.php` relies on, in case it's helpful for the theme directory.

```php
<?php

namespace WordPressdotorg\Pattern_Directory\MU_Plugin;

/*
 * Most of the site functionality is in `wp-content/plugins/pattern-*`, but some things are more appropriate for
 * an mu-plugin, or need to be done before plugins are loaded.
 *
 * We don't have a paid Jetpack plan, so the Search feature would normally be off. The Jetpack team has manually
 * created custom indexes for Patterns, though, letting us use this feature for free.
 *
 * This has to be done before `plugins_loaded`.
 * @link https://docs.wpvip.com/technical-references/elasticsearch/integrating-jetpack-search/.
 */

define( 'JETPACK_SEARCH_VIP_INDEX', true );

add_filter( 'pre_option_has_jetpack_search_product', '__return_true' );
add_filter( 'jetpack_active_modules', __NAMESPACE__ . '\enable_jetpack_search_module', 9999 );

function enable_jetpack_search_module( $modules ) {
	if ( ! in_array( 'search', $modules, true ) ) {
		$modules[] = 'search';
	}

	return $modules;
}
```

PR #49, opened 2021-03-29
Theme: Style page headers

ryelle:
This updates the page header (the blue section) style to match the mockups.

### Screenshots

Home
Screen Shot 2021-03-29 at 5 25 54 PM

Interiors
Screen Shot 2021-03-29 at 5 27 06 PM
Screen Shot 2021-03-29 at 5 25 43 PM
Screen Shot 2021-03-29 at 5 25 33 PM

Small screens
Screen Shot 2021-03-29 at 17 26 10
Screen Shot 2021-03-29 at 17 26 21

Regular pages still include the page title - this will probably change as the rest of the site is built out.
Screen Shot 2021-03-29 at 5 58 05 PM

1. View some pages (home, a single pattern, the "new pattern" page)
2. Make sure the headers look correct

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_site-header.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_site-header.scss
@@ -1,0 +1,56 @@
+body.home {
+	.site-header {
+		padding: 3.75rem 1rem;
+		text-align: left;
+	}
+
+	.site-branding {
+		display: grid;
+		grid-template-areas:
+			"title   callout"
+			"tagline callout";
+		grid-template-rows: auto 1fr;
+		grid-template-columns: 1fr 18rem;
+		gap: 0 4rem;
+		justify-content: center;
+		align-items: center;
+	}
+
+	// Duplicate class for specifity override.
+	.site-title.site-title {
+		grid-area: title;
+		margin-top: 0;
+		text-align: left;
+		font-size: 3rem;
+		line-height: $type__header--line-height;
+	}
+
+	.site-description {
+		grid-area: tagline;
+		margin: 0;
+		font-size: 1.25rem;
+		line-height: $type__line-height;
+		text-align: left;
+		color: $color-white;
+	}
+
+	.site-callout {
+		grid-area: callout;
+		padding: 1.125rem 1.5rem 1.5rem;
+		font-size: 0.875rem;
+		line-height: $type__line-height;
+		background: $color-white;
+		box-shadow: 0.75rem 0.75rem 0 rgba($color-black, 0.15);
+
+		h2 {
+			margin-top: 0;
+			font-size: 1.125rem;
+			line-height: $type__header--line-height;
+			font-weight: 600;
+		}
+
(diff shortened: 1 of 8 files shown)


public_html/wp-content/themes/pattern-directory/css/components/_site-header.scss
```diff
@@ -0,0 +1,88 @@
...
+		text-align: left;
+	}
+
+	.site-branding {
+		display: grid;
+		grid-template-areas:
+			"title   callout"
+			"tagline callout";
+		grid-template-rows: auto 1fr;
+		grid-template-columns: 1fr 18rem;
```
StevenDufresne: IE 11 doesn't support grid-template-* without prefixes, I believe. I think we have a prefixer configured but it's worth a manual check using IE 11 if you haven't already.

public_html/wp-content/themes/pattern-directory/functions.php
```diff
@@ -53,6 +56,8 @@ function enqueue_assets() {
 
 		wp_set_script_translations( 'wporg-pattern-script', 'wporg-patterns' );
 	}
+
```
StevenDufresne, line 59: Is `20151215` supposed to have any meaning?

StevenDufresne (commented):
In the non-home `site-header`, there is no padding added to any of the clickable elements making them hard to click with a mouse. 

Should we remove the padding from `site-header` and add it to the anchor elements?

Red is the clickable target:

StevenDufresne:
I'm getting a space under the mobile menu when it's open:

public_html/wp-content/themes/pattern-directory/functions.php
ryelle: It's the version used in [the parent theme's functions.php](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg/functions.php#L114), but I see now that the last commit to navigation.js was 12/09/2018 so I guess it should be `20181209`… but it looks like that change was not made to the minified file 🙃  

Well, that was a 🕳️ , but the minified file is now updated and I've created [this ticket](https://meta.trac.wordpress.org/ticket/5687) to register this file in the parent theme. I'll update the date here for now, and hopefully we can remove the duplicate setup later.

public_html/wp-content/themes/pattern-directory/css/components/_site-header.scss
ryelle: We didn't have the grid prefixer set up here, since it's an opt-in option on autoprefixer. I added it, and switch to manually laying out these items for IE11 compat.

ryelle:
> Should we remove the padding from site-header and add it to the anchor elements?

I don't think we need to do that, these are regular text links, not buttons. Other wp.org sites don't add any extra padding to these links.

> I'm getting a space under the mobile menu when it's open:

Looks like that happens if you don't assign a menu to the space. If you create a menu and add it, it should be correct.

Issue #48, opened 2021-03-29
Directory home page / list view


tellyworth:
The home page view of the Pattern Directory should feature a masonry-style grid, and a short list of filters/types for refining the view:

![patterns-signed-out-1x](https://user-images.githubusercontent.com/7200686/112904267-698b5d00-9134-11eb-91e5-29a3baad37d6.gif)

Along the top of the patterns grid is a “sort by” dropdown, followed by a list of selected categories, and then a search box. Changing the sorting or selecting a category updates the grid and exposes a secondary UI that indicates the current criteria (category, author, tag, etc) along with space for displaying related information like related categories or an author’s website.

<img width="1024" alt="image-8" src="https://user-images.githubusercontent.com/7200686/112904613-ecacb300-9134-11eb-88d7-3cd821be45f5.png">

The Patterns site focuses around a masonry-style grid of rendered patterns. The initial designs have a discrete “Load more” button, but infinite scroll could be added in future. Within the grid, Patterns focus on showing there rendered output, unlike the Plugins/Themes directory where a static image is required.

![mason-grid-2](https://user-images.githubusercontent.com/7200686/112905307-eb2fba80-9135-11eb-8b4d-245def33b2ee.gif)

iandunn:
What's a good way to determine the related categories? 🤔 

If there'll be a hardcoded or manually curated list of categories, then we could just have a hardcoded map.

If not, then maybe look at patterns that have multiple categories assigned, and then count the number of times categories are used together? e.g., The selected category is `header`, so look all the patterns with `header` assigned`, and then count the instances of other categories, and pick the top 3. That would probably have lots of false-positives and false-negatives, though.

WordNet could be helpful, but we'd have to know [which synset to specify](http://wordnetweb.princeton.edu/perl/webwn?s=header&sub=Search+WordNet&o2=&o0=1&o8=1&o1=1&o7=&o5=&o9=&o6=&o3=&o4=&h=), and it wouldn't work for jargon like `hero`. It looks like they'd also want a citation in the UI, which feels like a dealbreaker to me. 

Maybe something using [significant term aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-significantterms-aggregation.html) or [synonyms](https://www.elastic.co/blog/boosting-the-power-of-elasticsearch-with-synonyms) in ElasticSearch?

iandunn:
[Datamuse](https://www.datamuse.com/api/) looks interesting.

[...]

PR #50, opened 2021-03-30
[WIP] Add an active filter bar component.

StevenDufresne:
This PR introduces a front-end component that provides context to the user about their current selected pattern.

See #48 for more context

### Example Usage
Since I'm unclear on how the frontend or backend will work, this should be considered a first attempt. We may want to use a more composite approach to the component; breaking out the "actions". 

```
<ActiveFilterBar
  actionsTitle="Related Categories"
  actions={ [
    {
        label: 'Ecommerce',
        href: '#intro',
    },
    {
        label: 'Columns',
        href: '#columns',
    },
    {
        label: 'Marketing',
        href: '#marketing',
    },
  ] }
>
12 patterns by <b>Bill Brasky</b>
</ActiveFilterBar>
```

### Questions
- In terms of mobile, I just hid the "actions side" of the component
  - It only works on page load right now, I'm not sure what our/gutenberg's preferred way to track window resize.
- Are we keen on building out the components using JS or do we want to build them in PHP templates?

Although there isn't a great way to test this since the parent UI doesn't exist yet.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/package.json
+++ b/public_html/wp-content/themes/pattern-directory/package.json
@@ -28,10 +28,12 @@
 		"@wordpress/browserslist-config": "3.0.1",
 		"@wordpress/components": "13.0.0",
 		"@wordpress/compose": "3.25.0",
+		"@wordpress/data": "^4.27.0",
 		"@wordpress/element": "2.20.0",
 		"@wordpress/i18n": "3.19.0",
 		"@wordpress/keycodes": "2.19.0",
 		"@wordpress/scripts": "14.0.1",
+		"@wordpress/viewport": "^2.26.0",
 		"autoprefixer": "9.8.6",
 		"cssnano": "4.1.10",
 		"grunt": "1.3.0",
@@ -58,8 +60,5 @@
 			"**/*.css.map"
 		]
 	},
-	"dependencies": {
-		"@wordpress/data": "^4.27.0",
-		"@wordpress/viewport": "^2.26.0"
-	}
+	"dependencies": {}
 }
--- a/public_html/wp-content/themes/pattern-directory/package.json
+++ b/public_html/wp-content/themes/pattern-directory/package.json
@@ -15,10 +15,9 @@
 		"build:css": "grunt css",
 		"build:js": "grunt js",
 		"dev": "grunt",
-		"format:js": "echo \"No JS.\"",
+		"format:js": "wp-scripts format-js src",
 		"lint:css": "wp-scripts lint-style css",
 		"lint:js": "wp-scripts lint-js src",
-		"format:js": "wp-scripts format-js src",
 		"packages-update": "wp-scripts packages-update"
 	},
 	"browserslist": [
@@ -54,6 +53,13 @@
 	"prettier": "../../../../.prettierrc.js",
 	"stylelint": {
 		"extends": "../../../../.stylelintrc",
-		"ignoreFiles": ["**/*.css", "**/*.css.map"]
+		"ignoreFiles": [
+			"**/*.css",
+			"**/*.css.map"
+		]
+	},
+	"dependencies": {
+		"@wordpress/data": "^4.27.0",
+		"@wordpress/viewport": "^2.26.0"
 	}
 }
--- a/public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
@@ -1,0 +1,37 @@
+/**
+ * External dependencies
+ */
+import { select } from '@wordpress/data';
+import { store } from '@wordpress/viewport';
+import { Flex, FlexItem } from '@wordpress/components';
+
+function ActiveFilterBar( { children, actionsTitle, actions } ) {
+	const isSmall = select( store ).isViewportMatch( '< medium' );
+
+	return (
+		<Flex className="active-filter__bar" align="center">
+			<FlexItem className="active-filter__bar__copy">{ children }</FlexItem>
+			{ ! isSmall && actions.length > 0 && (
+				<FlexItem>
+					<Flex gap={ 0 }>
+						<FlexItem>
+							<h3 className="active-filter__bar__title">{ actionsTitle } </h3>
+						</FlexItem>
+
+						<FlexItem>
+							<ul>
+								{ actions.map( ( i ) => (
+									<li key={ i.href }>
+										<a href={ i.href }>{ i.label }</a>
+									</li>
+								) ) }
+							</ul>
+						</FlexItem>
+					</Flex>
+				</FlexItem>
+			) }
+		</Flex>
+	);
+}
+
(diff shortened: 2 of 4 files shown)


public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
```diff
@@ -0,0 +1,37 @@
...
+					<Flex gap={ 0 }>
+						<FlexItem>
+							<h3 className="active-filter__bar__title">{ actionsTitle } </h3>
+						</FlexItem>
+
+						<FlexItem>
+							<ul>
+								{ actions.map( ( i ) => (
+									<li key={ i.href }>
+										<a href={ i.href }>{ i.label }</a>
```
iandunn, line 25: The example showed the `href` values like `#columns`; would clicking on that trigger a change in the active filters and results?

If so, it seems like `<button>` is more semantically appropriate.

public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * External dependencies
+ */
+import { select } from '@wordpress/data';
+import { store } from '@wordpress/viewport';
+import { Flex, FlexItem } from '@wordpress/components';
+
+function ActiveFilterBar( { children, actionsTitle, actions } ) {
+	const isSmall = select( store ).isViewportMatch( '< medium' );
```
iandunn, line 9: _This is a minor thing, feel free to disregard it if it doesn't feel good to you_.

```suggestion
	const isGreaterThanSmall = select( store ).isViewportMatch( '>= small' );
```

This would allow the condition below to be `{ isGreaterThanSmall...` instead of `{ ! isSmall ...`. It feels more readable to me without the negative. 

> We may want to use a more composite approach to the component; breaking out the "actions".

That sounds good to me, and would allow using `ifViewportMatches` here.

iandunn, line 9: > It only works on page load right now, I'm not sure what our/gutenberg's preferred way to track window resize.

What's the reason to use a non-native solution necessary here (even one provided by the platform)? Media queries could be more straight forward and future-proof.

public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
```diff
@@ -0,0 +1,37 @@
+/**
```
iandunn, line 1: > Are we keen on building out the components using JS or do we want to build them in PHP templates?

Since it's interactive, JS seems to give a better UX IMO, since it will be faster/smoother.

public_html/wp-content/themes/pattern-directory/package.json
```diff
@@ -54,6 +53,13 @@
...
 		"extends": "../../../../.stylelintrc",
-		"ignoreFiles": ["**/*.css", "**/*.css.map"]
+		"ignoreFiles": [
+			"**/*.css",
+			"**/*.css.map"
+		]
+	},
+	"dependencies": {
+		"@wordpress/data": "^4.27.0",
+		"@wordpress/viewport": "^2.26.0"
```
ryelle: Can you put these in devDependencies with the other wp deps?

public_html/wp-content/themes/pattern-directory/package.json
```diff
@@ -15,10 +15,9 @@
 		"build:css": "grunt css",
 		"build:js": "grunt js",
 		"dev": "grunt",
-		"format:js": "echo \"No JS.\"",
+		"format:js": "wp-scripts format-js src",
 		"lint:css": "wp-scripts lint-style css",
```
ryelle, line 19: Oh nice cleanup 👍🏻

public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
ryelle: Since this is a navigation change, it should be a link. In production these will be category (or keyword) archives, ex: `https://wordpress.org/patterns/pattern-categories/columns/`

StevenDufresne: 👍

public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
StevenDufresne: Good question. 

I convinced myself to use Gutenberg's viewport hook based on these assumptions:
1. We will (probably) be using a number of Gutenberg components that will respond based on Gutenberg breakpoints. This will keep the UI in sync.
2. Using JS approach would make the component slightly more portable.

I could easily be convinced otherwise.

public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
StevenDufresne: Sounds good.

public_html/wp-content/themes/pattern-directory/package.json
StevenDufresne: Ah good catch. Thanks.

public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
StevenDufresne: Okay, i'll redo.

public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
iandunn: Ah, I didn't realize it was changing the URL, that's much better :)

public_html/wp-content/themes/pattern-directory/src/components/active-filter-bar/index.js
iandunn: I think those are good reasons, but personally I'm not sure it outweighs the resizing problems. 

There's the devex concern of course, but there's also a _UX_ concern, because some users will resize windows to fit the content, especially if they have larger monitors with room for multiple windows. 

In the Past GitHub didn't adapt when resizing the window, and it always bugged me.

That's probably a minority of users, though, so I can see the tradeoff, and don't feel strongly.

ryelle: We have the gutenberg breakpoints available in the settings, if you need them:

https://github.com/WordPress/pattern-directory/blob/3a81aeb0f200a2921b862d657134d530c82b2dd3/public_html/wp-content/themes/pattern-directory/css/settings/_structure.scss#L5-L11

Since this is just hiding content, I think a plain (CSS) media query would be better.

StevenDufresne:
I'm closing this as I have made the changes and merged the code into #65 since it is a standalone component.

PR #68, opened 2021-03-31
Theme: Add initial style and layout for the pattern grid

ryelle:
See #31, #62 — Start styling the pattern grid. PR only includes the PHP & CSS changes, I'm concurrently working on the JS side of things, but I think it makes more sense to do that in a following PR to avoid massive a changeset. So this does contain some placeholder sections (the preview, the filter bar).

### Screenshots

Screen Shot 2021-03-31 at 12 29 41

Smaller screen:
Screen Shot 2021-03-31 at 12 34 21

Hover or focus-within:
Screen Shot 2021-03-31 at 12 16 12

Hover or focus-within on a favorited pattern:
Screen Shot 2021-03-31 at 12 16 19

### How to test the changes in this Pull Request:

1. View any archive page (home, categories, search, … etc)
2. You should see a list of "patterns" (right now no preview, just a grey box of random height + the pattern ID).
3. Hover or tab through the patterns to see the title, copy, and favorite button.
4. Manually add the `is-favorited` class to a favorite button to see the favorited state.
5. Clicking on the pattern preview should bring you to the pattern itself.

Issue #66, opened 2021-03-31
Pattern Directory Structure: PHP and JS. 


StevenDufresne:
How do we want to organize the front end?

- Do we want to create a `front-page.php` for the grid?
- We currently have an `index.js` file in `src` that pulls in the pattern preview component. Are we going to leverage that `index.js` and turn it into an _app_ the same way an SPA works? 
   - If so, do need a router?
 - Alternatively, do we want to load a separate JS payload per _page?_ 

@ryelle @iandunn Thoughts?

ryelle:
I was already working on an idea here (for #62 ), so here's a quick overview of my thoughts… (I'll have at least a WIP PR up sometime today)

> Do we want to create a `front-page.php` for the grid?

The grid is on all archive pages & search results, so no, we would need to use it on all archives too. I ended up creating an `index.php`, and using a filter [like this](https://meta.trac.wordpress.org/changeset/10857) to override the archive & search templates to use that `index.php`.

I wanted to make this work with a non-SPA (PHP/HTML-first) approach, but overall the whole browsing experience is meant to be so interactive that I think it'll be easier to make a self-contained "SPA"-style component. And I just say SPA-style because I think we should keep the JS component inside the page content frame, and just for the grid browsing view.

> Are we going to leverage that `index.js` and turn it into an app the same way an SPA works? 

My intention was to have two main "components" (and possibly other functionality) in `src/index.js`. I set that file up to be an initializer of sorts, depending on the content on the page. So if it sees `.pattern-preview__container`, it will set up the resizable preview. If it sees a grid container, it will start up the grid component (WIP). This is pretty much where I stopped last night.

Ideally the default WP term archive & search URLs will still work, but whether we need a whole router I'm not sure yet. The navigation for most filters will "just" be clicking through different archive links, with dynamic loading of the matching patterns, and the ability to load more patterns infinite-scroll style.

> do we want to load a separate JS payload per page?

Maybe later, but I don't think we need to do that preemptively. It probably depends on the router (if we need to include something like react-router/reach-router, we don't need that on the single page).

StevenDufresne:
@ryelle While thinking about the JS structure, can you consider how we plan on organize how we fetch data? I'm almost ready to grab the taxonomies to fill the grid menu. I'm not too fussed about how we want to do it, just a general direction will help me with next steps.

For example:
1. Create a redux like store
2. Create a provider that stores the information and wraps the 'app'
3. Have high-level controls/modules that call via rest api
[...]

Issue #62, opened 2021-03-30
Pattern grid view functionality


ryelle:
Pulling this out of #48 — the pattern list view should be styled as a grid, and will either have a "load more" button or pagination. We can use JS to load the whole list (something SPA-style that controls the pattern grid), or just use URLs + JS for animation/flair. Probably will land on the JS solution, but will want something that works well across all archive & searching templates.

![](https://user-images.githubusercontent.com/7200686/112905307-eb2fba80-9135-11eb-8b4d-245def33b2ee.gif)

<img width="1020" alt="Screen Shot 2021-03-30 at 5 36 58 PM" src="https://user-images.githubusercontent.com/541093/113060363-89477300-917e-11eb-8081-59fff226bd40.png">

The preview cards themselves have a separate issue #31 , but there will probably be 1 PR for both issues.

ryelle:
Updated grid item design:

![Pattern Grid Item](https://user-images.githubusercontent.com/541093/117468963-3a31f200-af23-11eb-83e4-3a3872a0cdb2.png)

ryelle:
Closing this issue since the design has changed, and it makes more sense to open a new issue for the new card design.

PR #69, opened 2021-04-02
Theme: Try a `PatternGrid` component to display patterns

ryelle:
See #62 — This adds a few components to create a dynamic pattern grid. It uses the same structure and CSS introduced in #68, but also adds an iframe'd pattern preview.

This mainly gets the components in place, but doesn't touch anything data-y yet, not really. 

I'm using `select( coreStore ).getEntityRecords( 'postType', 'wporg-pattern' );` to get the list of patterns, but this generates an API request that only works for logged in editors. We'll want to create our own store to fetch just the plain `wporg-patterns` endpoint, not the edit context. See some discussion on https://github.com/WordPress/pattern-directory/issues/66#issuecomment-812450704.

❓ Would we consider this OK-enough to merge, or should we try to get the data side in first?

⚠️  I know the previews aren't great (design-wise), but I think I'd rather iterate on those in future PRs. Open to debate on that too 🙂 

### Screenshots

Screen Shot 2021-04-02 at 6 18 20 PM

0. Be logged in (see above re: API)
1. Load any archive page
2. It should render the list of patterns _(It will always show you the latest patterns, no filtering yet)_
3. Click on a pattern
4. It should take you to the pattern detail view, no change here


StevenDufresne (approved):
I'm fine with this being merged as-is. 

I'm a little skeptical of iframe usage at scale but we can adjust as the project matures.

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/index.js
```diff
@@ -0,0 +1,41 @@
...
+import PatternThumbnail from '../pattern-thumbnail';
+
+function PatternGrid() {
+	const posts = useSelect( ( select ) => {
+		// @todo This only works for logged in users, we'll need to create our own store for general use.
+		const { getEntityRecords } = select( coreStore );
+		return getEntityRecords( 'postType', 'wporg-pattern' );
+	} );
+	// `posts` will be null while the fetch happens.
+	const isLoading = ! Array.isArray( posts );
```
StevenDufresne, line 20: One thing to mention here is that `posts` will start as an empty array then switch to `null` when the query starts which makes sense for this component, but gives an intermediate state in the view which is more obvious in slower connections. It leads to some jank on load.

StevenDufresne (approved):

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/index.js
ryelle: Yeah, this is mostly placeholder for a real data infrastructure, where we could have a different selector for the `isLoading` state.

ryelle:
> I'm a little skeptical of iframe usage at scale but we can adjust as the project matures.

That's fair - we also have the option of images from mshots or e2e/puppeteer, or trying out a shadow DOM approach (I haven't used that before, but it sounds like it would have the same sandboxing effect as the iframe).

PR #71, opened 2021-04-05
Theme: Add store for pattern data, fetched from patterns API

ryelle:
See #66 , #62  — This PR adds a data store for fetching and displaying patterns from the CPT's API.

Right now the reducer only contains patterns `byId` and `queries`, but we could add a specific loading state after `FETCH_BLOCK_PATTERNS`. I've also assumed we'll add other reducers & etc here, for example for the current view or available pattern categories. My theoretical state looks something like this - this PR handles the `patterns` subtree.

```js
{
	patterns: {
		queries: {
			'': [ 1, 2, 3 ],
			'cat=2': [ 1, 3 ], // append when "load more"
			// 'cat=2&sort=new': -> not defined, so assume it's loading
		},
		byId: {
			1: { content: '<!-- wp:…' },
			2: { content: '<!-- wp:…' },
			3: { content: '<!-- wp:…' },
		},
	},
	currentQuery: 'cat=2&sort=new', // Current view
	categories: { … }, // The list of pattern categories for filtering, maybe?
	favorites: [ 3, 5 ], // If logged in, this would show the user's favorites
}
```

Adds selectors `getPattern, getPatterns, getPatternsByQuery, isLoadingPatternsByQuery`, but only `getPatternsByQuery` is used in the app now.

### Screenshots

N/A, no visual changes.

1. Run `yarn workspace wporg-pattern-directory-theme build`
2. Load the index page of patterns
3. It should show patterns whether you're logged in or not

This branch should run the JS unit test in the actions, but you could also try running those locally with `yarn workspace wporg-pattern-directory-theme test:unit`


StevenDufresne (approved):
Great PR, especially the unit tests!

PR #65, opened 2021-03-31
[WIP] Pattern Grid Menu

StevenDufresne:
This PR closes: #64 and will incorporate functionality from #50.

This PR should only be merged once it's integrated with #69 since it currently doesn't have a tie in to the interface.

### Questions
- This PR grabs the taxonomy terms (`wporg-pattern-category`)
  - Will this run into localization issues?
  - How can we control the order of how they appear?
  - Do we maybe just want to hardcore the categories for now?
-  How will we manage the "context" messaging that appears when a user is looking at a specific category?
   -  This PR just adds a temporary object (https://github.com/WordPress/pattern-directory/pull/65/files#diff-e8270c6022d0c897772f5a8dbb27343105f0f5f3d678fa1245176d6c6d17c936) 

### Notes:
- This PR currently doesn't implement the "Featured/Newest/Most Popular" dropdown which I think is the lowest priority.

### Screenshots
| Mobile | Desktop  |
| --- | --- |
| |  |

### How to test the changes in this Pull Request:
TBD

PR #72, opened 2021-04-06
Theme: Add Pattern Grid Menu.

### LINKED ISSUE #64: Pattern Directory: Users should be able to view patterns based on categories.
### Tasks
- [ ] Category bar (All, Header, Sidebar, Footer, Gallery ... )
- [ ] Search Bar
- [ ] Category dropdown ( Featured, Newest, Popular ) (Lowest priority)

This ticket will not implement the grid list. That is contained in a separate ticket: #62.

### Screenshot
patterns-signed-out-1x

StevenDufresne:
This PR replaces the experimental work done in #65.

Fixes #64, Implements UI in #50.

### Screenshots
| Mobile | Desktop  |
| --- | --- |
| |  |

1. Create a few blocks
2. Add a few categories to the Block Pattern "Category" taxonomy
3. Load `localhost:8888`


public_html/wp-content/themes/pattern-directory/css/components/_category-menu.scss
```diff
@@ -0,0 +1,89 @@
+$grid-unit: 1.5rem;
+$font-weight: 500;
+$font-size: 14px;
```
ryelle: Are these unique to this component, or can they be moved to [settings](https://github.com/WordPress/pattern-directory/tree/trunk/public_html/wp-content/themes/pattern-directory/css/settings)?

ryelle: Also, I've been trying to keep everything in `rem`s for better accessibility - so this should be:

```suggestion
$font-size: 0.875rem;
```

public_html/wp-content/themes/pattern-directory/css/components/_category-context-bar.scss
```diff
@@ -0,0 +1,72 @@
+$font-size: 13px;
+$grid-unit: 1.5rem;
+$border-radius: 2px;
```
ryelle: Same comment here — I don't think we need the `font-size` variables, and the others can be moved to settings.

public_html/wp-content/themes/pattern-directory/package.json
```diff
@@ -33,9 +32,12 @@
 		"@wordpress/data": "4.27.0",
 		"@wordpress/element": "2.20.0",
 		"@wordpress/i18n": "3.19.0",
+		"@wordpress/icons": "^2.10.0",
```
ryelle: It doesn't matter that much, but I've been pinning these to exact versions (no `^`) to avoid minor differences in envs / yarn.lock changes - it would be good to remove the caret here in these new dependencies.

public_html/wp-content/themes/pattern-directory/src/components/category-menu/default.js
```diff
@@ -0,0 +1,55 @@
...
+					>
+						{ i.label }
+					</a>
+				</li>
+			) ) }
+		</ul>
+	);
+};
+
+export default ifViewportMatches( '>= medium' )( DefaultMenu );
```
ryelle, line 56: Does this mean it exports null if `< medium`? It might be worth a comment in `category-menu/index.js` explaining this.

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid-menu/index.js
```diff
@@ -0,0 +1,82 @@
...
+	}, [ isFetching, categories ] );
+
+	useEffect( () => {
+		setCategoryContext( contextMessaging[ path ] );
+	}, [ path ] );
+
+	return (
+		<div className="pattern-grid-menu">
+			<div className="pattern-grid-menu__actions">
+				<div>
```
ryelle: This is leading to a lot of nested divs, could any of these be `<>` fragments instead?

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/index.js
```diff
@@ -21,10 +22,8 @@ function PatternGrid() {
 
 	return (
 		<>
-			<header
-				style={ { background: 'whitesmoke', padding: '24px', textAlign: 'center', margin: '0 0 24px' } }
-			>
-				Section filters
+			<header>
+				<PatternGridMenu />
```
ryelle: Looking at this in action, I think this should be a step higher (out of the `header`), and the context label (which is a header) should be the part in a header tag, while the menu should be in a `nav` - like this:

Screen Shot 2021-04-07 at 6 19 26 PM

public_html/wp-content/themes/pattern-directory/src/components/category-context-bar/index.js
```diff
@@ -0,0 +1,47 @@
...
+			setHeight( innerRef.current.offsetHeight );
+		} else {
+			setHeight( 0 );
+		}
+	}, [ message ] );
+
+	return (
+		<div className="category-context__bar" style={ { height: `${ height }px` } }>
+			<Flex align="center" ref={ innerRef }>
+				<FlexItem className="category-context__bar__copy">{ message }</FlexItem>
```
ryelle: This should be an `h2`, since it's functionally the archive title. And I believe this title should appear on all categories, regardless of whether it has related links.

ryelle (changes requested):
This is working well in my basic testing, but I have a few comments & questions about the framework.

public_html/wp-content/themes/pattern-directory/css/components/_category-menu.scss
StevenDufresne: I'll move the `$grid-unit` since it's very useful, but the other 2 are probably not useful.

StevenDufresne: Yep, i'll update the px references.

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/index.js
StevenDufresne: Yeah, good call. I got caught up in the weeds. That makes a lot more sense.

public_html/wp-content/themes/pattern-directory/src/components/category-context-bar/index.js
StevenDufresne: Yep.

public_html/wp-content/themes/pattern-directory/css/components/_category-menu.scss
StevenDufresne: In `_structure.scss` there's a `$gutter-default` property. I used that instead. It was set to `16px` so I updated it to `1.5rem` which I got from the pattern grid and since it doesn't have many references yet in the theme its a safe-ish change.
 I think we should update all gap/gutter references to use it moving forward.

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid-menu/index.js
StevenDufresne: I made some updates to remove some divs. It does make the component slightly less portable but it's a fair trade-off at this point.

public_html/wp-content/themes/pattern-directory/css/components/_category-menu.scss
ryelle: I still don't think we need local variables in these files - you're only using border-radius once, and if we want to keep the assorted font-sizes reusable let's make a set in [`settings/_typography.scss`](https://github.com/WordPress/pattern-directory/blob/trunk/public_html/wp-content/themes/pattern-directory/css/settings/_typography.scss). I'd rather not use any local variables because they aren't actually scoped to the file, if you use `$font-size` in another file, it will work, but you have no good way of knowing which font size it's using.

ryelle: Good call about the $gutter property, we can also make `$gutter-small` etc, as we need. (no change requested here, just saying -- sass math works just as well)

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid-menu/index.js
ryelle: I don't think we have immediate plans to use this whole component anywhere else (well, maybe the `CategoryMenu`), so I wouldn't worry about making it reusable yet - we can always iterate if we need to.

ryelle (approved):
Left a comment about the CSS variables, but we can leave them in for now if you want to get this merged.

public_html/wp-content/themes/pattern-directory/css/components/_category-menu.scss
StevenDufresne: > I still don't think we need local variables in these files - you're only using border-radius once, and if we want to keep the assorted font-sizes reusable let's make a set in [`settings/_typography.scss`](https://github.com/WordPress/pattern-directory/blob/trunk/public_html/wp-content/themes/pattern-directory/css/settings/_typography.scss). I'd rather not use any local variables because they aren't actually scoped to the file, if you use `$font-size` in another file, it will work, but you have no good way of knowing which font size it's using.

Very fair. I'll pull them out.

Issue #52, opened 2021-03-30
Users should be able to favorite patterns


ryelle:
Users should be able to favorite patterns from the pattern detail page, and then query for those patterns via the API. Plugins has a similar functionality, can probably look there for how to do the backend. We can create a followup issue for the proxy API `http://api.wordpress.org/patterns/`, this issue is just for updating the local endpoint.

The UI for marking favorites:

<img width="1018" alt="Screen Shot 2021-03-30 at 12 09 36 PM" src="https://user-images.githubusercontent.com/541093/113020669-ce54b080-9150-11eb-9080-b561ca2bbac6.png">

The heart animates when clicked, then turns red & the text is updated.

https://user-images.githubusercontent.com/541093/113020959-1b388700-9151-11eb-83b3-e2e74fca69c4.mov

PR #74, opened 2021-04-07
Theme: Users can copy pattern to clipboard.

### LINKED ISSUE #53: Users should be able to copy patterns
Users should be able to copy patterns. There is a copy button on both the details and list view. When clicked, the button should copy the pattern code (ex: `<!-- wp:group...`) to the user's clipboard, and show a success notice.

Screen Shot 2021-03-30 at 1 12 08 PM

If they click learn more, a modal pops up:

Screen Shot 2021-03-30 at 1 12 48 PM

[See this codepen, click Modal in the top row.](https://codepen.io/shaunandrews/pen/yLVjaWa)

StevenDufresne:
This PR allows users to copy a pattern to their clipboard.

### Key Points:
- This PR tries to use `@wordpress/components` as much as possible even though this slightly deviates from the design.
  - `Guide`, `Button`, `Notice`
- This PR leverages another [`render`](https://github.com/WordPress/pattern-directory/compare/update/copy-to-clipboard?expand=1#diff-6897a946f493ca3f66c4f3760d29fa084ffc6d48bfa634fe52566c36e63edebaR34) to integrate into the view.
  - We may want to consider having only one `render` for the whole preview...? Not sure what the pros & cons would be yet.
- This PR uses a hidden input to retrieve the pattern to copy to the clipboard (probably temporarily). 
  - We could pass it in the same way we do it for `.pattern-preview__container`, 
  - .. maybe it's retrievable from the state once we get that integrated.
 - This PR introduces an `AddToFavoriteButton` which is incomplete.

Fixes #53

### Screenshots

_Note: The add to favorites button needs updating. See #52._

| Mobile | Desktop |
| --- | --- |
|  |   |

1. Create a block pattern
2. Visit `localhost:8888`
3. Click on the block pattern in the grid
4. Click on the `Copy Pattern` button
5. Verify that a success message is presented
6. Paste the contents of your clipboard somewhere to verify the pattern was copied
7. Click the 'Learn more' in the success message
8. Verify a modal window opens
9. Close the window.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
@@ -12,9 +12,13 @@ body.single-wporg-pattern {
 
 	.entry-header {
 		max-width: $size__site-main;
-		padding: 2rem 0;
+		padding: 1.5rem;
 		margin-left: auto;
 		margin-right: auto;
+
+		@media only screen and (min-width: $breakpoint-small) {
+			padding: 1.5rem 0;
+		}
 
 		.entry-title {
 			margin-top: 0;
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
@@ -62,10 +62,22 @@ body.single-wporg-pattern {
 
 			&-content {
 				padding: 1.5rem;
+				line-height: 1.5;
 			}
 
 			&-title {
 				margin: 0;
+			}
+
+			&-shortcut {
+				background: $color-gray-light-300;
+				box-shadow: 0 0 0 1px $color-gray-light-700;
+				padding: 1px;
+				font-weight: 600;
+				min-width: 20px;
+				display: inline-block;
+				text-align: center;
+				border-radius: 2px;
 			}
 		}
 	}
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
@@ -16,7 +16,7 @@ body.single-wporg-pattern {
 		margin-left: auto;
 		margin-right: auto;
 
-		@media only screen and (min-width: $breakpoint-small) {
+		@media only screen and (min-width: $breakpoint-large) {
 			padding: 1.5rem 0;
 		}
 
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
@@ -26,6 +26,7 @@ body.single-wporg-pattern {
 		display: flex;
 		align-items: stretch;
 		flex-wrap: wrap;
+		overflow: hidden;
 
 		button + button {
 			margin-left: 2em;
@@ -34,6 +35,22 @@ body.single-wporg-pattern {
 		&__notice {
 			margin: 1.5rem 0 0;
 			flex-basis: 100%;
+			transform: translateY(0);
+			transition: transform 0.15s ease-in-out;
+
+			// Gutenberg wraps content
+			> * {
+				display: flex;
+				margin: 0;
+				justify-content: space-between;
+				align-items: center;
+			}
+
+			&--is-hidden {
+				margin: 0;
+				flex-basis: auto;
+				transform: translateY(60px);
+			}
 		}
 	}
 
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
@@ -25,9 +25,15 @@ body.single-wporg-pattern {
 	.pattern-actions {
 		display: flex;
 		align-items: stretch;
+		flex-wrap: wrap;
 
 		button + button {
 			margin-left: 2em;
+		}
+
+		&__notice {
+			margin: 1.5rem 0 0;
+			flex-basis: 100%;
 		}
 	}
 
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
@@ -30,7 +30,6 @@ body.single-wporg-pattern {
 		display: flex;
 		align-items: stretch;
 		flex-wrap: wrap;
-		overflow: hidden;
 
 		button + button {
 			margin-left: 2em;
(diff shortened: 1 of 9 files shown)


public_html/wp-content/themes/pattern-directory/src/components/pattern-preview-actions/copy-guide.js
```diff
@@ -0,0 +1,81 @@
...
+ */
+import { __ } from '@wordpress/i18n';
+import { Animate, Guide } from '@wordpress/components';
+import { createInterpolateElement } from '@wordpress/element';
+
+const CopyPasteImage = () => (
+	// Wrap the image to avoid the UI shift after the GIF loads
+	<div style={ { height: '220px' } }>
+		<img
+			src="https://shaunandrews.com/wp-content/uploads/2021/03/pattern-paste-demo-2.gif"
```
StevenDufresne: How do we want to ship images? Should they be external to the project somehow?

ryelle: We can include images like this into the theme, I believe they're proxied into the CDN on production.

StevenDufresne: ... added an `images` folder in the theme

public_html/wp-content/themes/pattern-directory/src/components/pattern-preview-actions/copy-guide.js
```diff
@@ -0,0 +1,81 @@
...
+);
+
+const CopyGuide = ( { onFinish } ) => {
+	return (
+		<Animate type="appear" options={ { origin: 'bottom' } }>
+			{ ( { className } ) => (
+				<Guide
+					className={ `pattern-actions__guide ${ className }` }
+					onFinish={ onFinish }
+					finishButtonText={ __( 'Close', 'wporg-patterns' ) }
```
ryelle, line 25: Is this a required prop, or can we get rid of this extra close button? There are a lot of other ways to close this modal, so this feels redundant.

public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
```diff
@@ -25,10 +29,77 @@ body.single-wporg-pattern {
...
+		}
+
+		&__guide {
+			max-width: $breakpoint-mobile / 1.25;
+
+			// @wordpress/components/guide overrides
+			height: auto;
+			max-height: none;
+
+			&-content {
```
ryelle: I know these `&{string}` shortcuts are convenient but they make these selectors hard to read, especially here - you need to backtrack 2 levels to figure out the actual class name. Can we keep `&` out of selector strings? Nesting would still be okay, like:

```scss
.foo {
  .bar & { // OK
    color: blue;
  }
  &-bar { // Hard to read
    color: red;
  }
}
```

public_html/wp-content/themes/pattern-directory/src/components/pattern-preview-actions/copy-guide.js
```diff
@@ -0,0 +1,81 @@
...
+										</li>
+										<li>
+											<p>
+												{ createInterpolateElement(
+													__(
+														'Paste the contents of your clipboard by pressing both the <span>⌘</span> and <span>v</span> keys, or right-clicking and choose "Paste" from the menu.',
+														'wporg-patterns'
+													),
+													{
+														span: <span className="pattern-actions__guide-shortcut" />,
```
ryelle: Could you use the `kbd` element here?

public_html/wp-content/themes/pattern-directory/src/components/pattern-preview-actions/success-message.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Button, Notice } from '@wordpress/components';
+
+const SuccessMessage = ( { showMessage, onClick } ) => (
```
ryelle: Can this be renamed to `CopySuccessMessage` or similar? I thought it was generic at first, so a more specific name would be clearer.

ryelle (changes requested):
Could you rebase this to get it up to date, and maybe get it working without the animations to start? The notice is always visible for screen reader users since it's technically on-screen, I think it would work better to have it render in on click. I'm also not sure the notice really _needs_ the scroll-up animation, TBH.

This focus style is a bit off… I think from the `overflow:hidden`?

Screen Shot 2021-04-19 at 1 25 24 PM

> We may want to consider having only one render for the whole preview...?

Yeah, it's worth thinking about - we could have one render that builds this whole section (maybe called `PatternDetails`? just thinking aloud), since the report flow will also need a modal. If we do that, we can use the `blockContent` parsed out from the preview container. I wouldn't say this change _blocks_ this PR, but it wouldn't hurt to do that here.

Screen Shot 2021-04-19 at 2 12 16 PM

public_html/wp-content/themes/pattern-directory/src/components/pattern-preview-actions/success-message.js
StevenDufresne: Yep!

public_html/wp-content/themes/pattern-directory/src/components/pattern-preview-actions/copy-guide.js
StevenDufresne: Yep!

public_html/wp-content/themes/pattern-directory/css/components/_pattern.scss
StevenDufresne: That's fair. Done.

public_html/wp-content/themes/pattern-directory/src/components/pattern-preview-actions/copy-guide.js
StevenDufresne: I made use of the `Guide` component because it's structurally the same as the design and it's quick to get up and running. However, since it does expect a sequence of user interactions, there is always a button on the last page. https://github.com/WordPress/gutenberg/blob/4ae96fb2867b28a5c8ecb2847e74f99c9904a909/packages/components/src/guide/index.js#L46. The `finishButtonText` allows you to override the default text which is by default `Finish`.

We could alternately build our own modal, but of me, the extra button to close is not a deal-breaker and does help mouse-driven users... who could potentially have a hard time finding the "x" button overlayed on the image.

StevenDufresne:
> > We may want to consider having only one render for the whole preview...?

> Yeah, it's worth thinking about - we could have one render that builds this whole section (maybe called `PatternDetails`? just thinking aloud), since the report flow will also need a modal. If we do that, we can use the `blockContent` parsed out from the preview container. I wouldn't say this change _blocks_ this PR, but it wouldn't hurt to do that here.

Can we address it in a future PR? As you mention, I think pushing that decision into the future will help us make a more accurate change.

ryelle (approved):
the render thing doesn't block this, thanks for updating it! Looks good now 👍🏻

Issue #63, opened 2021-03-31
List table improvements for the pattern post type


coreymckrill:
The pattern list table will primarily be used by people who are reviewing or moderating patterns that have been submitted by users on the front end. Thus the table columns and filters that are present should be optimized for reviewing/moderating tasks.

### Columns

* Title (Pattern name)
* Author (User name of pattern creator)
* Date (Date published, or last modified + status)
* Flags (#61)
* Categories?
* Keywords?

### Filters

Besides the defaults:

* View by has flags
* Sort by number of flags

### Quick Actions & Bulk Edit

When a moderator changes a pattern, they should include a reason so that it's easier to understand later why something changed (#57). So ideally, when a quick action or bulk edit is modifying patterns, it should trigger a field to appear in a modal or something where a comment about the change can be entered.

coreymckrill:
I split the "reason field" part of this out into a separate issue #105. The rest of this is done.

PR #75, opened 2021-04-07
Directory: Register custom post statuses

### LINKED ISSUE #56: Pattern statuses
* `draft` The pattern has not been submitted to the directory yet, or it was submitted but failed automated validation checks. See #36 
* `awaiting-review` (or maybe just use `pending`?) The automated checks determined that a human review is also needed. This could also be used when a published pattern gets flagged as inappropriate.
* `declined` The pattern was declined by a human reviewer. See #16 
* `publish` The pattern passed automated checks and/or human review. It is now published in the pattern directory and appears in searches via the API.
* `removed` The pattern was previously published, but was removed either by the author or a moderator. See #16 

There is also some discussion in #16 of having patterns that are only available to the pattern author until the author chooses to publish them. So an additional status might be:

* `private` The pattern has been submitted, and passed automated checks, but has not been published by the author.

coreymckrill:
This simply registers the two post statuses to be used for patterns that aren't already built-in. There could be need in the future for more functionality around these statuses, but we can build that as needed in separate PRs.

Fixes #56

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/bootstrap.php
+++ b/public_html/wp-content/plugins/pattern-directory/bootstrap.php
@@ -10,4 +10,5 @@ namespace WordPressdotorg\Pattern_Directory;
 
 require_once __DIR__ . '/includes/pattern-post-type.php';
 require_once __DIR__ . '/includes/pattern-validation.php';
+require_once __DIR__ . '/includes/post-status.php';
 require_once __DIR__ . '/includes/search.php';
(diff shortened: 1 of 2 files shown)


coreymckrill:
I guess we need some UI to be able to use these custom statuses. Some of it would be covered by #63, but we also probably need something in the block editor UI to be able to switch between the statuses listed in #56. That would need design, though, so perhaps it's better to address it in a separate issue/PR anyway...

ryelle (approved):
Looks good! Agreed that any UI can happen in another PR.

Issue #73, opened 2021-04-06
Retrieving pattern grid category filters.


StevenDufresne:
We register a [`wporg-pattern-category`](https://github.com/WordPress/pattern-directory/blob/456b6a2bf0f03322fa1bfd74fcf4e310488306e8/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php#L57) which I assume will be assigned to block patterns and used as the sorting mechanism for the pattern grid menu here: #48.

Questions:
- Is that assumption correct? Will we be using those categories?
  - How will we manage localization? 
  - How can we control the order of how they appear?
- How will we manage the "context" messaging that appears when a user is looking at a specific category?
  - This PR just adds a temporary object (https://github.com/WordPress/pattern-directory/pull/65/files#diff-e8270c6022d0c897772f5a8dbb27343105f0f5f3d678fa1245176d6c6d17c936)
    - We'll want to make it more dynamic.

We could alternatively hardcode the items for now.

Related PR: #72

ryelle:
> Is that assumption correct? Will we be using those categories? 

That is correct, those filters correspond to the categories. We'll launch with a set of categories, and users won't be able create new ones, only tag their patterns with the existing ones. I think it's likely we'll want to add new categories as the site grows, so I don't want to rely on a manual mapping if possible otherwise we'll always need a dev to update things.

I suppose we could add a "simple" ui to the term screen to manage up to 3 related terms, though?

I think @iandunn has some ideas here (https://github.com/WordPress/pattern-directory/issues/48#issuecomment-810418852) for more dynamic relationships.

I'd _assume_ that any related categories would come in as part of the term API response (for example, a field added by `register_rest_field` that lists an array of related term IDs or something).

> How will we manage localization?

Good question - maybe something like the Learn work that was just done? The logged out "create" page will also need handling for i18n, since it will be written in the editor too. I think our initial launch will just be in English, though.

> How will we manage the "context" messaging that appears when a user is looking at a specific category? 

The `45 <b>Header</b> patterns`? We should be able to build that dynamically with each page, using the total found posts + current filter name. We might need to tweak the data store to save the found posts total with the query results, though.

StevenDufresne:
Some of these suggestions have been added to #76 and #78. We can move these conversations to the open PRs and future ones. Closing for now.

PR #76, opened 2021-04-08
Add an ordinal to the category taxonomy.

StevenDufresne:
This PR adds a `display-order` property to `wporg-pattern-category` taxonomy so we can control the order of categories listed in the Pattern Grid Menu (Visible here: #48, being implemented here: #72). 

See #73.

**Example rest response:** 
```
[{
  "id":12,
  "count":0,
  "name":"Footer",
  "slug":"footer",
  "taxonomy":"wporg-pattern-category",
  "parent":0,
  "meta":[],
  "display_order":1,
...
}]
```

### Screenshots
| Add/View | Edit |
| --- | --- |
|   |    |

1. Visit `/wp-admin/edit-tags.php?taxonomy=wporg-pattern-category&post_type=wporg-pattern`
2. Add a category, include a numerical value in the `Display Order` field
3. Click `Add new category`
4. Verify that it shows up in the category table column `Display Order`.
5. Click on the category in the table
6. Verify that it shows up in the `Display Order` field on the edit category page.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-taxonomy-modifications.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-taxonomy-modifications.php
@@ -60,10 +60,11 @@ function edit_ordering_form_field( $term ) {
  * @param integer $term_id Term id.
  */
 function save_term_field( $term_id ) {
+	/* phpcs:disable WordPress.Security.NonceVerification.Missing */
 	update_term_meta(
 		$term_id,
 		CATEGORY_TAXONOMY_ID,
-		sanitize_text_field( $_POST[ CATEGORY_TAXONOMY_ID ] )
+		absint( wp_unslash( $_POST[ CATEGORY_TAXONOMY_ID ] ) )
 	);
 }
 
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-taxonomy-modifications.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-taxonomy-modifications.php
@@ -21,7 +21,7 @@ add_filter( "rest_prepare_{$CATEGORY_TAXONOMY}", __NAMESPACE__ . '\rest_prepare_
 function add_ordering_form_field() {
 	echo '
 		<div class="form-field">
-			<label for="category-order">Order</label>
+			<label for="category-order">Display Order</label>
 			<input type="number" min="1" name="category-order" id="category-order" />
 		</div>
 	';
@@ -40,7 +40,7 @@ function edit_ordering_form_field( $term ) {
 	echo '
 		<tr class="form-field">
 			<th>
-				<label for="category-order">Order</label>
+				<label for="category-order">Display Order</label>
 			</th>
 			<td>
 				<input name="category-order" min="1" id="category-order" type="number" value="' . esc_attr( $value ) . '" />
@@ -76,7 +76,7 @@ function save_term_field( $term_id ) {
  */
 function add_order_table_heading( $columns ) {
 	// Our our custom column heading
-	$columns[ CATEGORY_TAXONOMY_ID ] = __( 'Order' );
+	$columns[ CATEGORY_TAXONOMY_ID ] = __( 'Display Order' );
 
 	return $columns;
 }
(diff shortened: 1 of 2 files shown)


public_html/wp-content/plugins/pattern-directory/includes/pattern-taxonomy-modifications.php
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Saves the custom taxonomy field to the database.
+ *
+ * This is a callback for the `created_{$taxonomy}` and `edited_{$taxonomy}` filter, and it's used to save meta information to the database.
+ *
+ * @param integer $term_id Term id.
+ */
+function save_term_field( $term_id ) {
+	/* phpcs:disable WordPress.Security.NonceVerification.Missing */
```
StevenDufresne, line 63: I had to disable this rule. 

@iandunn Is it safe to assume we don't need to verify the nonce here?

iandunn:
🤔  is this for how the 3 `related categories` are sorted? If so, what led you to choosing this approach over a simpler/automatic method (like alphabetical)?

What should happens if two terms have the same order? I'm guessing we'll want to avoid having to manually update each term when a new one is inserted or changed.

public_html/wp-content/plugins/pattern-directory/includes/pattern-taxonomy-modifications.php
iandunn: I think [the best practice is to use a nonce](https://wordpress.stackexchange.com/questions/49453/do-i-need-a-nonce-field-for-every-meta-box-i-add-to-my-custom-post-type-admin). It won't be necessary in all cases, but can be sometimes. It also makes security review easier, similar to late escaping.

Similarly, it should probably also use `current_user_can()` to make sure the user submitting the request is allowed to edit terms for this taxonomy.

StevenDufresne:
> 🤔 is this for how the 3 `related categories` are sorted? If so, what led you to choosing this approach over a simpler/automatic method (like alphabetical)?

Not the 3 `related categories` but the navigation itself. I think the 3 related ones could be sorted alphabetically.



> What should happens if two terms have the same order? I'm guessing we'll want to avoid having to manually update each term when a new one is inserted or changed.

If they have the same order they will be placed beside each other. I expect us to only need to manage the order for ~5/6 to make sure they show up on the 'left' of the menu and don't get sliced out should we introduced a maximum length.

If this PR makes sense and that becomes a pain, we could write something more in-depth. Additionally, I was thinking a new page would definitely be more work at this point.

iandunn:
> Not the 3 related categories but the navigation itself

Ah, I missed that, thanks! How important is it to have a hardcoded order, there? It make the code and UI much simpler if it could just be alphabetical, or maybe "popularity" according to the # of assigned posts.

If hardcoding is desired, though, that WFM too.

Given it's just a few items, I agree it's not worth it to handle the case where two have the same value.

StevenDufresne:
> Ah, I missed that, thanks! How important is it to have a hardcoded order, there? It make the code and UI much simpler if it could just be alphabetical, or maybe "popularity" according to the # of assigned posts.

That's a good question and it's probably a bit early to determine, isn't it? Alphabetical is probably an easier start. I'll leave the PR hanging for a bit and if we don't find the need we'll close it. Thanks for the look-see!

iandunn:
👍🏻

shaunandrews:
I'm not sure about the implementation, but I expect that we _will_ want to be able to curate the list of categories we feature in that top area next to search. To start, we can try to match the list of categories that ships in Gutenberg:

image

And/or take some inspiration from WordPress.com:

image

--

I expect we can show an overflow menu as our list of categories grow, and as a way to allow the list to shrink as the viewport shrinks:

image

ryelle:
+1 to pulling out the categories used in gutenberg as the main navigation (minus the theme-generated ones). We'll need to do some content entry on the site to create these pattern categories, and can hardcode the category slugs for now. I think we could sort both sets alphabetically, until we get patterns submitted and can look at what the popularity is like.

I'm not sure if we need a "view all categories" button/page, since the categories list will be admin-curated and probably won't be more than a few dozen, if that.

ryelle:
@StevenDufresne I think we decided this functionality isn't necessary for launch, what do you think about closing this PR?

StevenDufresne:
Yep!

Issue #61, opened 2021-03-30
Flagging a pattern for review


coreymckrill:
This is a work in progress. Requirements listed here are up for discussion.

A pattern flagging system will have a few different components:

* Front end UI for flagging a pattern in the directory
* A REST API endpoint to submit the data from the UI
* A schema for storing flags for a pattern in a structured way
* Back end UI for surfacing flags to pattern moderators

### Front end UI

The pattern detail page will have a “Report pattern” link which opens a modal with some options to report a pattern:

<img width="414" alt="image-15" src="https://user-images.githubusercontent.com/916023/113040728-fa226700-914d-11eb-97fc-aff27c38b0f7.png">

Rather than a separate "Other" field, it might be worth considering requiring a comment regardless of which reason is selected. I.e. "Trademark Issue" isn't that helpful on its own without additional context provided.

A user should only be able to flag a particular pattern once. If they have already flagged it, the "Report pattern" link should reflect that.

### REST API endpoint

* The endpoint might look something like `wp/v2/wporg-pattern/(?P<id>[\d]+)/flag`
* A request to the endpoint should include the pattern's post ID and a text string reason for the flag.
* In order to flag a pattern, the user should be required to be logged in to their wp.org account so the user name can also be attached to the flag.

### Flag schema

* Pattern ID
* Date
* User name
* Reason
* Maybe a flag status? _So flags can be differentiated by whether they have been reviewed yet or not._

There are a few different ways flags could be stored:

* As a separate post type, with the pattern post as a parent
* As comments on the pattern post
* As meta fields on the pattern post (one entry would contain all the flag data)

Meta fields would be the easiest to initially implement, but would be more difficult to work with later if we want to be able to query by certain schema fields, e.g. is one user flagging a lot of patterns?

A separate post type would probably be the most flexible to work with, since all of the schema fields would fit into existing post fields and it would have great query support.

### Back end UI

* The number of pending flags on a pattern could show in a column on the pattern list table, perhaps similar to pending comments on a post.
* Clicking the number could go to a separate list table showing all pending flags for a specific pattern post.
* A row in the list table could have quick actions to view the pattern and mark the flag as reviewed. Unpublishing a pattern should probably not be a quick action, because we would want the moderator to include a comment/reason when changing a pattern's status (#57)

coreymckrill:
The "Front end UI" part of this has been pulled out in #80, and the rest of this was addressed in #77.

PR #77, opened 2021-04-09
Directory: Add a system for flagging patterns

coreymckrill:
This adds the REST API endpoint, flag schema, and back-end UI described in #61. The front end UI will probably get a separate PR, and maybe a separate issue as well.

This introduces the `wporg-pattern-flag` post type, the `wporg-pattern-flag-reason` taxonomy, and the `resolved` post status to store and manage data when a pattern in the pattern directory is flagged for review. A flag is comprised of:

* A reason for the flag, stored as a term in the `wporg-pattern-flag-reason` taxonomy. There will be a few pre-defined terms that a user can choose from on the front end as the reason for the flag.
* Details about the flag, stored in `post_excerpt`. This provides more context for the moderator reviewing the pattern and its flags.
* The ID of the "parent" pattern post, stored in `post_parent`
* The ID of the user who submitted the flag, stored in `post_author`.
* The status of the flag, stored in `post_status`. A new flag starts as `pending` and can be marked as `resolved` once it has been reviewed.

The still-in-the-works front end form will submit the flag data to the REST API endpoint, which is a slightly modified version of the standard endpoint for WP post types. The main differences are:

* To view flags, you must have edit permissions for the `wporg-pattern` post type.
* To create flags, you don't need edit permissions, but you must be logged into your wp.org account.
* There are checks to prevent submitting a flag to a pattern that isn't published, and to prevent submitting a flag to a pattern that you've already flagged previously, regardless of the status of previous flags.

For the back-end UI, this PR focuses on screens directly related to the new post type and taxonomy, namely a list table and a UI for creating new terms. Adding info about a pattern's flags to the patterns list table will be covered in a separate issue/PR.

The list table is a standard posts list table that has been modified via filters and actions. Of note are the custom columns, row actions, and bulk actions to facilitate reviewing flags.

### Screenshot

Screen Shot 2021-04-16 at 11 50 03

**Setup**

* Create one or more block patterns.
* Create one or more terms on the Reasons subpage.

**Test the API**

* This is most easily done with a client app like [Insomnia](https://insomnia.rest/).
* To create a new flag, send a POST request to `http://localhost:8888/wp-json/wp/v2/wporg-pattern-flag`. It needs to have the following parameters:
  * `parent`: The post ID of a pattern post
  * `wporg-pattern-flag-reason`: The ID of a Reason term
  * `excerpt`: A "details" string
* To get existing flags, send a GET request to `http://localhost:8888/wp-json/wp/v2/wporg-pattern-flag` with the usual optional parameters for any kind of posts.

**Test the back end UI**

* Once you have some flags created, point your browser to `http://localhost:8888/wp-admin/edit.php?post_type=wporg-pattern-flag`.
* Try resolving/unresolving and trashing/untrashing flags with row actions (links that appear when you hover over a flag row), and bulk actions.
* Try viewing only flags for one particular pattern.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/class-rest-flags-controller.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/class-rest-flags-controller.php
@@ -159,8 +159,8 @@ class REST_Flags_Controller extends WP_REST_Posts_Controller {
 		$flag_check = new WP_Query( array(
 			'post_type'   => $this->post_type,
 			'post_parent' => $parent->ID,
-			'post_status' => 'any',
-			'author' => get_current_user_id(),
+			'post_status' => 'pending',
+			'author'      => get_current_user_id(),
 		) );
 		if ( $flag_check->found_posts > 0 ) {
 			return new WP_Error(
--- a/public_html/wp-content/plugins/pattern-directory/includes/class-rest-flags-controller.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/class-rest-flags-controller.php
@@ -191,6 +191,16 @@ class REST_Flags_Controller extends WP_REST_Posts_Controller {
 			$prepared_post->post_status = $schema['properties']['status']['default'];
 		}
 
+		foreach ( $request['wporg-pattern-flag-reason'] as $term_id ) {
+			if ( ! term_exists( $term_id, FLAG_TAX ) ) {
+				return new WP_Error(
+					'rest_invalid_term_id',
+					__( 'Invalid term ID.', 'wporg-patterns' ),
+					array( 'status' => 400 )
+				);
+			}
+		}
+
 		return $prepared_post;
 	}
 
@@ -211,6 +221,8 @@ class REST_Flags_Controller extends WP_REST_Posts_Controller {
 			'context'     => array( 'view', 'edit' ),
 			'required'    => true,
 		);
+
+		$schema['properties']['wporg-pattern-flag-reason']['required'] = true;
 
 		return $schema;
 	}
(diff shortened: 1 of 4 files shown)


public_html/wp-content/plugins/pattern-directory/includes/class-rest-flags-controller.php
```diff
@@ -0,0 +1,273 @@
...
+				array( 'status' => 403 )
+			);
+		}
+
+		// Check if the user has already submitted a flag for the pattern.
+		$flag_check = new WP_Query( array(
+			'post_type'   => $this->post_type,
+			'post_parent' => $parent->ID,
+			'post_status' => 'any',
+			'post_author' => get_current_user_id(),
```
ryelle:
```suggestion
			'author' => get_current_user_id(),
```
Confusingly, this should just be `'author'` — otherwise it blocks reporting any pattern that has any existing flag.

ryelle (approved):
💯 This looks great! You've gotten really good at wrangling list tables 😄

> prevent submitting a flag to a pattern that you've already flagged previously

What if a pattern is flagged for a different reason, or it reverts after an update? I'm assuming this is to prevent spam/abusive flagging, but maybe that can just check for existing pending flags?

I noticed that if you pass in an invalid reason, the flag still succeeds - is that expected?

public_html/wp-content/plugins/pattern-directory/includes/class-rest-flags-controller.php
coreymckrill: Ah good catch

coreymckrill:
> What if a pattern is flagged for a different reason, or it reverts after an update? I'm assuming this is to prevent spam/abusive flagging, but maybe that can just check for existing pending flags?

Changed in 2c59651 to only count pending flags. I was sort of thinking that patterns wouldn't be editable after they are published, but if they are, then it makes sense to only query pending.

> I noticed that if you pass in an invalid reason, the flag still succeeds - is that expected?

Ah nope, good catch. Fixed in d4700d3

PR #78, opened 2021-04-12
Theme: Connect `PatternGridMenu` to `PatternGrid`

StevenDufresne:
This PR connects the `PatternGridMenu` to the `PatternGrid` allowing users to view patterns by categories.

I had hoped to have more atomic Pull Requests for some of the components & features introduced in this PR but found myself often blocked by missing pieces. 

### Key Updates:
- Introduces a `RouteProvider` to be used to set & follow path changes.
- Introduces a `Pattern` component to wrap `<PatternGridMenu>` & `<PatternGrid>` inside the `RouteProvider`
- Moves category-related state from the `PatternGridMenu` to the shared state introduced in #71.
- Adds a `QueryMonitor` that responds to updates to the browser's path, saves the new query in the state which triggers a new api request.

### Screenshots

_With Chrome Developer tools "Fast 3G" throttling._

Clicking through the menu items

1. Visit `/wp-admin/edit-tags.php?taxonomy=wporg-pattern-category&post_type=wporg-pattern`.
2. Create a few categories.
3. Create a few block patterns and associate them with the categories.
4. Visit `/`.
5. Expect to see the categories on the page.
6. Click on the categories and expect to see the associated block pattern.


StevenDufresne:
### Question 1
At this point, the `PatternGridMenu` shows up when viewing specific categories (ie `/pattern-categories/footer/`). Users can arrive on a specific category page if they click one of the categories and reload the page. It happens because we use the `link` property associated with the category from the REST API [in the nav menu](https://github.com/WordPress/pattern-directory/pull/78/files#diff-3ef39beb357b0b405e7570dde3383afd76097b3ef1f7a1cc9d8d814babd67c2fR36). 

**Example:**



#### Potential Solutions

**1. URL rewrite**
It may make sense to change that URL to something like `/patterns/footer/` and if they reload that page, we load the default `/` view using a url rewrite with the appropriate category selected. 

If users navigate to `/pattern-categories/footer/` from somewhere else in the theme we can load the `PatternGrid` component but not the `PatternGridMenu`. That would mimic the [WordPress.org Plugins tag archive page](https://wordpress.org/plugins/tags/conditional-blocks/).

**2. Load as if it's /**
We could also just load the view as if they visited `/`.

StevenDufresne:
## Question 2
How do we want to handle searches? The challenge is this, when a user is viewing a specific category and searches for `button`, do we show buttons that match the specific category, or do we search all patterns?

## 1. We search only within the category (This PR used this approach -- See GIF)
We can filter the current category and alert the user using a message like:

**Viewing Footer**: `3 footer patterns matching "button".`
**Viewing All**: `12 patterns matching "button".`

**Good**:
- Users have the ability to drill down within categories
- As users click on new categories, those categories maintain the keyword search and the UI stays consistent.
- We don't need a new view, or need to modify the category navigation

**Not so good**:
- .ORG usually searches all items


## 2. Search all patterns
We could alternatively update the view to show all the patterns. 

**Good**:
- Follows .org search functionality
- A user may want to see all the patterns regardless of current category?

**Not so good**:
- It would require an alternate/modified view.
  - How do we exit the search state?
    - Do users have to manually clear the keyword to back out? If they don't have to, does the keyword stay in the search bar but not affect what they see?

public_html/wp-content/themes/pattern-directory/src/components/category-context-bar/index.js
```diff
@@ -1,30 +1,82 @@
...
+		if ( searchTerm ) {
+			count = patterns.length;
+		}
+
+		if ( ! isAllCategory || searchTerm ) {
+			_context = getContextMessage( count, category.name, searchTerm );
+		}
+
+		setContext( _context );
+	}, [ isLoadingPatterns, patterns ] );
```
ryelle: Should this also depend on `category`?

public_html/wp-content/themes/pattern-directory/src/store/test/selectors.js
```diff
@@ -90,4 +91,103 @@ describe( 'selectors', () => {
...
+		} );
+	} );
+
+	describe( 'isLoadingCategories', () => {
+		it( 'should get true when state is null', () => {
+			const categoryState = {
+				categories: null,
+			};
+
+			expect( isLoadingCategories( categoryState ) ).toBeTruthy();
```
ryelle: Is there a reason to not check for `.toBe( true )`? Since we probably know what these values are, we might as well check for the real values. (same for the `toBeFalsy` checks)

ryelle (approved):
This is working great! Category filtering and searching are both working in my tests.

Both of your questions raise good points, but I don't think they need to block merging this PR — maybe bring them up in the list view tracking issue #48, and tag it for design? I think for Question 1 we could tweak the logic in `header.php` to show the large header on any archive view too, which should keep it in place for all category & search pages.

public_html/wp-content/themes/pattern-directory/src/store/test/selectors.js
StevenDufresne: No, that's a sneaky bad habit. Thanks for calling it out.

public_html/wp-content/themes/pattern-directory/src/components/category-context-bar/index.js
StevenDufresne: Good observation.

Technically, since patterns are not queried until we have built the query and saved it to the state, and building the query requires that categories have already loaded... `category` _should_ always be defined before patterns, at least based on the fact that we load this component immediately. With those assumptions in place, we can save a few renders by not watching the `category` updates. 

Should the component be conditionally rendered (or its parent) it would most likely need to observe `category` changes. Now, the likelihood that this component will ever contain children that could have expensive renders is very low, and thinking in terms of performance versus portability, it may make more sense for us to additionally track `category` to protect against upstream re-structuring. 

Lastly, and most importantly, I've totally overthought this 🤪.

PR #83, opened 2021-04-23
Theme: Add a window to report pattern.

### LINKED ISSUE #80: Theme: "Report this pattern" button and form
This issue is pulling out the **Front End UI** section of #61, and building off the work done in #77.

When viewing a single pattern in the directory, there should be a "Report this pattern" button to click in the case that the pattern is problematic for some reason. Clicking the button should bring up a modal similar to this mockup:

image-15

I would suggest that instead of a textarea that only applies to the "Other" category, the textarea should always be present beneath the radio button group, labeled "Details", and perhaps always be a required input. It seems that no matter which reason is chosen for reporting a pattern, more context will be necessary for a moderator to effectively review the report.

**Further requirements:**

* The options in the radio button group for "Please choose a reason" will be drawn from terms in the `wporg-pattern-flag-reason` taxonomy.
* When the form in the modal is submitted, it should send a `POST` request to `/wp-json/wp/v2/wporg-pattern-flag`. The request must have the following parameters:
  * `parent` The ID of the pattern post
  * `wporg-pattern-flag-reason` The ID of the "reason" taxonomy term
  * `excerpt` The user-input string of text providing details about the problem with the pattern
* The API will send back an error in some cases, which we may want to surface to the end user. Some examples of relevant errors:
  * `You have already flagged this pattern.`
  * `Flags cannot be submitted for this pattern.` (if it is not published, for example)
  * `You must be logged in to submit a flag.`
* Perhaps the "Report this pattern" button shouldn't be shown unless the user is logged in? Or perhaps if they aren't logged in, the modal should instead show an explanatory message and a link to log in that will then bring them back to the pattern page afterwards?
* Ideally the "Report this pattern" button would be aware if the current user has already submitted a flag for the pattern, and would show with a checkmark or something, as well as being unclickable.

StevenDufresne:
This PR allows users to report a pattern.

Closes #80

### Screenshots
| State | Screenshot |
| --- | --- |
| Not logged in. <br> _This currently just links to the login page._ |  |
| Logged in with no existing report |  |
| Using Browser validation |  |
| Confirmation |  |
| Logged in with existing report |  |
| Error Handling (although we don't allow users to open the window if they already submitted) |  | 
| Flow |  

**Logged Out**
- Create a block pattern.
- Create a few flags.
- Navigate to single pattern view.
- Notice there is a link to log in.

**Logged In**
- Create a block pattern.
- Create a few flags.
- Navigate to single pattern view.
- Click "Report Link"
- Fill in form details
- Click report
- Notice a confirmation message
- Close confirmation window
- Notice the text has changed where it used to say "Report this pattern".


public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
```diff
@@ -9,7 +9,25 @@
...
+	$args = array(
+		'author' => get_current_user_id(),
+		'post_parent' => get_the_ID(),
+		'post_type' => POST_TYPE,
+		'post_status' => 'any'
+	);
+	
+	$items = new \WP_Query( $args );
+	$userHasReported = $items->have_posts();
+}
```
StevenDufresne: I'm not sure where to place this code? @coreymckrill @ryelle @iandunn Should this be moved to a class in the plugin as opposed to having it in the theme? 

I think the Plugin Directory has a default class and is called everywhere for this like this.

public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
```diff
@@ -9,7 +9,25 @@
...
 get_header();
+
+$user_has_reported = false;
+
+if ( is_user_logged_in() ) {
+	$args = array(
+		'author' => get_current_user_id(),
+		'post_parent' => get_the_ID(),
+		'post_type' => POST_TYPE,
+		'post_status' => 'any',
```
coreymckrill:
```suggestion
		'post_status' => 'pending',
```

I think this should probably just check for flags that haven't been reviewed yet.

public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
coreymckrill: I can't think of a use case for this on the plugin side. But this could be moved to a function and put into the theme's **functions.php** file.

public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
```diff
@@ -44,7 +62,12 @@
...
 							?>
 						</div>
-						<div class="pattern__report">
+						<div id="pattern-report"
+							class="pattern__report"
+							data-post-id="<?php echo intval( get_the_ID() ); ?>"
+							data-logged-in="<?php echo intval( is_user_logged_in() ); ?>"
+							data-user-has-reported="<?php echo intval( $user_has_reported ); ?>"
+							">
 							<button class="button">Report this pattern</button>
```
coreymckrill, line 56: For some reason this is showing as "You've reported this pattern" even when `data-user-has-reported` is set to `0`.

public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
StevenDufresne: A potentially naive part of me thinks that reducing the number of places that plugin-specific logic is used amounts to a better strategy over time but I'll move this logic to functions.php for now.

StevenDufresne: I also noticed that the theme files (like this one) use the namespace `namespace WordPressdotorg\Theme` while the functions.php uses `namespace WordPressdotorg\Pattern_Directory\Theme`. Do you know if that's intentional?

public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
StevenDufresne: Of course it is, because `Boolean( "0" )` is equal to `true`.

_.....Arrested development theme song starts playing...._

:) Thanks for the catch.

public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
coreymckrill: No idea about the namespace thing...

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-button/index.js
```diff
@@ -0,0 +1,55 @@
...
+				<Icon icon={ check } />
+				{ __( "You've reported this pattern", 'wporg-patterns' ) }
+			</p>
+		);
+	}
+
+	if ( ! loggedIn ) {
+		return (
+			<p className="pattern-report-button__copy">
+				<a href="/wp-login.php">{ __( 'Login to report this pattern', 'wporg-patterns' ) }</a>
```
coreymckrill: The URL here should have a `redirect_to` query parameter with the current URL as the value so that after the user logs in they come back to the pattern page (otherwise they end up in WP Admin).

In PHP there's `wp_login_url()` which sets this up for you, but I'm not seeing an equivalent on the Gutenberg side...

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-modal/index.js
```diff
@@ -0,0 +1,155 @@
...
+	const renderView = () => {
+		if ( isLoading ) {
+			return <Spinner />;
+		}
+
+		if ( state.isSubmitted ) {
+			return (
+				<p className="pattern-report-modal__copy">
+					{ __(
+						'Thank you for your report submission. We will review this pattern and act accordingly.',
```
coreymckrill: It feels like this wording needs some tweaking, but I'm not sure what exactly would be better. Maybe just make it simple, something like `Your report has been submitted`?

coreymckrill (commented):
👍 The form is looking good, and works as expected. Left a couple more comments. I'm also thinking that we probably need to put a character limit on the Details field. We could just implement that on the API side and have it send back an error if the threshold is crossed, but it might also be nice to show a character counter beneath the field so users know that there's a limit and they're getting close to it.

@ryelle implemented a character counter for our Speaker Feedback Tool and I remember there were some complications with multibyte characters, but there's a workaround here:

https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js#L81-L97

public_html/wp-content/themes/pattern-directory/css/components/_pattern-report-modal.scss
```diff
@@ -0,0 +1,36 @@
...
+	@include breakpoint( $breakpoint-small ) {
+		width: 400px;
+	}
+
+	form > *:first-child {
+		padding-bottom: $gutter-default;
+	}
+}
+
+.pattern-report-modal--hasFixedHeight {
```
ryelle:
```suggestion
.pattern-report-modal__has-fixed-height {
```
Let's stick with [core CSS coding standards](https://developer.wordpress.org/block-editor/contributors/code/coding-guidelines/#naming) — names should be separated by `-`, not as camelCase, and sections separated by double underscores `__`.

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-button/index.js
StevenDufresne: Ah great! I didn't know there was an underlying redirect system.

We can use `addQueryArgs` to add the redirect.

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-modal/index.js
StevenDufresne: Yeah fair enough. Let's go with `Your report has been submitted` and let the moderators/beta users help us tweak it.

StevenDufresne:
@coreymckrill 
> 👍 The form is looking good, and works as expected. Left a couple more comments. I'm also thinking that we probably need to put a character limit on the Details field. We could just implement that on the API side and have it send back an error if the threshold is crossed, but it might also be nice to show a character counter beneath the field so users know that there's a limit and they're getting close to it.

I think we _could_ implement a backend restriction with a friendly user message which will work with the code as-is.
I'm not sure it's worth doing anything on the front end yet. Ideally, users never have to use this form. Moderation should filter out any obvious abusive patterns and the chance that a user will fill in the form with too much information is pretty low in my opinion. 

Moreover, moderators should be able to make a call with truncated details, so maybe not even send an error.

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-modal/index.js
```diff
@@ -0,0 +1,155 @@
...
+			path: addQueryArgs( '/wp/v2/wporg-pattern-flag' ),
+			method: 'POST',
+			data: {
+				parent: postId,
+				'wporg-pattern-flag-reason': selectedOption,
+				excerpt: details,
+			},
+		} )
+			.then( () => {
+				dispatch( { status: 'submitted' } );
```
ryelle, line 72: After submission, a screen reader user is left in limbo with no success message & focus on nothing, since the submit button disappears. You could add an announcement for success or error, using [wp.a11y.speak](https://github.com/WordPress/gutenberg/blob/cebcfb04281f105174303a453bc91cfa8ce3a136/packages/a11y/README.md#speak), and move focus back to the modal container (on success - error should be fine as-is, since the button stays on screen).

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-modal/index.js
```diff
@@ -0,0 +1,155 @@
...
+					message: err.message,
+				} );
+			} );
+	};
+
+	const handleClose = () => {
+		onClose( state.isSubmitted );
+	};
+
+	const mappedReasons = useMemo( () => {
```
ryelle: Could we do this in the useSelect request?

public_html/wp-content/themes/pattern-directory/src/store/selectors.js
```diff
@@ -53,7 +53,6 @@ export function getPattern( state, id ) {
 }
 
 /**
- * Get the current query.
```
ryelle: Why remove this?

public_html/wp-content/themes/pattern-directory/src/store/test/resolvers.js
```diff
@@ -52,3 +53,26 @@ describe( 'getCategories', () => {
 		} );
 	} );
 } );
+
+describe( 'getPatternFlagReasons', () => {
+	it( 'yields with the requested patterns', async () => {
```
ryelle:
```suggestion
	it( 'yields with the requested pattern flag reasons', async () => {
```

ryelle (changes requested):
Left a few suggestions you can take or leave, and an accessibility issue that should get fixed. The form itself is good, and good job with the store & tests 💯

public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
ryelle: Ahhh, I copied these from the wporg parent theme and didn't notice 😬  Fixed that in d8a68b6471d30b7eb34a1e1452602f659efde6b3

public_html/wp-content/themes/pattern-directory/css/components/_pattern-report-modal.scss
StevenDufresne: Right, my apologies. You already told me about this one :)

public_html/wp-content/themes/pattern-directory/src/store/selectors.js
StevenDufresne: Eek.... Rebase conflict fail.

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-modal/index.js
StevenDufresne: We could but I thought it makes more sense to keep the `reasons` exported from the state in their original form. This mapping is very specific to the radio control. If this were the innermost child component, I would have fewer reservations.

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-modal/index.js
StevenDufresne: `move focus back to the modal container`... Is there an ideal way to move the focus back?

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-modal/index.js
ryelle: Right, I didn't mean to change this in the state, but to do something like this inside the `useSelect` callback:

```js
const reasons = getPatternFlagReasons();

// map reasons for radiocontrol

return {
	isLoading: isLoadingPatternFlagReasons(),
	reasons: mappedReasons,
};
```

I don't see you needing `reasons` in any other part of this component, just `mappedReasons`.

StevenDufresne: Yea, I don't mean in the state exactly but I see `useSelect` as a representation of the state. For me, it would be akin to mapping for a component inside the traditional `function mapStateToProps(state, ownProps?)` function and you would never do that.

StevenDufresne:
@ryelle 
I had a hard time getting focus back to the container so I tried an alternative approach here: 
https://github.com/WordPress/pattern-directory/tree/add/report-pattern-feature-a11y

It displays another window modal on success to get the `focusOnMount` functionality to work.

It does read a bit weird using the "Screen Reader" chrome extension though:
"Entered dialog"
"Close dialog, Your report has been submitted"...

Thoughts?

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-modal/index.js
ryelle: I think it's fairly common to reshape the data from `useSelect` and `mapStateToProps` for whatever the component actually needs. In this case, it would remove the need for `useMemo`. You could call the return value `reasonList` or something to make it clear it's not a direct state value?

public_html/wp-content/themes/pattern-directory/src/components/report-pattern-modal/index.js
ryelle: Usually I would say using `ref`s, but it seems that Modal doesn't pass refs down to the dialog container… I tried playing around with a few different ideas, but ended up back at refs with a wrapper div that would stay on screen through all states. I pushed my solution to this PR (78332c3f2eec2a1a5148b60d6951b4066d2da7ec), but feel free to undo it if you want to look into another solution.

StevenDufresne: No, it's fine. I was hoping there was a function in one of the WordPress packages to grab the ref for the `Modal`. 👍

StevenDufresne:
@ryelle Can you check back in with this PR and advise? Thanks!

ryelle (approved):
Looks good! thanks for sticking with it :)

Issue #16, opened 2020-11-09
Create "My Submitted Patterns" page


MichaelArestad:
There should a page a user can see their submitted patterns and the status of those patterns.

* What are the various statuses of the patterns (submitted, reviewed, rejected)?
* How should a user get to this page? Is there something similar for plugins?

iandunn:
> rejected

The Community Team uses `declined` as a softer word in those situations, and I think it's worked well there.

iandunn:
>  Is there something similar for plugins?

[The Add Your Plugin page](https://wordpress.org/plugins/developers/add/)  will show everyone how many are in the review queue, but I think the accepted/rejected notes are sent via email.

Is that right, @Ipstenu ?

iandunn:
> How should a user get to this page?

I think the only way to see all of your plugins is on profiles.w.org. I think it'd be good to add patterns there, for consistency. 

I don't think that's easy to find, though. I like how [the 5ftF design has a `My Pledges` link in the navbar](https://wordpress.org/five-for-the-future/for-organizations/). That could link to the patterns on the profile, so we don't have to create duplicate pages.

Ipstenu:
Rejections are kinda saved -- everyone who has a rejected plugin, has it listed on the 'author' card:

![Screen Shot 2020-11-11 at  11 Nov - 12 12 11 PM](https://user-images.githubusercontent.com/807903/98859520-2a739e00-2417-11eb-9d74-501f84fd1b50.jpg)

And everyone gets an email (except for 'it's been 6 months...'). As you can see, we don't track 'why' something was rejected. Most of the time we use it for "We've rejected Bob for the same plugin 3 times. Check the emails and see if he's just not getting something or if we have a bigger issue."

In general, rejections are:

* Submitter took 6 months and never finished the review
* Plugin is not acceptable
* Specific issue with submitter

tellyworth:
@MichaelArestad raised the interesting idea that people might like to use this as a repository for their own patterns. For example a designer or integrator might keep a library of patterns for easy access on their client sites.

We won't have any way to authenticate users from an external editor, which means they can't be completely private. But I wonder if this hints at a more nuanced set of statuses than for the plugin/theme directories. Something like:

* I create/remix a pattern and save it.
* It's now accessible to me (as author) on the front end of the directory, but not visible there to others or logged-out users.
* From the Gutenberg editor on my site, I can search for my pattern with some kind of explicit prefix, like "user:tellyworth green button"
[...]

PR #84, opened 2021-04-23
Theme: Show patterns + status in header.

### LINKED ISSUE #54: Show user information on the homepage
If a user is logged in and has submitted patterns, the callout on the homepage should be updated to show relevant info:

Screen Shot 2021-03-30 at 1 15 21 PM

StevenDufresne:
This feature helps gives our users an overview of their patterns.

For more information see #54.

**Missing in this PR**:
- This PR does not currently support the `favorite` count feature yet since the back-end logic currently does not exist. Additionally, I think this feature can ship and be useful without it.
- Currently doesn't have a "View All" until #16 is implemented.

I think this is the most basic shippable version of this control that can be updated once the `view all` and `favorite` functionality is complete. If we agree, we should break the remaining items into new tickets.

Thoughts?

Closes #54

### Screenshots
| Design | This Implementation |
| --- | --- |
|  |  |
| | This uses `.org`'s message styles which differ slightly from the design |

1. Create a few patterns
2. Navigate to `/` 
3. See the callout in the header reflect your pattern(s) status(es)

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/patterns.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/patterns.php
@@ -46,4 +46,17 @@ class Patterns {
 
 		return get_posts( $args );
 	}
+
+	/**
+	 * Returns a list of pending patterns.
+	 *
+	 * @param \WP_Post[] $pattern_posts List of patterns.
+	 *
+	 * @return \WP_Post[] List of patterns in pending status.
+	 */
+	public function filter_pending_patterns( $pattern_posts ) {
+		return array_filter( $patterns_posts, function ( $pattern ) {
+			return $pattern->post_status === 'pending';
+		});
+	}
 }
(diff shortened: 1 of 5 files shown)


public_html/wp-content/plugins/pattern-directory/includes/patterns.php
```diff
@@ -0,0 +1,62 @@
...
+
+namespace WordPressdotorg\Pattern_Directory\Patterns;
+use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\POST_TYPE;
+
+/**
+ * The main class for handling information about patterns.
+ *
+ * @package WordPressdotorg\Pattern-directory
+ */
+class Patterns {
```
StevenDufresne, line 11: Do we want to have a helper-type class to contain some of the pattern-related queries/helper functions like this? 
I stole this approach from the plugin directory where there is a `PluginDirectory` class.

Please advise.

StevenDufresne:
This implements an out-of-date design.

PR #86, opened 2021-04-26
Pattern Details: Add a global asset URL variable, used to build links to theme assets

ryelle:
The theme path is different in different environments, so instead of hard-coding this value, we should provide a variable that uses the dynamic URL from WordPress. We can pass a global variable into the app using `wp_add_inline_script`, so I've set up `wporgAssetUrl`. This is used in the copy-paste guide to load the image from the correct location.

1. View a single pattern page
2. Click Copy, then click Learn More in the success message
3. The image in the modal should appear

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/package.json
+++ b/public_html/wp-content/themes/pattern-directory/package.json
@@ -54,7 +54,7 @@
 	"eslintConfig": {
 		"extends": "../../../../.eslintrc.js",
 		"globals": {
-			"wporgBlockPattern": "readonly"
+			"wporgAssetUrl": "readonly"
 		}
 	},
 	"prettier": "../../../../.prettierrc.js",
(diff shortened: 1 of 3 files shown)


public_html/wp-content/themes/pattern-directory/package.json
```diff
@@ -54,7 +54,7 @@
 	"eslintConfig": {
 		"extends": "../../../../.eslintrc.js",
 		"globals": {
-			"wporgBlockPattern": "readonly"
+			"wporgAssetUrl": "readonly"
```
ryelle, line 57: `wporgBlockPattern` was copied over from the eslintConfig in the pattern creator, so it can just be replaced here. it's not used in the theme.

StevenDufresne (approved):
Yep! Great addition.

PR #87, opened 2021-04-26
Pattern Previews: Always add a theme stylesheet to preview iframes

### LINKED ISSUE #31: Create a "Block Preview" component for displaying published patterns
This will be part of #11. We want a "card"-like view for the block patterns for use on list pages (homepage, your own patterns, etc). It would be nice if this could be a reusable component so we could use it for displaying block-plugins on wp.org/plugins, too. There is already a BlockPreview (or similar) component in gutenberg, might be a good starting point if it's publicly exported.

This is one iteration of the design, for reference:

Screen Shot 2020-12-18 at 5 59 30 PM

ryelle:
This updates the `Iframe` component to always provide a theme stylesheet for all pattern previews. It will default to Twenty Twenty-One, but can also be customized by providing a `themeSlug` to the `<Iframe>`.

This also switches the large preview from using the core "experimental iframe" to our forked iframe component, which I meant to do after #69 merged.

I think we can consider this Fixes #31 — I have a note about trying shadow DOM to see if that improves performance, but a) I'm not seeing any perf issues yet, and b) react shadow DOM libraries seem like more overkill than a few iframes. Happy to revisit this before launch if we notice issues though.

### Screenshots

Screen Shot 2021-04-27 at 12 52 09
Screen Shot 2021-04-27 at 12 52 29

0. Build the files: `yarn workspaces run build`
1. View the patterns grid, any archive view
2. The pattern previews should use Twenty Twenty-One styles
3. Click into any pattern
4. The large preview should also use Twenty Twenty-One styles

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/src/components/iframe/index.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/iframe/index.js
@@ -90,8 +90,17 @@ function setHead( doc, head ) {
 		'<style>body{margin:0}</style>' + head;
 }
 
-function Iframe( { contentRef, children, head, headHTML, ...props }, ref ) {
+function Iframe( { contentRef, children, head, headHTML, themeSlug, ...props }, ref ) {
 	const [ iframeDocument, setIframeDocument ] = useState();
+
+	headHTML +=
+		'<style>body{pointer-events:none;display: flex;align-items: center;justify-content: center;min-height: 100vh;} body > div {width: 100%}</style>';
+
+	if ( themeSlug ) {
+		headHTML += `<link rel="stylesheet" href="https://wp-themes.com/wp-content/themes/${ themeSlug }/style.css" media="all" />`;
+	} else {
+		headHTML += `<link rel="stylesheet" href="https://wp-themes.com/wp-content/themes/twentytwentyone/style.css" media="all" />`;
+	}
 
 	const setRef = useCallback( ( node ) => {
 		if ( ! node ) {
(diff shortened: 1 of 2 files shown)


public_html/wp-content/themes/pattern-directory/src/components/iframe/index.js
```diff
@@ -90,9 +90,18 @@ function setHead( doc, head ) {
...
+function Iframe( { contentRef, children, head, headHTML, themeSlug, ...props }, ref ) {
 	const [ iframeDocument, setIframeDocument ] = useState();
 
+	headHTML +=
+		'<style>body{pointer-events:none;display: flex;align-items: center;justify-content: center;min-height: 100vh;} body > div {width: 100%}</style>';
+
+	if ( themeSlug ) {
+		headHTML += `<link rel="stylesheet" href="https://wp-themes.com/wp-content/themes/${ themeSlug }/style.css" media="all" />`;
+	} else {
+		headHTML += `<link rel="stylesheet" href="https://wp-themes.com/wp-content/themes/twentytwentyone/style.css" media="all" />`;
```
StevenDufresne: Are these backticks necessary?

StevenDufresne (approved):
I agree we can revisit any speed improvements in the future. 

👍

public_html/wp-content/themes/pattern-directory/src/components/iframe/index.js
ryelle: nope, i've switched this back to single quotes.

PR #89, opened 2021-04-27
Theme: Add copy functionality to the block preview.

### LINKED ISSUE #88: Theme: Block Preview card "copy" button should copy the block pattern.
It should leverage code on the single pattern view page introduced in #74.

**Code Reference:**
https://github.com/WordPress/pattern-directory/blob/1c878592553a9dcc8f4f2c8cde05952436fe3f46/public_html/wp-content/themes/pattern-directory/src/components/copy-pattern-button/index.js#L13

**Thoughts**
We don't have designs for the "copy" action confirmation but maybe we could just update the button with a checkmark or copy? Maybe a toast?

Related to: #87

StevenDufresne:
This PR allows users to copy block patterns via the block preview "copy" button. 

As mentioned in the issue, there isn't a design for this so I took a crack at the experience. Since there is very little space to confirm to the user that the action has succeeded, I decided to update the copy button on success from `copy` to `copied` and set a timeout to return to the original text after 3 seconds. 

**Not so great**:
It isn't clear to me what to do when copies fail. I've done some research and the command is fairly reliable. 
It can fail if the copied string is too long but it shouldn't really be triggered by patterns. Additionally, should the copy fail, the lack of response may be enough of a cue? We could turn the button red and the copy to 'Copy failed', maybe. 

**Alternatives:**
a. Same approach with the `setTimeout` but have a more visual change, maybe using color or an icon.
b. Leverage a library to display a `Toast` that typically appears in the bottom right/left corner of the browser window saying something like "Copied pattern."

Fixes: #88 

1. Create a pattern
2. Visit any archive page
3. Hover over a block preview and click "Copy"
4. Notice that the button text has changed to "Copied"
5. Wait 3 seconds, notice the button copy has changed back to "Copy"
6. CTRL + V somewhere and expect to see pattern code.


public_html/wp-content/themes/pattern-directory/src/utils/index.js
```diff
@@ -1 +1,2 @@
 export { getCategoryFromPath, removeQueryString, removeEmptyArgs } from './category';
+export { copyToClipboard } from './copy';
```
StevenDufresne: This is a bit of a wonky file name since it's so specific but the function is kind of a one-off.

ryelle: There's a pattern in gutenberg of calling files by the name of the function they export (thinking of the `e2e-test-utils` package), so I think it's okay if you call this file `copy-to-clipboard.js` 👍🏻

ryelle (approved):
Looks good now 👍🏻

PR #93, opened 2021-05-05
Add API for saving block patterns to a user's favorites

ryelle:
See #52 — This sets out the API side of saving a pattern to a user's favorites. Favoriting will be done via API requests by a logged in user, so we will save/fetch favorites only for that user.

Endpoints introduced:

- `GET /wporg/v1/pattern-favorites` — List all favorites (array of pattern IDs)
- `POST /wporg/v1/pattern-favorites` with arg `id` — add a pattern to favorites, returns true or error
- `DELETE /wporg/v1/pattern-favorites` with arg `id` — remove a pattern from favorites, returns true or error

1. Try interacting with the API in Insomnia/other REST client
2. Make sure the tests pass

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/class-rest-favorite-controller.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/class-rest-favorite-controller.php
@@ -2,7 +2,7 @@
 
 namespace WordPressdotorg\Pattern_Directory\Favorites_API;
 
-use function WordPressdotorg\Pattern_Directory\Favorite\{save_favorite, get_favorites};
+use function WordPressdotorg\Pattern_Directory\Favorite\{add_favorite, get_favorites, remove_favorite};
 use WP_Error, WP_REST_Server, WP_REST_Response;
 
 add_action( 'rest_api_init', __NAMESPACE__ . '\init' );
@@ -87,7 +87,7 @@ function get_items( $request ) {
  */
 function update_item( $request ) {
 	$pattern_id = $request['id'];
-	$success = save_favorite( $pattern_id );
+	$success = add_favorite( $pattern_id );
 
 	if ( $success ) {
 		return new WP_REST_Response( true, 200 );
@@ -108,7 +108,7 @@ function update_item( $request ) {
  */
 function delete_item( $request ) {
 	$pattern_id = $request['id'];
-	$success = save_favorite( $pattern_id, null, false );
+	$success = remove_favorite( $pattern_id );
 
 	if ( $success ) {
 		return new WP_REST_Response( true, 200 );
--- a/public_html/wp-content/plugins/pattern-directory/includes/class-rest-favorite-controller.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/class-rest-favorite-controller.php
@@ -60,7 +60,7 @@ function permissions_check() {
 	if ( ! is_user_logged_in() ) {
 		return new WP_Error(
 			'rest_authorization_required',
-			__( 'You must be logged in to save a pattern.', 'wporg-patterns' ),
+			__( 'You must be logged in to favorite a pattern.', 'wporg-patterns' ),
 			array( 'status' => rest_authorization_required_code() )
 		);
 	}
--- a/public_html/wp-content/plugins/pattern-directory/includes/class-rest-favorite-controller.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/class-rest-favorite-controller.php
@@ -32,8 +32,8 @@ function init() {
 		'wporg/v1',
 		'pattern-favorites',
 		array(
-			'methods' => WP_REST_Server::EDITABLE,
-			'callback' => __NAMESPACE__ . '\update_item',
+			'methods' => WP_REST_Server::CREATABLE,
+			'callback' => __NAMESPACE__ . '\create_item',
 			'args' => $args,
 			'permission_callback' => __NAMESPACE__ . '\permissions_check',
 		)
@@ -85,7 +85,7 @@ function get_items( $request ) {
  * @param WP_REST_Request $request Full data about the request.
  * @return WP_Error|WP_REST_Response
  */
-function update_item( $request ) {
+function create_item( $request ) {
 	$pattern_id = $request['id'];
 	$success = add_favorite( $pattern_id );
 
--- a/public_html/wp-content/plugins/pattern-directory/includes/class-rest-favorite-controller.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/class-rest-favorite-controller.php
@@ -1,0 +1,122 @@
+<?php
+
+namespace WordPressdotorg\Pattern_Directory\Favorites_API;
+
+use function WordPressdotorg\Pattern_Directory\Favorite\{save_favorite, get_favorites};
+use WP_Error, WP_REST_Server, WP_REST_Response;
+
+add_action( 'rest_api_init', __NAMESPACE__ . '\init' );
+
+/**
+ * Set up the endpoints for favoriting patterns.
+ */
+function init() {
+	register_rest_route(
+		'wporg/v1',
+		'pattern-favorites',
+		array(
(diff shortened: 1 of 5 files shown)


public_html/wp-content/plugins/pattern-directory/tests/phpunit/favorite-test.php
```diff
@@ -0,0 +1,160 @@
...
+		$favorites = get_user_meta( self::$user_admin, META_KEY, true );
+		$this->assertSame( $favorites, array( self::$faved_pattern_id, self::$pattern_id ) );
+	}
+
+	/**
+	 * Test unfavoriting a pattern as the admin user.
+	 */
+	public function test_unfavorite_faved_pattern() {
+		wp_set_current_user( self::$user_admin );
+		$success = save_favorite( self::$faved_pattern_id, null, false );
```
StevenDufresne: I had to peek at the `save_favorite` function to understand that `false` un-saves the pattern. Should we create a function that calls `save_favorite` for readability? 

Maybe:
```
function remove_favorite($post, $user = 0) {
   return save_favorite($post, $user, false);
}
```

public_html/wp-content/plugins/pattern-directory/tests/phpunit/favorite-test.php
```diff
@@ -0,0 +1,160 @@
...
+	/**
+	 * Test favoriting a pattern as an anonymous (not logged in) user.
+	 */
+	public function test_favorite_pattern_anon() {
+		$success = save_favorite( self::$pattern_id );
+		$this->assertFalse( (bool) $success );
+	}
+
+	/**
+	 * Test favoriting a pattern as an anonymous user, for a real user.
```
StevenDufresne, line 133: Curious: how does this happen?

StevenDufresne (approved):
Yep, this API is simple and easy to use.

public_html/wp-content/plugins/pattern-directory/tests/phpunit/favorite-test.php
ryelle: It wouldn't happen via the API - we always have a logged in user there, and there are permission checks in place. `save_favorite` defaults to the current user, but a user ID could also be passed in- so on the server, we could call `save_favorite` directly like this (see the note directly below this line).

public_html/wp-content/plugins/pattern-directory/tests/phpunit/favorite-test.php
ryelle: Yeah, I agree it's not super intuitive - I admit I grabbed the base for this from plugin favoriting. I thought since we'll only be interacting with this via the API (which does have separate add/remove functionality), it would be abstracted enough.

public_html/wp-content/plugins/pattern-directory/includes/class-rest-favorite-controller.php
```diff
@@ -0,0 +1,122 @@
...
+ * Check if a given request has access to favorites.
+ * The only requirement for anything "favorite" is to be logged in.
+ *
+ * @return WP_Error|bool
+ */
+function permissions_check() {
+	if ( ! is_user_logged_in() ) {
+		return new WP_Error(
+			'rest_authorization_required',
+			__( 'You must be logged in to save a pattern.', 'wporg-patterns' ),
```
coreymckrill:
```suggestion
			__( 'You must be logged in to favorite a pattern.', 'wporg-patterns' ),
```

public_html/wp-content/plugins/pattern-directory/includes/favorite.php
```diff
@@ -0,0 +1,71 @@
...
+
+	$users_favorites   = get_user_meta( $user->ID, META_KEY, true ) ?: array();
+	$already_favorited = in_array( $post->ID, $users_favorites, true );
+
+	if ( $favorite && $already_favorited ) {
+		return true;
+	} elseif ( $favorite ) {
+		$users_favorites[] = $post->ID;
+	} elseif ( ! $favorite && $already_favorited ) {
+		unset( $users_favorites[ array_search( $post->ID, $users_favorites, true ) ] );
```
coreymckrill: This is a little bit hard to read in one line.

```suggestion
		$index = array_search( $post->ID, $users_favorites, true );
		unset( $users_favorites[ $index ] );
```

public_html/wp-content/plugins/pattern-directory/tests/phpunit/favorite-test.php
coreymckrill: It would be a bit more intuitive if the favorites were saved as separate user meta values instead of as a single serialized array. Then you would use `add_user_meta` and `delete_user_meta` instead of `update`, and you wouldn't necessarily need the boolean. But I'm not sure if there's a compelling reason to switch to that now. It doesn't seem like there would be a big performance difference either way, unless someone was favoriting thousands of patterns.

coreymckrill (approved):
👍 Working as expected using Insomnia. Some minor comments/suggestions below.

* Should we prevent favoriting patterns that aren't published yet? Right now you can favorite any ID that returns a valid pattern post, regardless of post status.

public_html/wp-content/plugins/pattern-directory/tests/phpunit/favorite-test.php
coreymckrill: Actually, thinking about this in light of us wanting to be able to show how many times a specific pattern has been favorited, it will probably be easier and more efficient to switch to storing these in individual user meta rows, instead of as a serialized array...

StevenDufresne:
> 👍 Working as expected using Insomnia. Some minor comments/suggestions below.
> 
> * Should we prevent favoriting patterns that aren't published yet? Right now you can favorite any ID that returns a valid pattern post, regardless of post status.

I think we could prevent them but we need to keep in mind that there will always be the case where a pattern may be removed from the directory and still in the favorite list. We won't be able to rely on the fact that the favorite list is only a list of published pattern posts. 

The front end just overlays the favorite information on already requested lists and when we do query for all pattern posts based on the favorited list we'll make sure that we request patterns posts that are published. So either way, I think we're good, as far as the front end is concerned.

coreymckrill:
I guess as long as we only show published patterns when rendering the grid of a user's personal favorites, there isn't too much to be concerned about around what can be favorited in the first place. Someone could use the API to determine all the valid pattern post IDs, but I don't know what they'd be able to do with that information.

ryelle:
> Should we prevent favoriting patterns that aren't published yet? 

I think we can allow favoriting of anything, but only show patterns the user has access to — as they might favorite their own, private patterns, which aren't technically published.

… although hmm, how do we verify that from the local WP site? If user-a favorites their private patterns, user-b could just enter user-a in their WP site to access the private patterns. Maybe they just don't show up, and you can only access private patterns from the wp.org site?

public_html/wp-content/plugins/pattern-directory/tests/phpunit/favorite-test.php
ryelle: Switched to separate user meta rows, using add_user_meta/delete_user_meta — agreed that this will be easier to query for per-pattern in the future. I also split the save_ function into `add_favorite` and `remove_favorite`, since there really is no reason to keep it confusing 🙂

public_html/wp-content/plugins/pattern-directory/includes/class-rest-favorite-controller.php
```diff
@@ -0,0 +1,122 @@
...
+				return is_numeric( $param );
+			},
+		),
+	);
+	register_rest_route(
+		'wporg/v1',
+		'pattern-favorites',
+		array(
+			'methods' => WP_REST_Server::EDITABLE,
+			'callback' => __NAMESPACE__ . '\update_item',
```
coreymckrill: Minor nitpick: now that we are actually adding a new row when favoriting something, it would be more accurate to have this be `WP_REST_Server::CREATABLE` and the method be `create_item`

PR #94, opened 2021-05-05
Theme: Hook up pattern favorite functionality

### LINKED ISSUE #52: Users should be able to favorite patterns

ryelle:
Fixes #52. This PR is based on #93, which sets up the API, and this PR sets up the data store and hooks up the actions to the `FavoriteButton`.

The `FavoriteButton` now has two props, `patternId` (required) and `showLabel` (defaults to true). This component handles fetching favorites, checking if this pattern (`patternId`) is favorited, and triggering the add/remove from favorite actions. `showLabel` is used to show/hide the "Add to favorites" text when we want to show a smaller button. The label is still available to screen reader users.

### Screenshots

Screen Shot 2021-05-05 at 5 58 43 PM
Screen Shot 2021-05-05 at 5 58 54 PM
Screen Shot 2021-05-05 at 5 59 01 PM

1. Visit the patterns list view, logged in
2. Try favoriting some patterns, they should turn red
3. Click into any favorited pattern, the button in the single pattern view should also be red
4. Unfavoriting patterns should also work, and persist around the site

If you're logged out, the buttons will still be there, but they're disabled (since favorites need a user). We can also just remove them or add a tooltip - wasn't sure which would be better.


StevenDufresne:
> If you're logged out, the buttons will still be there, but they're disabled (since favorites need a user). We can also just remove them or add a tooltip - wasn't sure which would be better.

Visually, it doesn't appear disabled so I think we **need** to do something before this merged. I'm fine if we just remove it for now since the list of favorites doesn't appear very discoverable yet -- at least in the current designs (Ie: How do I get there? ).
 Once it is surfaced better, we should probably push to log in users in most cases.

ryelle:
Updated to remove the ❤️ button until we know we have the ability to favorite patterns.

public_html/wp-content/themes/pattern-directory/src/store/test/resolvers.js
```diff
@@ -31,7 +31,7 @@ describe( 'getPatternsByQuery', () => {
 } );
 
 describe( 'getCategories', () => {
-	it( 'yields with the requested patterns', async () => {
+	it( 'yields with the requested categories', async () => {
```
StevenDufresne, line 35: Thanks for catching this.

StevenDufresne (changes requested):
This is looking good. 

Can we left-align the button on mobile?

StevenDufresne:
Also, we still have a favorite button in the PHP template and you can see the flash when you throttle your connection. We should probably open up a ticket to deal with the out-of-sync content-grid template because the problem exists for all the buttons on the page. It's most obvious for the favorite button because it is absent by default.

https://github.com/WordPress/pattern-directory/blob/efb41c76f60329897303ee69d6105068e8b02a13/public_html/wp-content/themes/pattern-directory/template-parts/content-grid.php#L22

[Here is a screencast](https://d.pr/i/WQTPQB) of the flash.

ryelle:
I've switched this to center the buttons at the mobile breakpoint (480px)
Screen Shot 2021-05-11 at 1 25 52 PM

Then when the buttons wrap, they stay centered:
Screen Shot 2021-05-11 at 1 26 06 PM

No change on larger screens:
Screen Shot 2021-05-11 at 1 26 23 PM

ryelle:
> Also, we still have a favorite button in the PHP template and you can see the flash when you throttle your connection.

👍🏻  opened #104, but I think it's worth waiting on implementing anything until the dust settles a bit. I dropped two possible approaches in a comment.

StevenDufresne (approved):
Good to go! 🥳

PR #95, opened 2021-05-06
Run js format.

StevenDufresne:
Some files have slipped through unformatted because `lint:js` is not 100% aligned with `format:js`. As a result it's currently hard to run `format:js` on your branch because it will include these changes.


ryelle:
I'd rather stick with the format that passes the lint checks, so that the actions here can pass.

Ideally we'd switch to `extends: 'plugin:@wordpress/eslint-plugin/recommended',` in `eslintrc.js`, which reads the prettier config to keep both lint & format in sync. Unfortunately there's a bug with eslint-plugin, where it doesn't read the prettier config if it's a string - I added an issue https://github.com/WordPress/gutenberg/issues/31557 and PR https://github.com/WordPress/gutenberg/pull/31561 to fix that, but for now we're stuck with the conflict.

StevenDufresne:
Closing this for now. By the time 

> I'd rather stick with the format that passes the lint checks, so that the actions here can pass.
> 
> Ideally we'd switch to `extends: 'plugin:@wordpress/eslint-plugin/recommended',` in `eslintrc.js`, which reads the prettier config to keep both lint & format in sync. Unfortunately there's a bug with eslint-plugin, where it doesn't read the prettier config if it's a string - I added an issue [WordPress/gutenberg#31557](https://github.com/WordPress/gutenberg/issues/31557) and PR [WordPress/gutenberg#31561](https://github.com/WordPress/gutenberg/pull/31561) to fix that, but for now we're stuck with the conflict.

I'll close this for now. I think by the time the Gutenberg PRs have merged the files that are changed here will be far out of sync.

Issue #59, opened 2021-03-30
Update creator UI based on mockup


ryelle:
The editor should now be full screen (disable non-fullscreen mode), and have the pattern title visible.

![Create Pattern](https://user-images.githubusercontent.com/541093/113039825-1c27e380-9166-11eb-84b9-7a3890519c5a.png)

The sidebar will have two sections (like the CPT editor), but instead of Document settings it will be Pattern meta information.

![Inspector Sidebar](https://user-images.githubusercontent.com/541093/113040099-64df9c80-9166-11eb-847a-9e2907068862.png)

The "View" link functionality is in #29.

StevenDufresne:
> The editor should now be full screen (disable non-fullscreen mode) [...]

I'm not sure if I understand how to do this nor am I sure I'm viewing the creator properly.

I've created a page called `new-pattern` and I visit `localhost:8888/new-pattern`. I get the creator embedded in the page. 

**Looks like this**:

![](https://d.pr/i/g72ME1.png)

When I run the snippet below, nothing happens; presumably, because i'm not loading this through `wp-admin`?

```
wp.data.dispatch( 'core/edit-post' ).toggleFeature( 'fullscreenMode' );
```

ryelle:
Yep, that's what it looks like right now, and the issue is that it's been totally redesigned to use a fullscreen mode only. The current editor here doesn't have any fullscreen code.

The new version looks a _lot_ closer to the edit-post editor, so I was thinking we might want to switch to that. I initially tried an exploration using `@wordpress/edit-post`, but the design was distinct enough that we couldn't use it (see #26 ). Since the design's changed, I'd like to bring try that again - it should be more stable than the forked editor we've got now.

I merged trunk into that branch & got it to build again: [`try/edit-post-package`](https://github.com/WordPress/pattern-directory/tree/try/edit-post-package)

If you'd like to work on the editor (I'll have a few more issues for other editor features in the next few days), could you investigate this approach? If there are any other gotchas, and/or see if there's a way around the `BrowserUrl` issue I mentioned in #26?

> `BrowserURL` helper updates the page URL to `site.com/pattern/pattern-name/post.php?…` so you can't reload the page. There doesn't seem to be any way to disable this.

StevenDufresne:
I'll hold off and wait for this to settle.

ryelle:
[...]

PR #97, opened 2021-05-10
Creator: Add welcome guide on first load.

### LINKED ISSUE #96: Creator: Pattern Creator Welcome Guide
In order to help users understand the process of creating and uploading patterns, we should provide them with information when they load the pattern creator for the first time.

**Designs**

StevenDufresne:
This PR closes #96, adding a welcome guide for new users. 

I have added the component into the current _Creator_ which will most likely be changed when #59 is resolved. I think it's fine to implement it here and then move it over later.

**To Do:**
This needs integration with the final `pattern-creator`.

### Screenshots

| 1 | 2 | 3 | 4 | Screencast | 
| --- | --- | --- | --- | --- |
|  |  |  |  |  |

1. Create a page with the title `new-pattern`
2. Navigate to `/new-pattern`
3. Expect to see the new welcome guide (same as screenshot)
4. Continue through the welcome guide until complete
5. Reload the page
6. Expect to not see the welcome guide


StevenDufresne:
@shaunandrews Do you think it's realistic we will get a "Learn More" page before launching?... Just considering whether this PR should wait for it or comment out the last view and try to merge in the meantime.

shaunandrews:
You can comment it out for now if you want.

I think the "Learn more" page could just be a page on the directory, like wordpress.org/patterns/guide and the content will likely be simple to start; A list of do's and don'ts and a few tips. Here's a draft I've worked up:

Pattern Guide

> What is a WordPress pattern?
> A WordPress pattern is a collection of blocks arranged together into an thoughtful, intentional, opionated section for a page of post. Once inserted into a page, the author can edit and customize the blocks to their needs.
> Patterns make it incredibly simple to add complex layouts and designs to any WordPress site without having to start from scratch everytime. Patterns can also highlight and expose features and combinations of blocks that may not be immediately obvious or intuitive.
> Use our Pattern Editor to create and submit your own designs. Every WordPress users benefits when you share your patterns and help democratize design.
> Examples of good patterns
> Good patterns combine multiple blocks together to create interesting or useful layouts that would normally be complicated to create from scratch.
> A good pattern highlights the capabilities of the blocks they contain and provide a perfect starting point for customizing their content.
> A good pattern is opinionated and has a well defined purpose.
> Things to avoid
> Patterns that contain a single block are unlikely to be useful.
> Overly simplistic patterns, like a single Paragraph with colors don’t save time or present a compelling purpose.
> Patterns shouldn’t be designed for a single theme.
> Tips and tricks
> Use the Group block to control spacing and color for more than on block at a time. Similary, you can use the Cover block to control your pattern’s height, including min-height and full-viewport designs.

StevenDufresne:
@shaunandrews Right on. That's looking good but I'll comment it out for now since it could get caught up in a few iterations.

StevenDufresne:
After commenting it out and using the guide, It feels like it's missing something about the publishing step? Have we considered some content explaining that patterns need publishing to be shared? It's implied in the "find your patterns" panel but maybe we should consider being more explicit? 

Currently:
- You can build patterns
- We have assets for you
- _Add your pattern to the directory_ <- Should we add this step? Or maybe combine it with the one below?
- Use them on all your sites -- in the editor!
- Need more help?

Maybe we do, maybe we don't. Thoughts?

shaunandrews:
An earlier iteration of the welcome design had a slide about submitting patterns:

image

I removed it because I thought the number of steps was getting too large. But I think we can probably replace the "Find your patterns..." one with the above step. That'd give us a welcome tour with the following steps:

image

ryelle:
Is this PR still in progress, depending on #115, or is it ready for review?

StevenDufresne:
> Is this PR still in progress, depending on #115, or is it ready for review?

Yeah, this is ready for review. The welcome guide will need to reintegrated differently with #115.

public_html/wp-content/plugins/pattern-creator/src/components/welcome-guide/index.js
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Internal dependencies
+ */
+import { ImageCollectionImage, PatternEditorImage, PatternsImage } from './images';
+import './style.css';
+
+/**
+ * Module constants
+ */
+const GUIDE_ID = 'welcomeGuide';
```
ryelle: Does this conflict with the core welcome guide? maybe it should be `wporgPatternGuide`?

ryelle (commented):
The editor crashes when I try to load it, so i think this'll need to be rebased after #115 merges. I gave a quick look through the code though, and it seems fine.

public_html/wp-content/plugins/pattern-creator/src/components/welcome-guide/index.js
StevenDufresne: It doesn't but it's a good idea.

StevenDufresne: Actually, it is the same as core to replace it completely in the editor. This makes sure it loads on page load and when users click `Welcome Guide` in the menu. We could explicitly remove core's version and add ours but this seems like _enough_. Thoughts?

StevenDufresne: After further investigation, now that the editor is initialized differently, we'll have to turn off the welcome guide if possible. There's a small flash of core's version.

StevenDufresne: It turns out you can't really remove the welcome guide (you can stop it from showing for new users on load, but the welcome guide menu link will still exist and call the core guide) but you can replace it by registering a `welcomeGuide` on top of it. 

Now, since we now use `registerPlugin`, which appears to have less priority than we used to have with our custom-built editor, there is a very slight shift from the old one to the new one (see below). The shift is pretty slight but exists nonetheless. The obvious downside of this approach is that if core changes the `id` of the welcome guide, this will no longer work... although this will fail pretty silently.

StevenDufresne:
@ryelle This has been integrated now. 

A couple of differences with the designs:
- The last button isn't the darker blue color (it's hard to target it)
- The close "x" is bigger than the designs (it's the Gutenberg default size)

ryelle:
I'm seeing both welcome guides on the page now, so I think we need a way to disable the core one. Could you toggle it off as soon as the page loads?

We should also look into replacing the Options menu entirely, since about half the menu should not be available to pattern authors.

Screen Shot 2021-06-14 at 5 33 53 PM

StevenDufresne:
> I'm seeing both welcome guides on the page now, so I think we need a way to disable the core one. Could you toggle it off as soon as the page loads?

Yes, added.


> We should also look into replacing the Options menu entirely, since about half the menu should not be available to pattern authors.

I was able to remove the `Tools` section but we won't be able to replace the `Preferences` section. 

What features are left that can't be supported? The preferences appear to be pretty timid and still more or less applicable?

shaunandrews:
> The preferences appear to be pretty timid and still more or less applicable?

Some of them are probably fine to keep, but others need to be removed.

The pre-publish checklist is being replaced with our modal, so this option will need to go.

image

The whole Panels section will probably need to go, too. There's a blank setting, and the Template setting doesn't actually appear. The Custom fields option is broken, but also seems unrelated to the pattern editor.

image

--

As for the tools section of the menu, I think some of these could be helpful. Keyboard Shortcuts, Copy All Content, and Welcome Guide (our variant in the issue) should probably remain available. I could see Help being updated to point to a relevant support page.

image

ryelle:
We'd need to replace the welcome guide to make it work with ours, otherwise it will toggle the core one. Same with Help, we probably can't control that link. We could add a new list with our own menu items though.

StevenDufresne:
> We'd need to replace the welcome guide to make it work with ours, otherwise it will toggle the core one. Same with Help, we probably can't control that link. We could add a new list with our own menu items though.

Correct. All the `Tools` items get removed at the same time. There aren't any ways to turn them off selectively... at least in a respectable manner.

Additionally, there isn't a way to turn off specific toggles in the `Preferences` modal... even with CSS. In my opinion, the publish toggle that would skirt our modal confirmation dialog is not a big deal since we force it off on load and that publish panel should still work anyway. If they click that toggle on, a user has specifically requested that flow and we'll need to make sure that flow works or at least runs validation properly because an astute developer could always turn that flow back on programmatically.

Regarding the empty and template panels, those can be removed (the empty one is weird.. hmmm). Custom fields are not a great feature to keep but are non-breaking.

Now I'm obviously taking the position that we can work with it because there isn't a great way to create our own dropdown. The more menu is only hideable via css and we would have to render a new root component in the menu to have a dropdown and that's not an ideal setup. 

I'll hold off until we have a consensus. 

@ryelle @shaunandrews Thoughts?

ryelle:
I moved the menu & preference discussion over to #192. @StevenDufresne, if you revert the menu changes here, we can probably merge this PR.

StevenDufresne:
> I moved the menu & preference discussion over to #192. @StevenDufresne, if you revert the menu changes here, we can probably merge this PR.

Yep, I removed the code that turns off the tools.

StevenDufresne:
@ryelle Can I get some eyes back on this?

ryelle (approved):

PR #98, opened 2021-05-10
Theme: Update category navigation indicator to be a rectangle

StevenDufresne:
This PR updates the active category indicator from a line to a rectangle to match new designs.

I also removed the existing animation which was too distracting when a square moves laterally as opposed to a line.

I think we can add a more appropriate animation in the future although maybe the loading indicator and the grid loading state is enough.

1. Create block pattern Categories
2. Visit `/`
3. Click different categories and verify the correct category is selected as in the screenshot/gif.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_category-menu.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_category-menu.scss
@@ -17,7 +17,7 @@
 		text-decoration: none;
 
 		&:hover,
-		&:active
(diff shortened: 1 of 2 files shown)


ryelle (approved):
Looks good! Personally I don't think this needs any animation, the background feels like it grabs my attention enough.

Adding a screenshot, since I'm not able to see the one in the description here.

Screen Shot 2021-05-10 at 5 18 54 PM

Issue #103, opened 2021-05-10
Pattern Grid: Update patterns with new design


ryelle:
Items should have a fixed height (195px in the design). Title and some meta information (see #100) appear under each pattern. On hover, the favorite & copy actions appear.

![](https://user-images.githubusercontent.com/541093/117468963-3a31f200-af23-11eb-83e4-3a3872a0cdb2.png)

ryelle:
Patterns in the author's "my patterns" list can also have a status label, ex:

![pattern-status](https://user-images.githubusercontent.com/541093/117728208-256d8c80-b1b7-11eb-91ed-7f6edabf998c.png)

These patterns also have an extra actions dropdown:

![Hovered](https://user-images.githubusercontent.com/541093/117728808-ef7cd800-b1b7-11eb-88ec-d7d934cbfd4a.png)

![Owner Options Menu](https://user-images.githubusercontent.com/541093/117728811-f0ae0500-b1b7-11eb-8e5d-9a9bfa346c1c.png)

shaunandrews:
> Items should have a fixed height (195px in the design)

My intention is that the rendered pattern should be contained within a box with a 1.5 aspect ratio — the height is dependent on the width of the column. The design uses a 960px max-width, and because of the way Figma renders box-shadows, the columns are actually meant to be 300px wide, which makes the rendered pattern 300x200px.

Here's a Figma that goes over how the grid should respond as the viewport changes: https://www.figma.com/file/bz4qs4jZCkqvQzPH2fmX6C/Pattern-Directory?node-id=1733%3A1

ryelle:
> My intention is that the rendered pattern should be contained within a box with a 1.5 aspect ratio

Thanks for clarifying 👍🏻 

When do you think it should drop to that 2-column layout? We have existing breakpoints at 960 or 782 wide.

shaunandrews:
782 seems fine if we don't want to add another breakpoint.

Issue #101, opened 2021-05-10
Pattern Grid: Add a loading state


ryelle:
When loading items in a grid, switching between categories, or searching, the grid should switch to a loading state:

![loading-patterns](https://user-images.githubusercontent.com/541093/117724019-2ac7d880-b1b1-11eb-8526-bdf1c3dca429.gif)

PR #106, opened 2021-05-12
Theme: Update `CategoryContextBar ` with new loading UI.

StevenDufresne:
This PR updates the loading state for the `CategoryContextBar` in reference to the new designs mentioned in #101.

These changes here are not enough to close the ticket; the pattern grid itself needs animation updates. This PR doesn't address those changes as the loading state for the pattern grid relies on yet to be made changes for #103.

**Notes**:
In some cases, there is some jank when the loader is animating out because it's also rendering the grid at the same time. It's most visible on quick internet connections. This should be a non-issue once the pattern grid loading is improved and rendering doesn't require such a significant layout recalculation.

See #101, #103

### Screenshots
Screencast of animations

1. Create some block pattern categories
2. Create some patterns
3. Associate categories to patterns
4. Load `/`, click through categories
5. Expect to see UI matching designs.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_category-context-bar.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_category-context-bar.scss
@@ -1,4 +1,4 @@
-.category-context__bar {
+.category-context-bar {
 	margin: 0 $gutter-default;
 	background: $color-gray-light-300;
 	border-radius: 2px;
@@ -44,7 +44,7 @@
 	}
 }
 
-.category-context__bar__copy {
+.category-context-bar__copy {
 	display: flex;
 	align-items: center;
 
@@ -54,7 +54,7 @@
 	font-weight: 400;
 }
 
-.category-context__bar__title {
+.category-context-bar__title {
 	margin: 0;
 	padding-right: $gutter-default;
 	color: $color__text;
@@ -63,7 +63,7 @@
 	text-transform: uppercase;
 }
 
-.category-context__bar__links {
+.category-context-bar__links {
 	display: flex;
 	align-items: center;
 
@@ -72,7 +72,7 @@
 	}
 }
 
-.category-context__bar__spinner {
+.category-context-bar__spinner {
 	display: flex;
 	width: 28px;
 	margin: 0;
@@ -85,14 +85,14 @@
 	}
 }
 
-.category-context__bar__spinner--is-hidden {
+.category-context-bar__spinner--is-hidden {
 	overflow: hidden;
 	width: 0;
 	opacity: 0;
 	transform: scale(0);
 }
 
-.category-context__bar--is-hidden {
+.category-context-bar--is-hidden {
 	height: 0;
 	overflow: hidden;
 }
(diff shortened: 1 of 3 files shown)


public_html/wp-content/themes/pattern-directory/css/components/_category-context-bar.scss
```diff
@@ -43,34 +42,57 @@
...
 	}
+}
 
-	&__copy {
-		margin: 0;
-		padding: $gutter-default * 0.75 $gutter-default;
-		font-size: 0.8125rem;
-		font-weight: 400;
-	}
+.category-context__bar__copy {
```
ryelle: I missed this in a previous review, but class names should only have one `__` separating the root export (`category-context-bar` in CSS) & descriptor (`copy`). Any other separators should be hyphens `-`. So this should be `.category-context-bar__copy`. We're roughly following [the gutenberg standards](https://developer.wordpress.org/block-editor/contributors/code/coding-guidelines/).

public_html/wp-content/themes/pattern-directory/css/components/_category-context-bar.scss
```diff
@@ -43,34 +42,57 @@
...
 }
+
+.category-context__bar__spinner--is-hidden {
+	overflow: hidden;
+	width: 0;
+	opacity: 0;
+	transform: scale(0);
+}
+
+.category-context__bar--is-hidden {
```
ryelle: Where is this class used?

ryelle (changes requested):
The animation is working as designed 👍🏻 

It feels a little strange when clicking between categories, especially on a faster internet connection, since the start of text moves left & right with the spinner. That might be fine, since people _probably_ won't be doing this in real life, but I think we'll want to keep an eye on this once it's merged & the pattern grid loading is in place.

https://user-images.githubusercontent.com/541093/118532929-7fb9a080-b715-11eb-8e9c-e701c5939664.mov

Can you disable the animation when `prefers-reduced-motion: reduce` is set?

The class naming issue can be resolved in this PR, or we can have a followup issue to check all classnames for the naming conventions.

public_html/wp-content/themes/pattern-directory/css/components/_category-context-bar.scss
StevenDufresne: Yep, makes sense. Thanks.

StevenDufresne: There was originally a root component called `category-context` which I refactored out.

public_html/wp-content/themes/pattern-directory/css/components/_category-context-bar.scss
StevenDufresne: Removed.

StevenDufresne:
> It feels a little strange when clicking between categories, especially on a faster internet connection, since the start of text moves left & right with the spinner. That might be fine, since people _probably_ won't be doing this in real life, but I think we'll want to keep an eye on this once it's merged & the pattern grid loading is in place.

Agreed.

> Can you disable the animation when `prefers-reduced-motion: reduce` is set?

Yep, done in 55c9ec3.

ryelle (approved):
Thanks for updating the classnames. Motion is disabled when "reduced motion" is selected 👍🏻 

Good to merge!

Issue #102, opened 2021-05-10
Pattern grid: Add pagination


ryelle:
Broken out from #62, with the new pagination styling. `getQueryString` will need to be updated so that it does _not_ remove the pagination parameters. Does this change anything else about how the patterns should be saved in the store?

<img width="686" alt="image-1" src="https://user-images.githubusercontent.com/541093/117724391-ad509800-b1b1-11eb-89a1-513c25c7c4c8.png">

PR #110, opened 2021-05-18
Update the patterns state to track results per-page

ryelle:
See #102 — the current patterns store was designed to ignore pagination & use infinite scroll. Now that we're bringing back pagination, we need to track the number of pages for each query response.

Before, `queries` just tracked a list of IDs per stringified query:
```js
'': [ 31, 25, 27, 26 ],
```

Now, we track an object with information about the query, including total pages, total pattern count, and then a list of IDs per page.
```js
'': {
	total: 4,
	totalPages: 2,
	1: [ 31, 25 ],
	2: [ 27, 26 ],
},
```

This also introduces new selectors for `getPatternTotalsByQuery` & `getPatternTotalPagesByQuery` (the former can be used in the context bar for result count). This shouldn't change any functionality in the theme yet, since the `getPatternsByQuery` selector still returns the patterns for the given query, and no UI for requesting other pages has been added yet.

1. Check that the current categories still work
2. Check that the tests pass: `yarn workspace wporg-pattern-directory-theme test:unit`

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/src/store/resolvers.js
+++ b/public_html/wp-content/themes/pattern-directory/src/store/resolvers.js
@@ -22,8 +22,8 @@ import { getQueryString } from './utils';
 async function parseResponse( response ) {
 	try {
 		return {
-			total: response.headers?.get( 'X-WP-Total' ),
-			totalPages: response.headers?.get( 'X-WP-TotalPages' ),
+			total: Number( response.headers?.get( 'X-WP-Total' ) || 0 ),
+			totalPages: Number( response.headers?.get( 'X-WP-TotalPages' ) || 0 ),
 			results: await response.json(),
 		};
 	} catch ( error ) {
@@ -43,8 +43,8 @@ export function* getPatternsByQuery( query ) {
 		yield loadPatterns( queryString, {
 			page: query.page || 1,
 			patterns: results,
-			total: Number( total ),
-			totalPages: Number( totalPages ),
+			total: total,
+			totalPages: totalPages,
 		} );
 	} catch ( error ) {}
 }
--- a/public_html/wp-content/themes/pattern-directory/src/store/resolvers.js
+++ b/public_html/wp-content/themes/pattern-directory/src/store/resolvers.js
@@ -2,7 +2,8 @@
  * WordPress dependencies
  */
 import { addQueryArgs } from '@wordpress/url';
-import { apiFetch } from '@wordpress/data-controls';
+// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
+import { __unstableAwaitPromise, apiFetch } from '@wordpress/data-controls';
 
 /**
  * Internal dependencies
@@ -16,16 +17,35 @@ import {
 	loadPatternFlagReasons,
 	loadPatterns,
 } from './actions';
-import { getQueryString } from './utils';
+import { PER_PAGE, getQueryString } from './utils';
+
+async function parseResponse( response ) {
+	try {
+		return {
+			total: response.headers?.get( 'X-WP-Total' ),
+			totalPages: response.headers?.get( 'X-WP-TotalPages' ),
+			results: await response.json(),
+		};
+	} catch ( error ) {
+		return {};
+	}
+}
 
 export function* getPatternsByQuery( query ) {
 	const queryString = getQueryString( query );
 	try {
 		yield fetchPatterns( queryString );
-		const results = yield apiFetch( {
-			path: addQueryArgs( '/wp/v2/wporg-pattern', query ),
+		const response = yield apiFetch( {
+			path: addQueryArgs( '/wp/v2/wporg-pattern', { ...query, per_page: PER_PAGE } ),
+			parse: false,
 		} );
-		yield loadPatterns( queryString, results );
+		const { total, totalPages, results } = yield __unstableAwaitPromise( parseResponse( response ) );
+		yield loadPatterns( queryString, {
+			page: query.page || 1,
+			patterns: results,
+			total: total,
+			totalPages: totalPages,
+		} );
 	} catch ( error ) {}
 }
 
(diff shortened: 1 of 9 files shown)


public_html/wp-content/themes/pattern-directory/src/store/resolvers.js
```diff
@@ -16,16 +17,35 @@ import {
...
-			path: addQueryArgs( '/wp/v2/wporg-pattern', query ),
+		const response = yield apiFetch( {
+			path: addQueryArgs( '/wp/v2/wporg-pattern', { ...query, per_page: PER_PAGE } ),
+			parse: false,
+		} );
+		const { total, totalPages, results } = yield __unstableAwaitPromise( parseResponse( response ) );
+		yield loadPatterns( queryString, {
+			page: query.page || 1,
+			patterns: results,
+			total: Number( total ),
```
StevenDufresne: This is going to `NaN` if something goes wrong with parsing (`Number(undefined)`), I expect we would want to return `undefined` or `0` in that case. Maybe consider moving `Number()` into the `parseResponse` function.

public_html/wp-content/themes/pattern-directory/src/store/utils.js
```diff
@@ -4,6 +4,9 @@
 import { addQueryArgs } from '@wordpress/url';
 import { __ } from '@wordpress/i18n';
 
+// Number of patterns per page, should be multiple of 2 and 3 (for 2- and 3-column layouts).
+export const PER_PAGE = 18;
```
StevenDufresne: Is this a default that can be added to the custom post type instead of here?

public_html/wp-content/themes/pattern-directory/src/store/test/resolvers.js
```diff
@@ -5,9 +5,10 @@ import apiPatterns from './fixtures/patterns';
 import apiCategories from './fixtures/categories';
 import apiPatternFlagReasons from './fixtures/pattern-flag-reasons';
 import { getCategories, getFavorites, getPatternFlagReasons, getPatternsByQuery } from '../resolvers';
+import { PER_PAGE } from '../utils';
 
 describe( 'getPatternsByQuery', () => {
-	it( 'yields with the requested patterns', async () => {
+	it( 'yields with the requested patterns & query meta', async () => {
```
StevenDufresne, line 10: Since the query is most important should we write a test to expose the error state for future error handling?

public_html/wp-content/themes/pattern-directory/src/store/resolvers.js
```diff
@@ -16,16 +17,35 @@ import {
...
 	const queryString = getQueryString( query );
 	try {
 		yield fetchPatterns( queryString );
-		const results = yield apiFetch( {
-			path: addQueryArgs( '/wp/v2/wporg-pattern', query ),
+		const response = yield apiFetch( {
+			path: addQueryArgs( '/wp/v2/wporg-pattern', { ...query, per_page: PER_PAGE } ),
+			parse: false,
+		} );
+		const { total, totalPages, results } = yield __unstableAwaitPromise( parseResponse( response ) );
```
StevenDufresne, line 42: I don't see this PR making use of the `AWAIT_PROMISE` dispatched action. Is it used internally somewhere or is there a future plan to rely on it for loading the grid?

StevenDufresne (commented):
This is a good, easy-to-understand PR. I left a few comments in the review.

public_html/wp-content/themes/pattern-directory/src/store/resolvers.js
ryelle: It's just an internal, there's no way to get the results of a promise otherwise.

public_html/wp-content/themes/pattern-directory/src/store/utils.js
ryelle: Hm, I think there was a reason I wanted to define it here, but I'm not sure what that was 🙃 

Changing the default in for the CPT is a better idea

public_html/wp-content/themes/pattern-directory/src/store/test/resolvers.js
ryelle: I'm not sure what you mean here, can you explain?

public_html/wp-content/themes/pattern-directory/src/store/resolvers.js
ryelle: Done in https://github.com/WordPress/pattern-directory/pull/110/commits/4dbef7c0c97e8e9e8506ad2be4b2036f1ef690ce - Moved the `Number` conversion to `parseRequest`, with a fallback to 0.

public_html/wp-content/themes/pattern-directory/src/store/utils.js
ryelle: Done in https://github.com/WordPress/pattern-directory/pull/110/commits/fe84f2445ca34ce6a5043996c48fdbcc08cc344d - any request that doesn't specify a per_page will use the default of 18.

public_html/wp-content/themes/pattern-directory/src/store/test/resolvers.js
StevenDufresne: Since we are catching/disregarding errors when we query the API, smart error handling will be tricky since it's not clear what the state will look like when a request fails. I'm wondering if we should write a test, intercept the API, fail it however we want, and expose what the state will look like in that scenario. It would give future developers some information on how to improve query retries or help provide better user messaging if we don't get to that before MVP launch.

For example, it appears to me, by looking at the code that if the query fails we should get:

```
{
 page: 1,
 patterns: undefined,
 total: undefined,
 totalPages: undefined,
}
```
A test would communicate that clearly.

StevenDufresne: I'm not stuck on this though, feel free to merge without it.

StevenDufresne (approved):

public_html/wp-content/themes/pattern-directory/src/store/test/resolvers.js
ryelle: Oh I always planned on us getting to this before launch, just not necessarily here 😁  In fact, i ended up doing some basics in #113.

Issue #33, opened 2021-01-18
Search and/or group patterns by block


tellyworth:
It's likely that the editor UI will have a flow where users could discover patterns that contain a particular block.

As a hypothetical example, if I'm adding a Buy Now button to content, it should be easy for me to discover that my e-commerce plugin already has some patterns for product layouts that contain the Buy Now button, to save me doing it all manually.

That suggests some things would be required on the back end:

* A way to search both local and remote patterns for an exact match by block slug.
* Possibly a way to retrieve all local patterns grouped by block (if for example the inserter UI was to display blocks and patterns side by side).

This is still a feature idea in progress so it's not clear exactly how it might pan out and what the final flow and requirements might be.

mtias:
Related https://github.com/WordPress/gutenberg/issues/27575.

ryelle:
Block patterns can have a `blockTypes` property, which identifies blocks that can be converted into this pattern. For example, the paragraph block:

<img width="684" alt="Screen Shot 2021-04-21 at 2 31 17 PM" src="https://user-images.githubusercontent.com/541093/115603585-593b4d80-a2ae-11eb-9102-35c61289a4aa.png">

We should add some (optional) way of flagging patterns as related to block types, and on the validation side, make sure the pattern can be transformed from the chosen block. It seems like patterns _can_ be related to more than one block type, but I think in practice, it's usually just one.

The API & proxy API will also need to be updated to allow a "blockTypes" parameter for fetching patterns.

ryelle:
There is a PR on the Gutenberg side to use the `blockTypes` meta value when registering patterns: https://github.com/WordPress/gutenberg/pull/32113

We still need to create a flow for setting the blockType value when creating a pattern.

ryelle:
We can set `blockTypes` for a pattern, and that will pass through the API to the editor to make it work (once WordPress/gutenberg#32113 is merged).

The unanswered question here is how that should work (in the UI) when creating a pattern. Since we still don't have a design here, and the gutenberg API has no traction (so it won't make 5.9), I've pulled this out of the pattern creation v1 milestone. We can revisit it post-beta or post-launch.

ryelle:
[...]

PR #111, opened 2021-05-21
Pattern CPT & API: Add block types meta field

ryelle:
See #33 — this adds a meta field for block types, which is exposed in the API as `meta.wpop_block_types`. This is an array of block type names. Once this is added to the CPT and deployed, it will become available to `http://api.wordpress.org/patterns/1.0/`. A corresponding PR will be made in gutenberg to add support to the `/__experimental/pattern-directory/patterns` endpoint.

I went with a meta field for now, since it can easily be set (in UI or WP-CLI) for the existing core patterns, and doesn't need the full infrastructure of a taxonomy. If we decide once building the UI that a taxonomy makes more sense, we can write a script to convert the post meta (but I think meta will be the way to go).

1. Add a custom field called `wpop_block_types` to a pattern. The value must be a block type name, formatted like `namespace/name` - it should strip any non-alphanumeric characters out (allowing / and -).
2. GET /wp-json/wp/v2/wporg-pattern/[ID] — the meta value should include the value you entered.
3. GET /wp-json/wp/v2/wporg-pattern/ — all patterns should have a `meta.wpop_block_types` value, defaulting to empty array.

Try adding various values to the meta field, add multiple, etc.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -143,6 +143,25 @@ function register_post_type_data() {
 			),
 		)
 	);
+
+	register_post_meta(
+		POST_TYPE,
+		'wpop_block_types',
+		array(
+			'type'              => 'string',
+			'description'       => 'A list of block types this pattern supports for transforms.',
+			'single'            => false,
+			'sanitize_callback' => function( $value, $key, $type ) {
+				return preg_replace( '/[^a-z0-9-\/]/', '', $value );
+			},
+			'auth_callback'     => __NAMESPACE__ . '\can_edit_this_pattern',
+			'show_in_rest'      => array(
+				'schema' => array(
+					'type' => 'string',
+				),
+			),
+		)
+	);
 }
 
 /**


public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
```diff
@@ -143,6 +143,25 @@ function register_post_type_data() {
 			),
 		)
 	);
+
+	register_post_meta(
```
iandunn, line 147: > I went with a meta field for now, since it can easily be set (in UI or WP-CLI) for the existing core patterns and doesn't need the full infrastructure of a taxonomy. 

🤔  At first glance, a taxonomy seems easier to me UI-wise, and has significant perf improvements for searching. I feel like I'm probably missing something, though?

Those benefits will probably be mitigated when we (presumably) switch to Elasticsearch, though. And you're right that we can always migrate if needed.

I'm open to either, just curious.

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
```diff
@@ -143,6 +143,25 @@ function register_post_type_data() {
...
+		POST_TYPE,
+		'wpop_block_types',
+		array(
+			'type'              => 'string',
+			'description'       => 'A list of block types this pattern supports for transforms.',
+			'single'            => false,
+			'sanitize_callback' => function( $value, $key, $type ) {
+				return preg_replace( '/[^a-z0-9-\/]/', '', $value );
```
iandunn, line 155: _(not a blocker)_

Is there not already a function _(from Core or elsewhere)_ to sanitize block names?

If not, it might be good to pull this out into one, but that can always be done later too.

iandunn (approved):
Left a few questions, but fine to merge as-is.

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
ryelle: Not that I found - core just [runs some regexes](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/class-wp-block-type-registry.php#L60) and flags it as `_doing_it_wrong` if they don't match.

ryelle: I'd rather not worry about it now - if we need to validate/sanitize block names again, I'll make it then.

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
ryelle: We'll probably have to build our own UI for it regardless, or use components that will make no real difference (between sources). I also went with meta since we could use the registered block types as the list to choose from, while a taxonomy would have to be populated and kept up to date by a human.

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
iandunn: 👍🏻

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
iandunn: 👍🏻

PR #113, opened 2021-05-26
Pattern Grid: Add support for paged requests

ryelle:
See #102, based off #110 — This sets up some more groundwork for bringing in pagination, everything but the pagination component itself.

- Prevent the 404 redirect, so the app always controls empty states.
- "Handle" error states when loading patterns (pass the error through, not doing anything with it yet)
- Pass the currentCategory down to the category menu, rather than the entire path
- Pull the query out of PatternGrid, this will help us use PatternGrid on the author page too
- Refactor `getCategoryFromPath` to get the value after `pattern-categories`, rather than a fixed offset
- Abstract `getCategoryFromPath` to `getValueFromPath`, and use this to get the page number too
- Tests for the above

### Screenshots

No visual changes here

1. Make sure the tests pass
2. Create more than 18 patterns (or adjust [the value here](https://github.com/WordPress/pattern-directory/blob/4dbef7c0c97e8e9e8506ad2be4b2036f1ef690ce/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php#L341)) so that you have paged results
3. View the home page, it should show you the first 18 results
4. Go to `/page/2`, you should see different patterns, the second page
5. Try going to categories, and manually adding pages, ex `/pattern-categories/columns/page/2`, it should work (assuming you have patterns)
6. If you go to a page that doesn't exist, right now it just removes the spinner - eventually we'll want a proper error message or empty state


public_html/wp-content/themes/pattern-directory/src/store/reducer.js
```diff
@@ -36,10 +36,16 @@ function queries( state = {}, action ) {
 	const patternIds = ( action.patterns || [] ).map( ( { id } ) => id );
 	const { page, total, totalPages } = action;
 	switch ( action.type ) {
-		case 'LOAD_BLOCK_PATTERNS':
+		case 'LOAD_BLOCK_PATTERNS': {
```
StevenDufresne, line 39: Is the idea here that we want `{}` for multiple lines?

StevenDufresne (commented):
This is a nice improvement. Thanks!

StevenDufresne (approved):

public_html/wp-content/themes/pattern-directory/src/store/reducer.js
ryelle: It creates local scope so that I can use `const _queryState` in both case branches. I'm not suggesting it as practice going forward, just a workaround to keep the variables understandable.

PR #114, opened 2021-05-26
Pattern Grid: Add the pagination component

### LINKED ISSUE #102: Pattern grid: Add pagination

ryelle:
Fixes #102, based on #113. This PR finally adds the pagination component 🎉 

This adds a new component, Pagination, which takes a currentPage and totalPages & outputs a possibly truncated list of pages to navigate to. It expects to be within the Route provider, since it uses `path` & `updatePath` to navigate around pages.

I created a helper function for the truncated list, with tests in `utils/test/get-pagination-list.js`.

### Screenshots

Basic pagination with 2 pages
Screen Shot 2021-05-26 at 5 57 19 PM

Pagination gets truncated after 2 if there are more than 5 total pages
page-first

Previous & next only show up if there are next/prev to navigate to
page-second

If you navigate into the middle of a truncation, it truncates to either side
page-lots

On smaller screens, the spacing shrinks a bit and the Next/Previous labels switch to arrows
page-small

1. Create more than 18 patterns (or adjust [the value here](https://github.com/WordPress/pattern-directory/blob/4dbef7c0c97e8e9e8506ad2be4b2036f1ef690ce/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php#L341)) so that you have paged results
2. View any pattern list page, it should show the pagination at the bottom
3. Click through the pages, it should work
4. If you have a category with few patterns (just one page), it should not show pagination

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_components.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_components.scss
@@ -29,6 +29,7 @@
 @import "page";
 @import "pattern-grid-menu";
 @import "pattern-grid";
+@import "pattern-pagination";
 @import "pattern-preview";
 @import "pattern-report-button";
 @import "pattern-report-modal";
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern-pagination.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern-pagination.scss
@@ -1,0 +1,39 @@
+.pagination {
+	max-width: 960px;
+	margin: 1.5rem;
+
+	@include breakpoint( $breakpoint-large ) {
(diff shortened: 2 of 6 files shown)


StevenDufresne (commented):
This looks good.

If I visit a page outside of the total # of patterns, say `page/60/`, nothing happens. We try to load the page and it's empty. Should we handle that in this PR?

ryelle:
That's expected (I mentioned it in #113). This PR is just for the pagination, so I don't want to add another thing for empty states. Added #116 so we can get a design for this.

PR #112, opened 2021-05-24
Creator: Publish submission window

StevenDufresne:
This adds a window to retrieve pattern information and save it to the pattern directory. This will remain incomplete until the following is added:
- Connect it to the editor
- Add link URL to success confirmation (currently unknown)

Fixes #60

### Example Usage
```
<PublishModal  
    isSubmitted={ true/false } // If true, shows the save success confirmation window
    onSubmit={ () => savePost() }
    onClose={ () => setShowModal( false ) }
/>
```

### Screenshots

| View 1 | View 2 | Confirmation |  GIF |
| --- | --- |  --- | --- |
|  |  |   |  |

Issue #23, opened 2020-11-16
Show demo content in dynamic blocks when creating and previewing a pattern


tellyworth:
Should it be possible for patterns to contain placeholder blocks, such as "your image goes here"?

If so, what's needed to support that?

What kinds of placeholders are needed, besides images?

ryelle:
I think all the media blocks have default views when an asset hasn't been picked yet:

![Screen Shot 2020-11-16 at 5 16 47 PM](https://user-images.githubusercontent.com/541093/99314720-83846d00-282f-11eb-8a47-90f3aeacab38.png)
![Screen Shot 2020-11-16 at 5 17 53 PM](https://user-images.githubusercontent.com/541093/99314805-aadb3a00-282f-11eb-8c9f-e095124c1f7c.png)

_brainstorming some ideas…_

If a pattern author wanted to upload their own assets, they could upload an arbitrary image that says "placeholder"? We could provide placeholder assets, that they could upload?

We could deregister the media blocks in the creator, create our own "Placeholder Cover" (etc), and swap them out somewhere in the API process?

tellyworth:
I like the idea of extending media blocks with our own that display a special placeholder. That way we could make them look like anything during a preview, and change it retrospectively if needed.

On the editing side, those default picker views seem just fine as a starting point. As long as we can make it save and validate sensibly (which it seems the editor is happy to do already).

We probably need to distinguish between two use cases: one where the pattern author wants to upload a fixed image/asset for reuse in the block; and the other where it's a placeholder. That's probably technically trivial but the UX might need some thought.

ryelle:
Blocks that could use special "placeholder" handling due to file uploads:

- core/audio
- core/cover
- core/file
- core/gallery
- core/image
- core/media-text
- core/video

ryelle:
Other placeholders might be needed for dynamic content blocks, like:

- core/archives
- core/calendar
- core/categories
- core/latest-comments
- core/latest-posts
- core/tag-cloud

These display the content on the site, but we don't want to demonstrate with the content on w.org/patterns. We should pull together some demo content that can be displayed instead.

shaunandrews:
I don't think we need to do anything special to the media block placeholders. It's up to the designer if they want their pattern to provide an image that can be replaced, or an empty Image block; Both should work.

We will have to provide a collection of site content for dynamic blocks. Does something like this exist currently for themes that we can reuse, or will we need to create a bunch of example content ourselves?

[...]

PR #115, opened 2021-05-27
Try `@wordpress/edit-post` for the creator UI.

### LINKED ISSUE #59: Update creator UI based on mockup

StevenDufresne:
This PR is an experiment to test the use of `import { initializeEditor } from '@wordpress/edit-post';`

@ryelle Since I didn't author all the changes in the branch, can you take a really critical eye? 

Closes #59.

**Implements**:
- Various style updates

**Doesn't yet Implement**:
- Tie into Publish window #112

- Visit `/new-pattern/`
- Add a title
- Add some blocks
- Add a description
- Select some categories
- Change the viewport width
- Select block scope

In the browser console, paste the following one by one:

```
wp.data.select('core/editor').getEditedPostAttribute('title')
wp.data.select('core/editor').getEditedPostAttribute('pattern-categories')
wp.data.select('core/editor').getEditedPostAttribute('meta')
```
Expect to see your updates values.


public_html/wp-content/plugins/pattern-creator/src/plugins/viewport-header-control.js
```diff
@@ -0,0 +1,159 @@
...
+				) }
+				label={ __( 'Select a viewport width', 'wporg-patterns' ) }
+			>
+				{ ( { onClose } ) => (
+					<>
+						<MenuGroup label="Viewport Width">
+							<p className="viewport-header-control__copy">
+								This is used when displaying a preview of your pattern.
+							</p>
+							<TextControl
```
StevenDufresne, line 118: I wanted to use the `UnitControl` component here but since we are injecting the button into the dom via the `insertButton` function, we can't manage the state properly. Every time the state updates, in this case, `viewportWidth`, the button is removed and rebound to the DOM. So technically we can only watch for changes in one variable. The way the `<UnitControl>` works, it requires you to manage both the number and the unit separately.

The alternative to injecting the button is to `pin` an icon into the header and rearrange the order using CSS to match the order in the designs.  I looked into that and it appears you can't just pin an icon. You have to provide a `PluginSidebar` which we don't want. We only want a dropdown menu.

public_html/wp-content/plugins/pattern-creator/src/api-middleware/filter-endpoints.js
```diff
@@ -0,0 +1,13 @@
+// Use a middleware provider to intercept and modify API calls.
+// Short-circuit POST requests, bound queries, allow media, etc.
+export default function( options, next ) {
+	// Add limits to all GET queries which attempt unbound queries
+	options.path = options.path.replace( 'per_page=-1', 'per_page=10' );
```
ryelle, line 5: I don't remember if I added this initially or not, it's been a while, but do you know what problem this is solving?

public_html/wp-content/plugins/pattern-creator/src/components/unit-control/index.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Modules constants
+ */
+const UNITS = [
+	{ value: 'px', label: 'px' },
+	{ value: '%', label: '%' },
+	{ value: 'em', label: 'em' },
+	{ value: 'rem', label: 'rem' },
+	{ value: 'vw', label: 'vw' },
+];
```
ryelle: You can address this in a follow-up PR/issue, but the only value supported by patterns is px, so we'll need to figure out conversions or tweak this control.

public_html/wp-content/plugins/pattern-creator/src/plugins/block-scope-settings-panel.js
```diff
@@ -0,0 +1,76 @@
...
+							checked={ metaBlockTypes.includes( i.value ) }
+							onChange={ ( checked ) => {
+								if ( checked ) {
+									setMeta( {
+										...meta,
+										[ META_KEY ]: [ ...metaBlockTypes, i.value ],
+									} );
+								} else {
+									setMeta( {
+										[ META_KEY ]: metaBlockTypes.filter( ( term ) => term !== i.value ),
```
ryelle, line 63: Do you need `...meta` here too? Maybe we need a `usePostMeta` hook so it can handle saving just the one meta key?

public_html/wp-content/plugins/pattern-creator/src/plugins/category-settings-panel.js
```diff
@@ -0,0 +1,65 @@
...
+import usePostData from '../hooks/use-post-data';
+
+/**
+ * Module constants
+ */
+const PATTERN_CAT_KEY = 'pattern-categories';
+
+const CategorySettingsPanel = () => {
+	const categories = useSelect( ( select ) =>
+		select( coreStore ).getEntityRecords( 'taxonomy', 'wporg-pattern-category', { per_page: 20 } ) || []
```
ryelle: Should we just load all categories, `-1`?

public_html/wp-content/plugins/pattern-creator/src/plugins/main-dashboard-button.js
```diff
@@ -0,0 +1,11 @@
+import { registerPlugin } from '@wordpress/plugins';
+/* eslint-disable-next-line @wordpress/no-unsafe-wp-apis -- Experimental is OK. */
+import { __experimentalMainDashboardButton as MainDashboardButton } from '@wordpress/edit-post';
+
+registerPlugin( 'main-dashboard-button-test', {
+	render: () => (
+		<MainDashboardButton>
+			<a className="main-dashboard-button" href="http://wordpress.org">{ '←' }</a>
```
ryelle: Ideally this will go back to the previous page, which might be the homepage, or it might be a pattern details page. That can happen in a followup PR.

public_html/wp-content/plugins/pattern-creator/src/plugins/save-post-modifier.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * WordPress dependencies
+ */
+import { dispatch, select } from '@wordpress/data';
+import { store } from '@wordpress/editor';
+import { useEffect, useState } from '@wordpress/element';
+import { registerPlugin } from '@wordpress/plugins';
+
+window.gutenbergSavePost = dispatch( store ).savePost;
```
ryelle, line 9: Why save this as a global on window, rather than a local const or something?

public_html/wp-content/plugins/pattern-creator/src/plugins/save-post-modifier.js
```diff
@@ -0,0 +1,56 @@
...
+		dispatch( store ).savePost = ( props ) => {
+			if ( props && props.isAutosave ) {
+				return;
+			}
+
+			// We need to wait until now to get the status otherwise we will be working with a previous context
+			const postStatus = select( store ).getEditedPostAttribute( 'status' );
+
+			// The post is set to 'publish' before 'savePost' is called,
+			// we can rely on that to make sure we only show the window when necessary
```
ryelle, line 35: I'm not sure I follow this logic - what sets the `postStatus` to publish?

public_html/wp-content/plugins/pattern-creator/src/plugins/save-post-modifier.js
```diff
@@ -0,0 +1,56 @@
...
+			if ( 'publish' === postStatus ) {
+				setShowModal( true );
+			} else if ( 'auto-draft' === postStatus ) {
```
ryelle: This one makes sense to me - if it's currently an auto-draft, we can keep saving the draft as normal (same can probably be done for `draft`).

public_html/wp-content/plugins/pattern-creator/src/plugins/viewport-header-control.js
```diff
@@ -0,0 +1,174 @@
...
+					onClose: updateControlWithEditedData,
+					onFocusOutside: updateControlWithEditedData,
+				} }
+				label={ __( 'Select a viewport width', 'wporg-patterns' ) }
+			>
+				{ ( ) => (
+					<>
+						<MenuGroup label="Viewport Width">
+							<p className="viewport-header-control__copy">
+								This is used when displaying a preview of your pattern.
```
ryelle: Missing i18n wrapper here

public_html/wp-content/plugins/pattern-creator/src/plugins/viewport-header-control.js
```diff
@@ -0,0 +1,174 @@
...
+				) }
+				popoverProps={ {
+					onClose: updateControlWithEditedData,
+					onFocusOutside: updateControlWithEditedData,
+				} }
+				label={ __( 'Select a viewport width', 'wporg-patterns' ) }
+			>
+				{ ( ) => (
+					<>
+						<MenuGroup label="Viewport Width">
```
ryelle: Missing i18n wrapper here, too

public_html/wp-content/plugins/pattern-creator/src/style.css
```diff
@@ -1,54 +1,90 @@
...
+
+/* Override gutenberg component style */
+.viewport-header-control .components-menu-group__label {
+	color: #2f2f2f;
 }
 
-.interface-interface-skeleton__footer {
-	position: absolute;
+.viewport-header-control__copy {
+	font-size: 10px;
```
ryelle: This is _very small_ in practice. What do you think, @shaunandrews?

Screen Shot 2021-06-01 at 5 39 58 PM

(I don't think changing this needs to block merge)

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
```diff
@@ -130,16 +130,17 @@ function register_post_type_data() {
 		POST_TYPE,
 		'wpop_viewport_width',
 		array(
-			'type'              => 'number',
+			'type'              => 'string',
+			'pattern'           => '[0-9]+(px|%|rem|em|vw)',
```
ryelle: The `viewportWidth` when used in patterns [needs to be an integer](https://github.com/WordPress/gutenberg/tree/94db813bc53f52bd28268fd3f7b4daf7982109cb/packages/block-editor/src/components/block-preview#viewportwidth), so this value needs to stay a number.

public_html/wp-content/plugins/pattern-directory/package.json
```diff
@@ -9,7 +9,7 @@
 		"format:js": "wp-scripts format-js",
 		"lint:css": "echo \"No CSS.\"",
 		"lint:js": "wp-scripts lint-js",
-		"start": "wp-scripts start pattern-post-type=./src/pattern-post-type.jst"
+		"start": "wp-scripts start pattern-post-type=./src/pattern-post-type.js"
```
ryelle: Can you push this change to trunk & remove it from here? No need for another PR.

public_html/wp-content/plugins/pattern-directory/src/pattern-post-type.js
```diff
@@ -11,39 +11,41 @@ import { useDispatch, useSelect } from '@wordpress/data';
 const PatternPostType = () => {
 	const { editPost } = useDispatch( 'core/editor' );
 	const postMetaData = useSelect( ( select ) => select( 'core/editor' ).getEditedPostAttribute( 'meta' ) || {} );
+	const postTitle = useSelect( ( select ) => select( 'core/editor' ).getEditedPostAttribute( 'title' ) || '' );
 	const [ description, setDescription ] = useState( postMetaData.wpop_description );
-	const [ viewportWidth, setViewportWidth ] = useState( postMetaData.wpop_viewport_width );
```
ryelle, line 16: Do you want to use `usePostData` here?

ryelle (commented):
Looking great so far! Left a few comment/questions, most things wouldn't block a merge but I tried to mention that.

I did notice this extra scrollbar on some container when I scrolled. It would be great to clean that up if we can.

Screen Shot 2021-06-01 at 5 21 16 PM

public_html/wp-content/plugins/pattern-creator/src/components/unit-control/index.js
StevenDufresne: Ha, dang. Getting all the units working was a big pain. :)

public_html/wp-content/plugins/pattern-creator/src/plugins/main-dashboard-button.js
StevenDufresne: Yeah, I was going to tie into `history.back()` but I'm still not certain I understand all ways into the editor.

public_html/wp-content/plugins/pattern-creator/src/plugins/category-settings-panel.js
StevenDufresne: Until we have a load more functionality we'll probably need to slice the list no matter what, otherwise it will make the Block editor pretty unusable.

public_html/wp-content/plugins/pattern-creator/src/plugins/block-scope-settings-panel.js
StevenDufresne: Nice catch. Thanks.

public_html/wp-content/plugins/pattern-creator/src/api-middleware/filter-endpoints.js
StevenDufresne: 🤔 No, I didn't add this.

public_html/wp-content/plugins/pattern-creator/src/plugins/category-settings-panel.js
ryelle: Categories is a fixed set of items though, so we shouldn't have an unusable amount. (I agree this is a problem for keywords, though)

StevenDufresne:
> Looking great so far! Left a few comment/questions, most things wouldn't block a merge but I tried to mention that.
> 
> I did notice this extra scrollbar on some container when I scrolled. It would be great to clean that up if we can.

Can you be more specific? I do see a scrollbar when I try to scroll the editor, but I also see the same one when using Gutenberg for post/pages.

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
StevenDufresne: Ah, I guess I misread: https://github.com/WordPress/pattern-directory/issues/59#issuecomment-849039740.
It has to be an integer. I'm going to reverse out the unit support and force `px`s.

ryelle:
I turned off the toolbar on the frontend (in your user profile, uncheck "Show Toolbar when viewing site"), and it's much more obvious (also we should handle the case of no toolbar better 😆). You can see as I scroll up and down on the header, this scrollbar appears.

https://user-images.githubusercontent.com/541093/120409952-0b3f4e00-c320-11eb-86fe-120eef288420.mov

public_html/wp-content/plugins/pattern-creator/src/plugins/category-settings-panel.js
StevenDufresne: That's fair, but I don't know that we would ever want more than 20 categories to show up. So if the backend gets more categories (which there is a design for) we don't need to come back here in a hurry, although at that point the first page probably wouldn't be representative.

public_html/wp-content/plugins/pattern-creator/src/plugins/save-post-modifier.js
StevenDufresne: The Gutenberg `publish` button does. 

https://github.com/WordPress/gutenberg/blob/022d624665bb60e63a29cf235e788bf5d6a30845/packages/editor/src/components/post-publish-button/index.js#L138

public_html/wp-content/plugins/pattern-directory/src/pattern-post-type.js
StevenDufresne: Since it's technically a separate codebase, I decided not to share it. It's not really necessary.... yet.

public_html/wp-content/plugins/pattern-creator/src/style.css
StevenDufresne: Also noting that since this uses `@wordpress/components/UnitControl`, "Viewport Width" label is not 'hideable' like other input labels are for Gutenberg input components.

StevenDufresne:
> I turned off the toolbar on the frontend (in your user profile, uncheck "Show Toolbar when viewing site"), and it's much more obvious (also we should handle the case of no toolbar better 😆). You can see as I scroll up and down on the header, this scrollbar appears.

Got it. It doesn't appear in chrome because the `is-fullscreen-mode` class is being set as expected. It looks like in firefox it's not adding that class? I don't know what the difference is but I made 2 changes:

Remove missing CSS style: 315f247
Enforce `top:0` positioning: 63d68b2

 Can you try again?

public_html/wp-content/plugins/pattern-creator/src/plugins/category-settings-panel.js
StevenDufresne: Updated to `-1`.

public_html/wp-content/plugins/pattern-directory/src/pattern-post-type.js
ryelle: 🙃  I was looking through so many files I stopped looking at filepaths. Sounds good.

public_html/wp-content/plugins/pattern-creator/src/style.css
ryelle: Since this is value can't have units, could we switch to a NumberControl or TextControl? would that let you hide the label?

ryelle:
The classname isn't a browser issue, it's a setting- I have fullscreen off, so I see the toolbar in Chrome & FF. If I switch to fullscreen, the toolbar goes away, but I still see a scrollbar issue. Ideally it would force the fullscreen and remove this option, but we can figure all this out in a followup PR.

Screen Shot 2021-06-02 at 11 09 19 AM

I see another issue that needs to be addressed before we can merge this approach. Clicking "Save Draft" updates the page URL to `site.com/new-pattern/post.php?…`, which is a page that doesn't exist. A better flow would be to update the URL to the correct `/pattern/{whatever}/edit`, or at least disable the change. This came from `BrowserURL` when I looked into this last year (https://github.com/WordPress/pattern-directory/pull/26).

shaunandrews:
I noticed the following error appear while the editor was loading:

> Deprecated: allowed_block_types is deprecated since version 5.8.0! Use allowed_block_types_all instead. in /var/www/html/wp-includes/functions.php on line 5381 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-includes/functions.php:5381) in /var/www/html/wp-admin/admin-header.php on line 9

shaunandrews:
Very cool to see this coming together. I'm not sure if you need/want detailed design feedback yet, but here's a few things that stood out to me as I tinkered with the current state of this branch:

--

There's some weird spacing happening with the Heading block's level menu.

image

--

Adding a Navigation block lets me then add "All Pages" from the directory site. This is pretty awkward, and I'm not entirely sure what to do about it right now.
image

--

I think we should treat viewportWidth like `max-width` within the editor. Otherwise, the pattern canvas can extend outside the viewport and I can't scroll horizontally.

image

--

At smaller viewports the Preview button appears. Clicking it opens a new tab, but it never actually loads the preview. I think we should nix the Preview button entirely.

image

--

For some reason the Social Icons and Navigation blocks seem to always have a white background, even when they're nested inside a Group block with it's own background color.

image

--

Its fine for now (don't let is slow you down!) but hopefully we can get closer to matching the designs. In the comparison (PR / Design) below notice the font sizes, font weights, spacing, and icon are all different.

image

--

The viewport Width input should always have a value. If one of the predefined options (Narrow, Default, Full) are selected I'd expect the input to reflect that value as well. If I do set a custom value, the predefined option should show as _unchecked._

viewport-width

public_html/wp-content/plugins/pattern-creator/src/style.css
shaunandrews: We can probably follow the lead from the Tools menu, which uses a 13px font-size.

image

ryelle:
For the navigation block, we have an issue for creating placeholder content for all the dynamic blocks: https://github.com/WordPress/pattern-directory/issues/23

public_html/wp-content/plugins/pattern-creator/src/style.css
StevenDufresne: Moved to `px` and manually hid the label in 2d3c6dd.

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
StevenDufresne: Done in: 2d3c6dd

StevenDufresne:
> The viewport Width input should always have a value. If one of the predefined options (Narrow, Default, Full) are selected I'd expect the input to reflect that value as well. If I do set a custom value, the predefined option should show as _unchecked._
> 
> viewport-width

Updated:
1. MaxWidth: e41def5
2. Pass in current value to unit control input: ef4e965

StevenDufresne:
> I see another issue that needs to be addressed before we can merge this approach. Clicking "Save Draft" updates the page URL to `site.com/new-pattern/post.php?…`, which is a page that doesn't exist. A better flow would be to update the URL to the correct `/pattern/{whatever}/edit`, or at least disable the change. This came from `BrowserURL` when I looked into this last year (#26).

Can this be sorted with a rewrite? I don't think there's a sane way to stop it from happening on the front apart from changing it back.

iandunn:
I haven't dug deep into it, so take this all w/ a grain of salt, but a rewrite sounds like a good workaround if the JS isn't configurable/overwritable.

Some alternative ideas if any of them seem useful:
* submit a PR to Gutenberg to make the component more customizable. maybe adding a JS filter to `getPostEditURL()`
* add a JS event handler to catch the location change, prevent the default action, and go directly to the correct URL. or one to run _after_ the location change, and update the URL to the correct one
* replace the Save Draft button w/ a custom one, if that's possible?
* Could maybe also use a PHP filter the edit-post URL on before it's passed to JS

I'd rely on Kelly's thoughts more than any of that, though.

public_html/wp-content/plugins/pattern-creator/src/plugins/save-post-modifier.js
StevenDufresne: For some reason, Gutenberg will re-render plugins. I think it's because parent components are unmounted. So if they reload, when it _re-stashes_ the function, it's already the one we replaced the default with and we never have a copy of the original function.

public_html/wp-content/plugins/pattern-creator/src/plugins/save-post-modifier.js
StevenDufresne: Yep, we need to support editing an existing pattern. I haven't been able to get the editor to work in that context yet.

StevenDufresne:
Much appreciated!

> Some alternative ideas if any of them seem useful:
> 
> * submit a PR to Gutenberg to make the component more customizable. maybe adding a JS filter to `getPostEditURL()`

This seems like a good idea. I wonder how long it would take to do this (or is it possible?). @ryelle Any opinion on this?

> * add a JS event handler to catch the location change, prevent the default action, and go directly to the correct URL. or one to run _after_ the location change, and update the URL to the correct one

Can't do much here since we call `replaceState` directly in JS (we can't listen for [popstate](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onpopstate)) and do nothing in Gutenberg's state object. We could maybe replace `window.history.replaceState` but it's really risky and we'll most likely get the intercept wrong in some contexts.

> * replace the Save Draft button w/ a custom one, if that's possible?

We can replace it but the control that updates the URLs listens for a change to `draft` and then triggers so we wouldn't be able to change the post status.

> * Could maybe also use a PHP filter the edit-post URL on before it's passed to JS

Not familiar with how this would work. @ryelle Any opinion on this?

ryelle:
> Can this be sorted with a rewrite?

That would work, though I don't love the idea. We'll need to check for a post ID and load in the pattern content into the editor on new-pattern as well as when editing a pattern, but it's a good workaround.

> Could maybe also use a PHP filter the edit-post URL on before it's passed to JS

Unfortunately it's not pulled from PHP, the redirect to `post.php` is [hardcoded in `getPostEditUrl`](https://github.com/WordPress/gutenberg/blob/4834fad30e39e47d71a93e2e08d31b3b856711e1/packages/edit-post/src/components/browser-url/index.js#L16-L18).

> > submit a PR to Gutenberg
>
> I wonder how long it would take to do this (or is it possible?)

It's definitely possible, but probably depends on whether we can prove our use-case to whoever reviews it. We _have_ a real use-case here, so that's good. If you do create an issue, I suggest explaining the use case and making some suggestions, rather than jumping to a PR. We could shortcut out of `BrowserUrl` entirely, or add a JS filter, or figure out some way to get the URL from the post API, so that we can edit the url in PHP.

This feels like a pretty big blocker to this approach, so I'd like to have a plan before this is merged.

StevenDufresne:
I was able to get it working using redirects. I don't think it's the best approach, I think getting something into Gutenberg would be a better approach long term but it does work and isn't a bad experience.

What do y'all think?

**Here is the code:** 
https://github.com/WordPress/pattern-directory/compare/try/edit-post-package-first...try/edit-post-package-redirect?expand=1

**Here is what it looks like functionally:**
 [View GIF](https://d.pr/i/dcAevF)

iandunn:
🤔 , I wonder if the explicit `RewriteRule` in `.htaccess` and `add_rewrite_rule()` are necessary? Production uses nginx, so we'd need a sysreq to have similar functionality, but I'm guessing the `wp_safe_redirect()` would work w/out it.

> `if ( isset( $_GET['post'] ) && isset( $_GET['action'] ) && 'edit' === $_GET['action'] ) {`

You might need to make sure that the post ID is a pattern post type, and that the user isn't in wp-admin, to avoid redirecting other post types and manually adding patterns.

It's odd that there's the FOUC, doing the `wp_safe_redirect()` in `init` should short-circuit that 🤔

ryelle:
I think that `RewriteRule` pops up any time you refresh permalinks, I've had to discard it before.

Do we need a sysreq for `add_rewrite_rule`? I thought WP itself handles those. Would the redirect to `/pattern/[ID]/edit/` work without a rewrite rule to handle it?

> It's odd that there's the FOUC, doing the wp_safe_redirect() in init should short-circuit that

I think that's the page before the JS content loads, not to do with the redirect, unless I'm misunderstanding FOUC (flash of unstyled content?).

> I don't think it's the best approach, I think getting something into Gutenberg would be a better approach long term but it does work and isn't a bad experience.

Yeah, I agree.

StevenDufresne:
> 🤔 , I wonder if the explicit `RewriteRule` in `.htaccess` and `add_rewrite_rule()` are necessary? Production uses nginx, so we'd need a sysreq to have similar functionality, but I'm guessing the `wp_safe_redirect()` would work w/out it.

Yeah, the `.htaccess` rule seems to be automatically added on permalinks save. I reset the file, resaved the permalinks and it came back so I left it in. I have no clue why it keeps getting re-added. I just assumed it's something .`wp-env` needs. ❓ ❓

> > `if ( isset( $_GET['post'] ) && isset( $_GET['action'] ) && 'edit' === $_GET['action'] ) {`

> You might need to make sure that the post ID is a pattern post type, and that the user isn't in wp-admin, to avoid redirecting other post types and manually adding patterns.

Good call, I'll make it more specific.

> It's odd that there's the FOUC, doing the `wp_safe_redirect()` in `init` should short-circuit that 🤔

Yeah, that flash is the page itself, it's technically the wp header on the page without styles. The react app appears on top once the JS loads. We'll need to add a better loading state for that but the redirect does work as expected.

iandunn:
Ah, I see. You're right, we'd only need a sysreq if we needed an nginx rewrite similar to the `.htaccess` one.

StevenDufresne:
I've added the redirect code to this branch and I think I've closed out all the main issues identified.

public_html/wp-content/plugins/pattern-creator/pattern-creator.php
```diff
@@ -13,7 +13,9 @@
 namespace WordPressdotorg\Pattern_Creator;
 use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\POST_TYPE;
 
+const AUTOSAVE_INTERVAL = 30;
 const QUERY_VAR = 'edit-pattern';
+const PATTERN_ID_VAR = 'patternId';
```
ryelle: Typically keys like this use kebab-case in PHP.

```suggestion
const PATTERN_ID_VAR = 'pattern-id';
```

public_html/wp-content/plugins/pattern-creator/pattern-creator.php
```diff
@@ -144,3 +189,17 @@ function inject_editor_template( $template ) {
...
 }
 add_filter( 'template_include', __NAMESPACE__ . '\inject_editor_template' );
+
+/**
+ * Add a rewrite rule to handle editing a pattern.
+ */
+function rewrite_for_pattern_editing() {
+	add_rewrite_rule( '^pattern/(\d+)/edit', 'index.php?pagename=new-pattern&' . PATTERN_ID_VAR . '=$matches[1]', 'top' );
+
+	if ( isset( $_GET['post'] ) && isset( $_GET['action'] ) && 'edit' === $_GET['action'] && POST_TYPE !== get_post_type( $_GET['post'] ) && ! is_admin() ) {
```
ryelle: This would be easier to read on multiple lines, and I think there's a logic error here, should it be `POST_TYPE === get_post_type( $_GET['post'] )`?

```suggestion
	if ( 
		isset( $_GET['post'] ) &&
		isset( $_GET['action'] ) &&
		'edit' === $_GET['action'] &&
		POST_TYPE !== get_post_type( $_GET['post'] )
		&& ! is_admin()
	) {
```

public_html/wp-content/plugins/pattern-creator/pattern-creator.php
```diff
@@ -82,9 +98,31 @@ function enqueue_assets() {
 		$post_id = $post->ID;
 	}
 
+	// Update the post id if we're editing a pattern.
+	if ( should_load_creator() && is_editing_pattern() ) {
+		$post_id = get_query_var( PATTERN_ID_VAR );
```
ryelle, line 103: You should also reset `$post` here too, since it's used in the `allowed_block_types` filter below.
```suggestion
		$post_id = get_query_var( PATTERN_ID_VAR );
		$post    = get_post( $post_id );
```

public_html/wp-content/plugins/pattern-creator/src/plugins/viewport-header-control.js
```diff
@@ -0,0 +1,182 @@
...
+	if ( ! btnDomRef ) {
+		return;
+	}
+
+	// We may re-insert the same button if it state's changes
+	maybeRemoveContainer();
+
+	insertContainer( btnDomRef );
+
+	/* eslint-disable no-undef*/
```
ryelle: What's undefined here? We can pull `render` from `@wordpress/element` if that's it?

ryelle (approved):
A few small notes, but once those are addressed I think this can be merged so we can iterate on it in trunk.

public_html/wp-content/plugins/pattern-creator/src/plugins/save-post-modifier.js
ryelle: I think this is also affecting the "Save Draft" button -- if i click Publish, it does nothing (expected, i think?), but then clicking Save Draft also does nothing. Only noting so we can make sure Save Draft works as expected with the publish modal (it should not trigger the modal).

public_html/wp-content/plugins/pattern-creator/src/plugins/viewport-header-control.js
StevenDufresne: Yeah, this is `ReactDOM`.

public_html/wp-content/plugins/pattern-creator/pattern-creator.php
StevenDufresne: Meeep.. that was a test. 🙇

StevenDufresne: By the way, I formatted it using `format:php`... it doesn't match the example above.

ryelle: Does it reformat the multiple-line version? The phpcs linter fix isn't as aggressive as prettier, so it's still up to the human developer to make code style decisions 🙂

StevenDufresne: It does a multi-line format but the first`isset` is pulled up into the line as`if(`... etc...

PR #119, opened 2021-06-02
Patterns: Use CopyPatternButton in thumbnail

ryelle:
This consolidates the copy code, removing the duplicate code from PatternThumbnail. There should be no functionality changes here.

### Screenshots

Screen Shot 2021-06-02 at 3 08 00 PM

1. Make sure the copy buttons in the grid and on single patterns still work as expected.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/src/components/copy-pattern-button/index.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/copy-pattern-button/index.js
@@ -2,34 +2,60 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
+import { speak } from '@wordpress/a11y';
 import { Button } from '@wordpress/components';
+import { useEffect, useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
  */
 import { copyToClipboard } from '../../utils';
 
-const CopyPatternButton = ( { onSuccess } ) => {
-	const handleClick = ( { target } ) => {
+const CopyPatternButton = ( { isSmall = false, onSuccess, content } ) => {
+	const [ copied, setCopied ] = useState( false );
+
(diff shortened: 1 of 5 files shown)


public_html/wp-content/themes/pattern-directory/src/components/copy-pattern-button/index.js
```diff
@@ -1,35 +1,71 @@
...
+		const success = copyToClipboard( content );
 
-		const success = copyToClipboard( blockPattern );
+		setCopied( success );
 
 		// Make sure we reset focus in case it was lost in the 'copy' command.
 		target.focus();
 
-		if ( success ) {
+		if ( success && 'function' === typeof onSuccess ) {
```
StevenDufresne: Would it make sense to set a default value for `onSuccess` to communicate its optionality as opposed to check it as a function?  When I see this I always think.. "What else could it be?"

StevenDufresne (approved):
Minor comment.

Everything still works as expected.

public_html/wp-content/themes/pattern-directory/src/components/copy-pattern-button/index.js
ryelle: Well, it could be undefined. I don't think a default value makes sense, because we don't _need_ to do anything here. I initially had this as just `! onSuccess`, but I figured why not be a little more specific.

StevenDufresne: Gutenberg controls typically use `lodash.noop`:

https://github.com/WordPress/gutenberg/blob/cdbab7fff0d285787a565fe0e23b620b28ab1922/packages/components/src/input-control/index.js#L37

PR #151, opened 2021-06-09
Theme: Make Menu & Search components more universal

ryelle:
The `CategoryMenu` & `CategorySearch` components don't need to be tied to "category", since they both accept options that can display different content. This PR switches them to `Menu` and `Search`. `Menu` will (in another PR) be used on the My Patterns and Favorites pages, see #16.

### Screenshots

No visual changes.

1. View any grid pages
2. The category navigation and search should work as expected


public_html/wp-content/themes/pattern-directory/src/components/search/index.js
```diff
@@ -3,34 +3,36 @@
...
 import { __ } from '@wordpress/i18n';
 import { Icon, search } from '@wordpress/icons';
+import { useInstanceId } from '@wordpress/compose';
 
-const CategorySearch = ( { defaultValue, isLoading, isVisible, onUpdate, onSubmit } ) => {
+const Search = ( { defaultValue, isLoading, isVisible, onUpdate, onSubmit } ) => {
+	const instanceId = useInstanceId( Search );
 	if ( isLoading ) {
-		return <span className="category-search--is-loading" />;
+		return <span className="search is-loading" />;
```
StevenDufresne, line 11: I feel like the class `search` will have some old rules interfering when we get to .org since it's a common class name.

ryelle: That's fair - I was back-and-forth about whether it should be `pattern-search` as the class name, but I don't think we've been consistent about that (requiring `pattern-` a prefix for component classNames). What do you think?

StevenDufresne: Yeah, we haven't been consistent. I'm good with moving to a `pattern` prefix or maybe just `wporg`?

StevenDufresne (approved):
This looks good and is working as expected. Go for it!

public_html/wp-content/themes/pattern-directory/src/components/search/index.js
ryelle: We've at least been somewhat consistent with `pattern`, so I'll switch to that and maybe write up a short doc we can refer to.

PR #153, opened 2021-06-09
Add UI for unlisting a pattern from the directory

### LINKED ISSUE #107: Block Editor: Add UI for switching a pattern to a custom status
In #75 we introduced some custom post statuses for patterns as part of the moderation flow. However, there isn't a way to apply these statuses from the block editor.

Ideally, I think, we'd add our own status dropdown with all of the options specified in #56.

Related: #105

coreymckrill:
Adds a way to change the status of a pattern to "unlisted" from within the block editor. An unlisted pattern can then be republished by clicking the standard Publish button. An unlisted pattern has a notice at the top of the editor explaining that it won't be included in the public pattern directory. This workflow is intended for moderators who need to review patterns that have been flagged by users as inappropriate.

This also enhances the Patterns list table so that the Unlisted table view has an accurate count of the number of unlisted posts, but unlisted posts are not included in the "All" view for the table.

Originally this was going to add a dropdown where you could choose between different post statuses. However, the Block Editor makes this awkward to accomplish since the built in statuses of draft, pending, and publish are all set from different parts of the UI. It also seemed to me that we didn't really need multiple custom statuses, and "Unlisted" could probably cover all of our use cases.

Fixes #107

### Screenshots

Screen Capture on 2021-06-14 at 11-38-31

Screen Shot 2021-06-14 at 11 41 37

1. Create and publish a pattern
2. Go back to the Edit screen for the pattern and click the Unlist button
3. A notice should appear at the top of the block editor explaining that the pattern is now unlisted.
4. Click the Publish button to re-list the pattern.
5. Try unlisting patterns that are in the draft and pending statuses as well.
6. With at least one pattern marked as unlisted, go to the Patterns list table. Unlisted patterns should not be included in the "All" view, but should appear when specifically viewing Unlisted.


ryelle:
I can check out the code more tomorrow, but after a quick UI test I have some feedback.

Is it possible to update the "visibility" here? or add a note in place of the Unlist button when a pattern is unlisted? I know there's a big banner, but this is where I would check for status.

Screen Shot 2021-06-14 at 6 00 49 PM

When I unlist a pattern it says it's now a draft. Is it possible to provide a different string to this message, to make it clear it's unlisted?

Screen Shot 2021-06-14 at 6 01 01 PM

coreymckrill:
@ryelle I can't see a way to modify the Visibility section of that panel, or the snackbar message, unfortunately. Any ideas, @StevenDufresne?

The one thing I can do is add a note in place of the "Unlist" button, to go along with the banner message. I'll work on that.

StevenDufresne:
> @ryelle I can't see a way to modify the Visibility section of that panel, or the snackbar message, unfortunately. Any ideas, @StevenDufresne?

It would be nice to add a custom post type visibility, but that's not supported (yet?):
https://github.com/WordPress/gutenberg/issues/13794

I don't think you can do much for the snackbar:
https://github.com/WordPress/gutenberg/blob/fd3df7c247fc365bc64cbfb8974896145aef9fd2/packages/editor/src/store/utils/notice-builder.js#L43

A couple of ideas:
- Could we hijack the "private" visibility and set the post to "unlist" within a specific context and use that as the 'unlisting' mechanism?
  - Use some kind of filter that listens for the post status change to `private`? 
- The current `unlist` button/link is very unassuming, could we use the split button approach we used in the [revisions-extended](https://github.com/WordPress/revisions-extended) plugin?
  - And turn off `Status & Visibility` altogether?

coreymckrill:
> Could we hijack the "private" visibility and set the post to "unlist" within a specific context and use that as the 'unlisting' mechanism?

Are we still planning to enable people to save private patterns to the directory and not necessarily publish them to the public listing? If so, using the `private` status for unlisting would probably cause a collision there...

> And turn off Status & Visibility altogether?

We could probably remove the default panel and build our own that still has some of the default controls like the Pending checkbox, but without the Visibility setting. Seems kind of overkill, but that may be the only way.

coreymckrill:
Interesting, here's how Edit Flow handles its custom post statuses feature in the block editor:

extended-post-status-1

extended-post-status-2

We could implement something similar, but specific to unlisting, since I don't think we will need any other custom statuses.

ryelle:
> Are we still planning to enable people to save private patterns to the directory and not necessarily publish them to the public listing? If so, using the private status for unlisting would probably cause a collision there...

Yeah, the private status already has a meaning, and unlisted is different.

> The one thing I can do is add a note in place of the "Unlist" button, to go along with the banner message. I'll work on that.

I think that would be fine, just enough to note that it's unlisted.

> Edit Flow

This seems a little overkill for just unlisting, but I'm not sure what you're picturing here.

coreymckrill:
Alright, updated to add notes explaining both what the button will do, and what it did.

unlist-button-notice

unlisted-notice

ryelle (approved):
Looks good now, thanks for adding the Unlisted notice.

PR #152, opened 2021-06-09
Create the "My Patterns" page

ryelle:
See #16. This adds a single My Patterns page, and lays some groundwork for the Favorite page.

### Screenshots

Screen Shot 2021-06-10 at 13 51 57-fullpage

On the drafts & pending pages, those are highlighted:

Screen Shot 2021-06-10 at 1 51 41 PM

### Notes

This doesn't complete #16, I'm aiming to get there in a series of smaller PRs. Still on my radar:

- Make the Favorites page
- Add the order by dropdown to the menu
- Add the pattern counts to the statuses in the menu
- Update the pattern thumbnail designs

### How to test the changes in this Pull Request:

1. Create a page called "My Patterns", it should have the slug `/my-patterns`.
2. View the page, you should see only patterns by your account
3. Try it logged out (this likely needs some design input, I'll make an issue for it once this merges)

PR #169, opened 2021-06-11
Turn off full width pattern-grid-menu decendants on desktop.

### LINKED ISSUE #165: Pages with Menu overflow horizontally.
**Describe the bug**
If you load `/` on trunk, the menu will cause a horizontal scrollbar on desktop because of this style:

https://github.com/WordPress/pattern-directory/blob/e69788eb194e390ecb98b78e00861028a83cf892/public_html/wp-content/themes/pattern-directory/css/components/_pattern-grid-menu.scss#L9

StevenDufresne:
This PR limits a `width:100%` rule to smaller viewports where it's necessary. 

@ryelle This was introduced in #151 which was in preparation for the work in #152. Can you verify that it doesn't regress anything for that PR?

Fixes #165

### Screenshots
N/A

1. Load `/`
2. Expect to not see a horizontal navigation bar.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern-grid-menu.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern-grid-menu.scss
@@ -15,7 +15,7 @@
 		width: calc(100% - #{$gutter-default * 2});
 	}
 
-	@media only screen and ( min-width: $breakpoint-medium ) {
+	@media only screen and ( min-width: #{$breakpoint-medium + 1} ) {
 		margin: $gutter-default;
 		flex-direction: row;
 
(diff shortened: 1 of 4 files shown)


ryelle:
Good catch with #165, I don't seem to have that problem in firefox. I tweaked the CSS a little here, since we can apply the `100%` just to the element that needs it, and avoid making the screen-reader-text also 100%. While testing I also noticed that the viewport rules don't work like we both assumed, so both mobile and default menus were visible at 600px. This is because the `viewportMatch` rules all fall back to media queries, and both `min-width: 600px` and `max-width: 600px` are true at 600px wide.

Anyway, I switched it to use a single `useViewportMatch`, so we can be sure that it will either be true or false.

Edit: I wrote 600 because [I was testing out the various viewport match functions](https://codesandbox.io/s/dark-meadow-bs5wj?file=/src/App.js&resolutionWidth=600&resolutionHeight=765) with `small`, but it should be 782px.

ryelle (approved):

PR #164, opened 2021-06-10
Pattern Grid: Update design of pattern cards

ryelle:
Fixes #103 - This updates the UI of the pattern cards to match the new design:
- Title & meta are always visible, only adding favorite count for now since we don't have views
- Show small author avatar
- Show a status label on draft/pending/declined patterns (seen in My Patterns)
- Fix an issue with encoded characters in pattern titles (ex, `&amp;`)
- Fix extra space in the preview — should fix "Extra space on the left/right of full-width items" in #156 

❗  This uses a fake favorite value for now, to get the UI working. I think we'll need a solution with caching to get this value for real, so that discussion can happen in #100 

### Screenshots

Browsing patterns:
all-patterns

My patterns:
my-patterns

Responsive styles - drops to 2 column around 770px, and to 1 col around 525px.
all-patterns-small

all-patterns-mobile

### How to test the changes in this Pull Request:

- Build the branch with `yarn workspaces run build`
- Make sure you have a page called My Patterns (see #152), and a handful of patterns by your user
- View the pattern archives (all and categories)
- They should match the figma design
- View My Patterns
- The patterns should match the design, draft & pending patterns should be labelled
- Resizing the screen shouldn't break anything

Issue #162, opened 2021-06-10
Link colors aren't displaying


melchoyce:
**Describe the bug**
Link colors aren't being parsed on the front-end of the patterns: https://wordpress.org/patterns/pattern/event/

**To Reproduce**
1. Create a new pattern. 
2. Add a cover block with a dark background, and inside, add a link.
3. Change the link color.
4. Preview the pattern and see that the link color you chose isn't being displayed.

**Expected behavior**
The link would be the color I chose in the editor.

**Screenshots**

Editor | Front-end
------------ | -------------
![image](https://user-images.githubusercontent.com/2846578/121594633-58e84480-ca0b-11eb-8be2-c9b161a32527.png) | ![image](https://user-images.githubusercontent.com/2846578/121594674-66053380-ca0b-11eb-9f92-b755259d6c83.png)

ryelle:
I'm pretty sure the same issue is what's causing the social links thing mentioned in #156

ryelle:
@melchoyce is this working now? I believe #177 should have fixed it.

melchoyce:
Yup! Thanks 👍

Issue #156, opened 2021-06-10
Pattern display bugs


kjellr:
As we continue populating the directory, we're running into some visual bugs with the pattern previews. I'm consolidating a list here of bugs.

---

- [x] **Display is cut off**

1. https://wordpress.org/patterns/pattern/centered-introduction-text-with-button-and-offset-images/
2. https://wordpress.org/patterns/pattern/title-area-with-image-and-heading/
3. https://wordpress.org/patterns/pattern/introduction-with-image-on-the-left/

Editor|Preview
---|---
![templates-editor](https://user-images.githubusercontent.com/1202812/121545710-c038d100-c9d8-11eb-84c2-5068d1321eab.gif)|<img width="1164" alt="Screen Shot 2021-06-10 at 10 42 54 AM" src="https://user-images.githubusercontent.com/1202812/121545714-c0d16780-c9d8-11eb-82b3-fecceadfd8c1.png">

<img width="1354" alt="Screen Shot 2021-06-10 at 2 22 27 PM" src="https://user-images.githubusercontent.com/1202812/121577221-50860e80-c9f7-11eb-98b2-961f6391195f.png">

<img width="1354" alt="Screen Shot 2021-06-10 at 2 22 12 PM" src="https://user-images.githubusercontent.com/1202812/121577207-4d8b1e00-c9f7-11eb-90f1-5fc5a935d8d1.png">

---

- [x] **Missing social links block**

1. https://wordpress.org/patterns/pattern/team-social-cards/
2. https://wordpress.org/patterns/pattern/contact-and-social-links/

Editor|Preview
---|---
<img width="1269" alt="Screen Shot 2021-06-10 at 10 48 30 AM" src="https://user-images.githubusercontent.com/1202812/121546620-6f75a800-c9d9-11eb-9df4-2f43a4ac43b3.png">|<img width="943" alt="121382990-1d6b4e80-c915-11eb-8fde-e3b26bbaf01a" src="https://user-images.githubusercontent.com/1202812/121546625-700e3e80-c9d9-11eb-9417-7b1c6ac7e27f.png">

Editor|Preview
---|---
[...]

PR #177, opened 2021-06-11
Fix visual bugs in pattern previews

ryelle:
This fixes a few of the visual bugs flagged in #156.

- Fix the padding on the container to remove the small space on the left/right of full-width content.
- Fix the vertical alignment of the preview so that small patterns stay centered, but tall patterns start at the top and scroll.
- Fix the rendering of patterns in the single view so that all the relevant processing is done, this fixes the link color in #162  and the social icons.

I think the wide/full width issue is due to the Twenty Twenty-One styles, I think it's fixed with the switch to Twenty Nineteen in #164.

### Screenshots

Screen Shot 2021-06-11 at 17 38 56-fullpage
Screen Shot 2021-06-11 at 17 40 09-fullpage

It's easiest to test this with a sandbox - push up the code and check the linked pages in #156. Otherwise, create patterns that replicate the issues, and check that this fixes them.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -182,3 +182,23 @@ function bypass_404_page( $preempt, $wp_query ) {
 	}
 	return $preempt;
 }
+
+/**
+ * Get the full, filtered content of a post, ignoring more and noteaser tags and pagination.
+ *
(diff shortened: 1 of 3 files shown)


StevenDufresne (approved):
This PR is definitely tricky to test even though I grabbed a production dataset. It may be better to get someone more familiar with the issues to test.

With that being said, the CSS makes sense so I'll approve it and let you decide whether we should get a second opinion.

ryelle:
I'll merge it so it can go to production for better testing.

Issue #224, opened 2021-06-23
Translation mechanism for submitted patterns


tellyworth:
Here's a proposal for handling translation of third-party user-submitted patterns in the directory.

Some of the pieces of the core pattern translation mechanism would be used, except GlotPress would not be involved at all.

When creating and submitting a new pattern, the user will choose a locale for their pattern. That will be stored in a taxonomy as described in #223.

Translating a pattern to a new language will be essentially the same as creating a copy of the original, editing the text as needed, and setting a different locale for the new pattern.

Translated patterns will have the `post_parent` field set to record the relationship to its original source.

In other words, translating a pattern is basically the same as remixing. This means that translating can incorporate non-text changes as needed. For example, images could be changed, and designs could be changed to suit right-to-left languages.

The main pieces needed for this are:

* [x] Create a locale ~taxonomy~ storage mechanism (#229)
* [x] Add UI for setting the locale when creating a pattern (#229)
* [x] #244 v1 w/ WP_Query
* [x] #244 v2 w/ ElasticSearch
* [x] Cron job that creates/updates patterns from GlotPress
- [x] #391
- [ ] #392

~Add a Translate This button, that: (#17)~
  * Opens a new creator window with a copy of the pattern
  * with the `post_parent` set to the original pattern post_id
  * and a different locale selected

coreymckrill:
I converted this to a tracking issue so we can connect and add other issues for specific parts of this.

ryelle:
Pulling this out of initial launch, since the remaining tasks are for post-launch.

>  Cron job that creates/updates patterns from GlotPress

This one's done, right?

iandunn:
Yeah, #250 did it. I marked it as done, thanks!

ryelle:
Once pattern creation is built, how should submissions work across languages? I'm assuming the creator will be available on all Rosetta sites, so a designer could create a Japanese pattern on ja.w.org/patterns, with no English original. Then again, someone could also create an English pattern on ja.w.org/patterns, so maybe we'll need a dropdown with the current site's locale as default.

How do we see the GlotPress translation flow working with future submitted patterns? Was the GlotPress flow just for this initial set of patterns, or is that the preference of translators?

varlese:
We had a small discussion about the proposed options for translation mechanisms [over on Make/Polyglots](https://make.wordpress.org/polyglots/2021/09/19/how-to-handle-block-pattern-translations/). From the feedback:
[...]

Issue #223, opened 2021-06-23
Core pattern translation mechanism


tellyworth:
Here's a proposal for supporting translation of **core** patterns in the pattern directory:

Canonical versions of WordPress core patterns should live in code, in a GitHub repo. A separate repo from WP core, so it can be updated and new patterns added outside of the core release cycle.

Those canonical patterns should be translated as usual with GlotPress. We'll use the same project name etc so there is no change to any existing translation work.

Core pattern translations will be automatically imported into the pattern directory database, either when they change or hourly.

Changes to the core pattern repo will be automatically sent to GlotPress and imported into the pattern directory database, again either hourly or at the time of the change.

All patterns in the directory will have a locale taxonomy to track the language.

Translated patterns will use the post_parent field to record the relationship to the original pattern.

The main pieces needed are:

* Create a new repo (or a subdirectory in this one?) and add the core patterns as PHP code
* Add a locale taxonomy in the pattern directory and populate it
* Add some code to set and make use of the post_parent field for translated patterns
* Create a cron/async script to import the core patterns from the new repo into GlotPress
* Create a cron/async script to export the translated patterns from GlotPress into the database (maybe combine the two scripts?)

See #34

coreymckrill:
Some locale thoughts/questions, also relevant to #224:

* Would we want the full list of locales from `GP_Locale`, or only top level ones? (E.g. `es_ES`, `es_MX`, `es_PE` vs. simply `es`)
  * If we made the full list available, we could still have the search prioritize results based on the top level locale, like we do for workshops on WP Learn
* If we use a taxonomy, we have to store the terms in the wp_terms database table, and they could potentially be edited via WP Admin (though maybe there would be a way to turn that off?). If we used a post meta value instead, we could control the list of available locales programmatically (pulled directly from `GP_Locale`) and changes would be reflected in version control.
  * But, I'm not sure if there are significant query efficiency differences between using terms and using post meta.

ryelle:
> Canonical versions of WordPress core patterns should live in code, in a GitHub repo. A separate repo from WP core, so it can be updated and new patterns added outside of the core release cycle.

The concept of core patterns is going away - once the pattern directory is launched, and after it’s integrated into Gutenberg, the core patterns will be no different than any other patterns. Would this just be temporary, until we have a functional editor/remix flow?

tellyworth:
[...]

PR #229, opened 2021-06-25
Add locale picker to pattern editor

coreymckrill:
This adds a locale field to the Pattern Details panel in the pattern editor (front and back end) The locale data that populates the field comes from the same mu-plugin used elsewhere on dotorg to supply locale data, so this also updates the local dev configuration to contain the required files and database table.

This uses postmeta to store the locale value rather than a taxonomy for the reasons described [here](https://github.com/WordPress/pattern-directory/issues/223#issuecomment-867874228).

See #223, #224

### Screenshots

pattern-details

1. Rebuild your local dev environment with `yarn run create`
2. Create or edit a pattern
3. Check the Pattern Details panel in the editor sidebar. The default language should be English, try choosing something else and saving, then reloading to ensure the value actually gets saved
4. Try using it from the New Pattern page and also from WP Admin


bin/index.sh
```diff
@@ -26,3 +26,6 @@ yarn wp-env run cli wp theme activate pattern-directory
 
 # Change permalinks
 yarn wp-env run cli wp rewrite structure '/%postname%/'
+
+# Import tables
+npm run wp-env run cli wp db import wp-content/uploads/data/wporg_locales.sql
```
ryelle:
```suggestion
yarn wp-env run cli wp db import wp-content/uploads/data/wporg_locales.sql
```

composer.json
```diff
@@ -35,14 +35,25 @@
 						"url": "https://meta.svn.wordpress.org/sites/",
 						"reference": "trunk/wordpress.org/public_html/wp-content/themes/pub/wporg/"
 					}
+				},
+				{
+					"name": "wordpress-meta/pub",
```
ryelle, line 40: I'm getting `wordpress-meta/pub could not be downloaded` when running `composer install`, any ideas? I'm not sandboxed, and my proxy seems fine.

public_html/wp-content/plugins/pattern-directory/src/pattern-post-type/details.js
```diff
@@ -1,33 +1,44 @@
...
 	useEffect( () => {
 		editPost( {
 			meta: {
 				...postMetaData,
 				wpop_description: description,
+				wpop_locale: locale,
 			},
 		} );
-	}, [ description ] );
+	}, [ description, locale ] );
```
ryelle: We removed the `useEffect` call here in #112, so you can set this using `editPost` in the `onChange` directly - you shouldn't need the local state `locale`.

ryelle (commented):
The code looks good, but I get a fatal error when I try to test (since the locale plugin won't download). Any ideas why composer can't find it?

composer.json
coreymckrill: Huh, weird. I just tried deleting the `pub` folder and re-running `composer install` and it's still working for me. You shouldn't even need the proxy anyway since it's meta.svn.

I copied this over from Learn's composer file because we use the locale stuff there too. If you run `composer install` on your Learn checkout do you get the same error?

public_html/wp-content/plugins/pattern-directory/src/pattern-post-type/details.js
coreymckrill: Fixed with 840e9cc

StevenDufresne:
I'm not sure what's happening but the publish modal seems messed up in this branch. I don't see the same problem in trunk.

ryelle:
@StevenDufresne I'm seeing that on and off (even in trunk) -- I think it has to do with the gutenberg version, maybe? Either way, I don't think it's related to this branch.

composer.json
ryelle: yeah, same thing on learn, `wordpress-meta/locale-detection could not be downloaded` - the interesting thing is that it claims that `svn: command not found`, but I definitely have SVN here.

In any case, I did a few manual steps and got it working enough to test.

@StevenDufresne Did you run `yarn run create` on this branch, without issues? If it's just my machine, I can figure it out later, but if you have trouble too, we should figure that out.

public_html/wp-content/plugins/pattern-directory/includes/admin-patterns.php
```diff
@@ -90,6 +92,17 @@ function pattern_list_table_render_custom_columns( $column_name, $post_id ) {
...
 			break;
+
+		case 'language':
+			$locale        = $current_pattern->wpop_locale ?: 'en_US';
+			$locale_labels = get_locales_with_english_names();
+
+			if ( isset( $locale_labels[ $locale ] ) ) {
+				echo esc_html( $locale_labels[ $locale ] );
+			} else {
+				echo '&mdash;';
```
ryelle, line 103: Is this case something we should be concerned about? A locale property exists but isn't found in the registered list?

ryelle (approved):
This is looking good to merge (so I'm approving), but I'd like @StevenDufresne to confirm the composer thing first.

public_html/wp-content/plugins/pattern-directory/includes/admin-patterns.php
coreymckrill: It's unlikely, but it could theoretically happen if our locale data from GlotPress changes.

StevenDufresne:
> This is looking good to merge (so I'm approving), but I'd like @StevenDufresne to confirm the composer thing first.

I have no issues with composer when running `yarn run create` on this branch 👍 .

composer.json
iandunn: [OS X recently removed it's bundled `svn`](https://stackoverflow.com/questions/60869347/mac-command-line-tools-11-4-no-longer-has-svn), so if you also have it installed via homebrew, then it could be that composer is using the wrong path or something.

I ran into something like that w/ git-svn, and had to [muck w/ paths to fix it](https://github.com/Homebrew/homebrew-core/issues/52490#issuecomment-792604853)

Issue #163, opened 2021-06-10
Pattern Directory Discoverability


bph:
After I saw [Ana's tweet](https://twitter.com/Ana_Segota/status/1402714564366684163), I wanted to check out her patterns, and couldn't find just her patterns. It would be not only attractive for pattern authors, if attribution would be visible and link able. 

Use cases: 
- Sharable link `patterns/anarieldesign`
- display on author's wp.org profile next to themes and plugins. 
- Filter by on pattern directory, similar to plugin & themes
- Add author name to cards on single pattern view with link to profile page.

_It seems to be on the road map as the single page says "More from this designer"_
cc @anarieldesign

melchoyce:
Hey @bph! This is incoming. See: https://github.com/WordPress/pattern-directory/pull/152

Issue #116, opened 2021-05-27
Pattern Grid: Blank page when no patterns are found


ryelle:
When searching for patterns and none are found, it shows this sad empty state

<img width="1013" alt="Screen Shot 2021-05-27 at 10 40 29 AM" src="https://user-images.githubusercontent.com/541093/119846253-f86bea00-bed7-11eb-8239-d8ee72af71a3.png">

I'm assuming this one should give an error message, then either point the user to other categories, or show a set list of popular patterns…?

-------

A similar thing happens if someone tries to go "out of bounds", like to `/page/2000`, it also shows empty

<img width="1007" alt="Screen Shot 2021-05-27 at 10 40 54 AM" src="https://user-images.githubusercontent.com/541093/119846316-06216f80-bed8-11eb-91cb-bd3b8138ca61.png">

This should probably be a simple 404 page, or maybe redirect back to the first page of the query?

cc @shaunandrews

shaunandrews:
For the empty search results maybe we can show some featured patterns:

![Empty Search](https://user-images.githubusercontent.com/191598/120842752-1e7a3580-c53b-11eb-9133-10bda514117f.png)

Here's a design for the 404 page:

![404](https://user-images.githubusercontent.com/191598/120842781-2934ca80-c53b-11eb-97d6-e43f1b6e10db.png)

PR #230, opened 2021-06-25
Theme: Display patterns by a single author

### LINKED ISSUE #191: Pattern API: Get patterns by author
Add a parameter to the endpoint for requesting patterns by a specific author(s). For example, if you only want to get patterns by the wordpressdotorg author, or just by your own user.

This could be used in place of the Core keyword to request only "core" (by wordpressdotorg) patterns.

ryelle:
This PR makes the author archives work. For example, view `/author/ryelle` (or any user on your site), and it should display just their patterns. Pagination and all other pattern-grid functionality should work.

This also adds a parameter to the patterns API to accept an `author_name`. By default the only way to get posts by author uses the author's ID, which is harder/impossible to get client-side for users not registered on _this_ site. Instead, we pass in the author's username, and PHP can get the ID, and their patterns.

For example, this is the request:
`GET /wp-json/wp/v2/wporg-pattern?author_name=wordpressdotorg`

Fixes #191, #163.

Note that `/author/[yourname]` is not the same as `/my-patterns` — the author archive is the public view, and My Patterns is for pattern management

⚠️  Not done in this PR (will be followups, unless anyone thinks they're blockers): empty state for a user with no patterns (#116), invalid username (use 404 from #116).

### Screenshots

Author archives, ex `/author/wordpressdotorg`:

wporg-author-page

1. View the home page, click on a username
2. Or, go directly to `yoursite.com/author/[anyone]`
3. The author page should display only patterns by that one author, with pagination if needed

API:

- Test the API directly, by making requests with various `author_name` values. A valid author should return an array, and an invalid username should return an error.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -424,8 +424,12 @@ function filter_patterns_collection_params( $query_params ) {
 	}
 
 	$query_params['author_name'] = array(
-		'description' => __( 'Limit result set to patterns by a single author.', 'wporg-patterns' ),
-		'type'        => 'string',
+		'description'       => __( 'Limit result set to patterns by a single author.', 'wporg-patterns' ),
+		'type'              => 'string',
+		'validate_callback' => function( $value ) {
+			$user = get_user_by( 'slug', $value );
+			return (bool) $user;
+		},
 	);
 
 	return $query_params;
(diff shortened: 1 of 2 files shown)


StevenDufresne (approved):
I can't speak to the PHP approach, but the code works as expected and the `QueryMonitor` code looks good. If we continue adding params we may want to cycle back and come up with something more flexible.

javierarce:
I just wanted to point out something odd that I saw that is also currently happening on the production site. Sorry in advance if this is not the best place to report it, but I thought this PR could be related or maybe even a place to fix this:

If you look at the list of results by `wordpressdotorg` in the image above, there's an avatar from another user:

Screenshot 2021-06-28 at 16 18 23

ryelle:
@javierarce thanks for flagging that - it's not related to this PR, but I was able to fix it and just pushed the fix to production.

PR #232, opened 2021-06-28
Theme: Main Theme Header + Search form

### LINKED ISSUE #160: Update the theme header
The latest design has removed the callout (which was removed in https://github.com/WordPress/pattern-directory/commit/e69788eb194e390ecb98b78e00861028a83cf892), and added some pattern images, a search bar, and a link to create patterns.

Header

Given the feedback about the callout, let's hold off on adding the "create patterns" link until that's ready for submissions.

The interior pages also have a slightly changed header - the search should move up into the blue bar. There's another issue for the dropdown, that can be omitted here.

Screen Shot 2021-06-10 at 3 55 39 PM

StevenDufresne:
This PR updates the main theme header to match the designs in #160. In order to add the search form into the header (and the subnav) and reuse the current functionality, this PR moves some of the static HTML into react components which creates more reusable code, smoother interactions, and an overall better experience.

The alternative to this approach would be to make the search bars disconnected from react and always post back to the page. I played with that approach but found this felt better.

There are a few important shortcomings in this PR:

- The single pattern page is still considered a separate page and therefore not a seamless transition. 
- Search form only activates a search on enter, as opposed to delayed updates on KeyPress which was how it used to work.
- The primary menu is hardcoded (this can be scripted later).

Fixes #160
 | Main | With Search Term | Single |
| --- | --- | --- |  
|  |  |  |
| | This view is also the one loaded when users go directry to a category link. | |

1. Visit `/`
2. Search for something in the search form
3. Reload the page
4. Click on a pattern
5. Search in the search form from the single page.


public_html/wp-content/themes/pattern-directory/header.php
```diff
@@ -24,60 +24,3 @@
...
-					</nav><!-- #site-navigation -->
-				<?php endif; ?>
-			</div><!-- .site-branding -->
-		</header><!-- #masthead -->
-
-		<?php if ( ! is_singular() ) : ?>
-		<div class="notice notice-warning notice-alt" style="max-width:960px; margin:1rem auto;">
-			<p style="font-size:0.9375rem; margin:0.75rem 0;">
-			<?php
-				esc_html_e( 'You’re a bit early to the party! This directory hasn’t yet fully launched.', 'wporg-patterns' );
```
ryelle, line 78: I had this in the PHP on purpose, so that if something happens to break the JS loading again, it would still appear.

ryelle (commented):
We actually get a lot from the default theme headers, which would need to be re-implemented in JS if we go this route, for example, [author archives](https://wordpress.org/patterns/author/wordpressdotorg/) are labeled:
Screen Shot 2021-06-29 at 12 18 11 PM

It also looks like the My Patterns page is missing the header.

This link goes to a page that doesn't exist:
Screen Shot 2021-06-28 at 12 26 33 PM

What was the drawback to leaving the header in PHP/CSS? Since we're using browser navigation for pages & search, I don't see a reason to make the header & nav JS components.

StevenDufresne:
> We actually get a lot from the default theme headers, which would need to be re-implemented in JS if we go this route, for example, [author archives](https://wordpress.org/patterns/author/wordpressdotorg/) are labeled:
> Screen Shot 2021-06-29 at 12 18 11 PM
> 
> It also looks like the My Patterns page is missing the header.
> 
> This link goes to a page that doesn't exist:
> Screen Shot 2021-06-28 at 12 26 33 PM
> 
> What was the drawback to leaving the header in PHP/CSS? Since we're using browser navigation for pages & search, I don't see a reason to make the header & nav JS components.

The main drawback was that the search bar would have to post to the server and reload the page for it to work. We would lose that SPA feeling when searching for patterns. Additionally, the breadcrumbs in the sub-nav don't match the designs and we would need to rebuild them in php anyways. 

I did get the alternate approach pretty close here as well, so moving in that direction is not going to slow us down much. Take a look:
https://github.com/WordPress/pattern-directory/compare/update/main-page-header-post?expand=1

public_html/wp-content/themes/pattern-directory/header.php
StevenDufresne: I understand. It won't be in the right place if we do go with this approach and keep it in php.

ryelle:
> The main drawback was that the search bar would have to post to the server and reload the page for it to work. 

You've already got it so that the user needs to hit enter to search (which is a good thing), I don't think we loose anything by making the page refresh. I'm also wondering if it still makes sense for search to search within a category now that it's not in the category filter bar, it might seem unexpected that it's not a site-wide search.

> Additionally, the breadcrumbs in the sub-nav don't match the designs and we would need to rebuild them in php anyways.

They're pretty much there already, the main difference I see is the "Categories" link in the design, but that page doesn't exist, so it would be removed anyway.

I like the alternate branch you've linked here, `update/main-page-header-post`, I think that has more promise, especially for a shorter launch timeline. Maybe once the initial launch milestone is cleared out, we can loop back for interactive improvements.

I did notice, on that branch, searching from the single pattern page doesn't work - it loads `/pattern/pattern-name/?search=buttons`. I can give it a proper review once you've written up a PR 👍🏻

StevenDufresne:
Closing in favour of #269. We'll reopen if the other PR hits a real blocker.

PR #236, opened 2021-06-28
Theme: Add a resolver for the `getPattern` selector

ryelle:
This PR adds a resolver for the `getPattern` selector, so it can be used outside of query requests. This is useful for the single pattern preview - rather than passing in various pattern meta, we can request the pattern itself via the API. Between #232 and an in-progress PR I have for the "more patterns" section, the `pattern-preview__container` element is getting overloaded.

Now, we can get the pattern data with just a single call:

```js
const pattern = useSelect( ( select ) => select( patternStore ).getPattern( postId ), [ postId ] );
```

And we can use `pattern.title.rendered` or `pattern.author_meta`, etc.

### Screenshots

No visual changes

1. No UI changes, everything should work the same
2. Tests should pass

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/src/components/pattern/index.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/pattern/index.js
@@ -1,17 +1,28 @@
 /**
+ * WordPress dependencies
+ */
+import { useSelect } from '@wordpress/data';
+
+/**
  * Internal dependencies
  */
 import PatternPreview from '../pattern-preview';
 import PatternPreviewActions from '../pattern-preview-actions';
 import ReportPatternButton from '../report-pattern-button';
+import { store as patternStore } from '../../store';
 
-const Pattern = ( { content, postId, userHasReported, loggedIn } ) => {
+const Pattern = ( { postId, userHasReported, loggedIn } ) => {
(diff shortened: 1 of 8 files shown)


public_html/wp-content/themes/pattern-directory/src/components/pattern/index.js
```diff
@@ -1,17 +1,28 @@
...
+	const pattern = useSelect( ( select ) => select( patternStore ).getPattern( postId ), [ postId ] );
+	if ( ! pattern ) {
+		return null;
+	}
+
 	return (
 		<>
 			<PatternPreviewActions postId={ postId } />
-			<PatternPreview blockContent={ content } />
+			<PatternPreview blockContent={ pattern.content.rendered } />
```
StevenDufresne, line 25: Is it possible that `rendered` is not present?

StevenDufresne (approved):
This looks straightforward. Go for it!

public_html/wp-content/themes/pattern-directory/src/components/pattern/index.js
ryelle: No, if the pattern exists, then it was a successful API request which has `content.rendered`.

PR #238, opened 2021-06-28
Theme: Display other patterns by this author on single page

### LINKED ISSUE #234: Hide "More from this designer" in case there's just one pattern
If a user has just one pattern, we are currently showing the "More from this designer" section. 

In the future, I imagine that we could show related patterns using the tags or the description of the current item, but for the time being I think we should hide this section completely.

Screenshot 2021-06-28 at 15 38 31

ryelle:
This adds the "More from this designer" section to single pattern pages. This displays up to 3 other patterns that are also made by the current pattern's author.

Fixes #234, see #230.

Note, this is based on #236 since it uses the `getPattern` selector.

### Screenshots

User with a lot of patterns, shows the 3 most recent (excluding the current pattern).

three

User with just two published patterns, shows only 1 in "More…"

one-pattern

User with only 1 pattern, shows an empty section.
no-patterns

1. Build the branch, `yarn workspaces run build`
2. View a single pattern by an author with multiple published patterns
3. Under the pattern preview, there should be a section "More from this designer"
4. View a single pattern by an author with only one published pattern
5. There's nothing displayed under the pattern preview.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
+++ b/public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
@@ -38,16 +38,12 @@ $raw_block_content = get_the_content();
 
 				<div
 					hidden
-					class="pattern-preview__container"
+					class="pattern__container"
 					data-post-id="<?php echo intval( get_the_ID() ); ?>"
 					data-logged-in="<?php echo json_encode( is_user_logged_in() ); ?>"
 					data-user-has-reported="<?php echo json_encode( $user_has_reported ); ?>"
-				></div>
+				></div><!-- .pattern__container -->
 
-				<div class="entry-content">
-					<h2><?php esc_html_e( 'More from this designer', 'wporg-patterns' ); ?></h2>
-					<div class="pattern-grid"></div>
-				</div><!-- .entry-content -->
 			</article><!-- #post-## -->
 
 		<?php endwhile; ?>
--- a/public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/index.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/index.js
@@ -10,7 +10,7 @@ import { useSelect } from '@wordpress/data';
 import Pagination from './pagination';
 import { store as patternStore } from '../../store';
 
-function PatternGrid( { before, children, query, showPagination = true } ) {
+function PatternGrid( { header, children, query, showPagination = true } ) {
 	const { isLoading, posts, totalPages } = useSelect( ( select ) => {
 		const { getPatternTotalPagesByQuery, getPatternsByQuery, isLoadingPatternsByQuery } = select(
 			patternStore
@@ -25,7 +25,7 @@ function PatternGrid( { before, children, query, showPagination = true } ) {
 
 	return (
 		<>
-			{ posts.length ? before : null }
+			{ posts.length ? header : null }
 			<div className="pattern-grid">{ isLoading ? <Spinner /> : posts.map( children ) }</div>
 			{ showPagination && <Pagination totalPages={ totalPages } currentPage={ query?.page } /> }
 		</>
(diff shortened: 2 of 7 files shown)


public_html/wp-content/themes/pattern-directory/single-wporg-pattern.php
```diff
@@ -38,16 +38,12 @@
 
 				<div
 					hidden
-					class="pattern-preview__container"
+					class="pattern__container"
```
StevenDufresne, line 41: I coincidentally made the same change in:
https://github.com/WordPress/pattern-directory/pull/232/files#diff-a8a05ff78c3ae805334de62d6ed0efeb453b125ee358026d5dc43247ec90b13aR37

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/index.js
```diff
@@ -25,8 +25,9 @@ function PatternGrid( { query, children } ) {
 
 	return (
 		<>
+			{ posts.length ? before : null }
```
StevenDufresne: I'm not opposed to this but it does 'lock-in' functionality. If we wanted to show a section with a before and a "no items" view we couldn't (although that would also require a small refactor). Have you considered any other approaches?

public_html/wp-content/themes/pattern-directory/src/components/pattern/author-details.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * WordPress dependencies
+ */
```
StevenDufresne: Is this necessary?

StevenDufresne (commented):
I'm wondering if the author information should be in the pattern header as opposed to down here especially since we added it to the card preview?

public_html/wp-content/themes/pattern-directory/src/components/pattern/author-details.js
ryelle: nope

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/index.js
ryelle: > If we wanted to show a section with a before and a "no items" view we couldn't

I think that would be a different prop - like, `beforeNone` or something. If we wanted to show something before any content, we could just put it in the parent, like we do in Patterns & MyPatterns, like so:

```
<Menu />
<PatternGrid query={ modifiedQuery }>
	{ ( post ) => <PatternThumbnail key={ post.id } pattern={ post } /> }
</PatternGrid>
```

The point of `before` is to only show when there is content.

public_html/wp-content/themes/pattern-directory/src/components/pattern/author-details.js
ryelle: nope

ryelle:
> I'm wondering if the author information should be in the pattern header as opposed to down here especially since we added it to the card preview?

I'd rather not change the design here - but feel free to make an issue to bring up with @shaunandrews when he's back from AFK.

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/index.js
StevenDufresne: I typically find components that have a very small number of props easier to modify and make components easier to understand. I can see how a `before` pattern leads to us then having an `after` prop, and so on.... and then we're stuck toggling on/off UI views in a component that should just be a grid of patterns.

ryelle: Would it change anything if the prop was called `header` instead of `before`? I was aiming to keep the `Pattern` component separate from `PatternGrid`, and not fetch the by-author query in both places. I didn't want to overcomplicate `Pattern` by introducing `getPatternsByQuery` & `isLoadingPatternsByQuery`, just to do something that's easily done in `PatternGrid`.

StevenDufresne: Right, I understand the compromise. I think `header` is better than before. 👍

StevenDufresne (approved):

PR #240, opened 2021-06-29
Add/pattern grid loading

StevenDufresne:
This PR adds a skeleton card to display while we are loading patterns. The implementation differs slightly from the designs seeing that meta-information has drifted.

This PR doesn't address 2 outstanding issues:

1. All the `iframe`s load at once and are messy when rendering. We should do something about that in the following PR
2. The context bar loading seems to be fairly buggy. Its height is inaccurate. 

Designs: #101 

### Screenshots

Check out the [animated GIF.](https://d.pr/i/PQ453A)

1. Visit `/`
2. Expect to see a loading state
3. Click on a category, expect to see the loading state


public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/card-skeleton.js
```diff
@@ -0,0 +1,45 @@
...
+		}
+	}, [ wrapper ] );
+
+	return (
+		<div ref={ wrapper }>
+			<SkeletonWrapper>
+				<Skeleton height={ height } />
+				<Skeleton height="14px" marginTop="1rem" />
+
+				<SkeletonWrapper styles={ { flexDirection: 'row', marginTop: '0.5rem', alignItems: 'center' } }>
```
ryelle: Why `styles` instead of `style`?

public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/canvas.js
```diff
@@ -11,6 +11,16 @@ import Iframe from '../iframe';
...
+/**
+ * Returns the height of the preview window.
+ *
+ * @param {HTMLElement} element Html element
+ */
+
+export const getFrameHeight = ( element ) => {
+	return element.clientWidth * ASPECT_RATIO;
+};
+
```
ryelle: What do you think about moving this & `ASPECT_RATIO` to `src/utils`?

ryelle (commented):
Working great! I had two small comments about the code.

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/card-skeleton.js
StevenDufresne: I was thinking that since it was going to be passed on specifically, not like `<Component {...props} />`, that it maybe should be different, but thinking about it out loud, probably not necessary.

public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/canvas.js
StevenDufresne: That could work... the calculation will still need to match though, making the assumption that `clientWidth` would always be used presents potential debt.

public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/card-skeleton.js
StevenDufresne: Updated in a8e2ce7.

public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/canvas.js
ryelle: Yeah, I meant moving it to the utils folder and importing that function here & in card-skeleton -- in fact, if you made `getFrameHeight` take a width instead of an HTMLElement, you can use it [here too](https://github.com/WordPress/pattern-directory/blob/ff09c1afe4fad91686b9d41b508b5fc4cdfa7c13/public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/canvas.js#L41), and then `ASPECT_RATIO` could stay "internal" to that utils folder.

StevenDufresne: Oh right, yeah totally, will do.

ryelle (approved):

PR #249, opened 2021-06-30
Pattern Creator: Set up creating/editing permissions for logged in wp.org users

### LINKED ISSUE #30: Pattern CPT permissions for creation & editing
Right now, the pattern CPT endpoint works as long as you're a user on the site, but we're not going to add every wp.org user to the patterns site. We need to update the permissions checks to allow any logged in wp.org user to edit/create a pattern. Any custom validations could also be added in this step, maybe? See #22.

Probably this would be done by subclassing `WP_REST_Posts_Controller`, adding that as the `rest_controller_class` for the CPT, and overriding `create_item_permissions_check` & `update_item_permissions_check`.

coreymckrill:
Enables logged in wp.org users to create and edit patterns on the front end of the Pattern Directory without having to be an actual member of the patterns site.

First I approached this in the way suggested in #30, modifying the permissions checks in the rest controller for the patterns post type. However, it became apparent that the capability checks that were preventing non-site members from creating patterns were used in other places besides the rest controller, so I decided to switch gears and grant the necessary capabilities in certain limited circumstances. Thus the rest controller is not modified at all here.

This also fixes an issue with the pattern theme's stylesheet that was preventing it from loading correctly when trying to access the New Pattern screen while logged out.

Fixes #30

1. For testing in the local dev environment, create a new user with the Subscriber role and log into the site as that user. This simulates being a wp.org user who is not a member of the site, but who is logged in.
2. Go to the [New Pattern screen](http://localhost:8888/new-pattern/). The editor should load with no errors in the console or broken XHR requests.
3. Create and publish a new pattern. This should succeed and there should be no console errors.
4. Try going to [WP Admin](http://localhost:8888/wp-admin/) and see if you have access to the Block Pattern menu or any of the screens within it. You shouldn't.
5. Try to find scenarios where the Subscriber user has access that it shouldn't.
5. Now log out and log back in as an admin. You should have full access to block patterns in WP Admin and pattern creation on the front end.

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/view/editor.php
+++ b/public_html/wp-content/plugins/pattern-creator/view/editor.php
@@ -8,8 +8,6 @@ use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\POST_TYPE;
 
 get_header();
 
-// @todo Permissions TBD, see https://github.com/WordPress/pattern-directory/issues/30.
-
 $is_logged_in = is_user_logged_in();
 $can_edit     = current_user_can( 'edit_post', get_query_var( PATTERN_ID_VAR ) );
 
--- a/public_html/wp-content/plugins/pattern-creator/view/editor.php
+++ b/public_html/wp-content/plugins/pattern-creator/view/editor.php
@@ -11,9 +11,15 @@ get_header();
 // @todo Permissions TBD, see https://github.com/WordPress/pattern-directory/issues/30.
 
 $is_logged_in = is_user_logged_in();
-$can_edit     = current_user_can( 'edit_post', get_the_ID() );
+$can_edit     = current_user_can( 'edit_post', get_query_var( PATTERN_ID_VAR ) );
 
-if ( is_singular( POST_TYPE ) ) {
+$current_page_query_args = array( 'pagename' => 'new-pattern' );
+if ( get_query_var( PATTERN_ID_VAR ) ) {
+	$current_page_query_args[ PATTERN_ID_VAR ] = get_query_var( PATTERN_ID_VAR );
+}
+$current_page_url = add_query_arg( $current_page_query_args, home_url() );
+
+if ( is_editing_pattern() ) {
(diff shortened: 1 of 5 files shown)


ryelle:
This is great! I can't find any issues when using a subscriber role, I also tested a few places with admin and contributor.

I noticed is that if the user tries to edit a pattern they don't own (`/pattern/[ID]/edit/`), it doesn't load anything:

Screen Shot 2021-07-01 at 16 50 57

And on the logged out screen, the header styles are a little off (same in the above screenshot too). If this isn't an easy fix, it's fine to leave for another PR.

Screen Shot 2021-07-01 at 16 54 07

coreymckrill:
> I noticed is that if the user tries to edit a pattern they don't own (/pattern/[ID]/edit/), it doesn't load anything

Interesting. I'm guessing this is because WP doesn't consider the `/pattern/[ID]/edit/` URI to be a `singular` view of that pattern. We might have to come up with a different conditional to use in the editor template...

coreymckrill:
Ok, this should be fixed with 2d36ceb

coreymckrill:
> And on the logged out screen, the header styles are a little off (same in the above screenshot too).

I'm not seeing specifically how they are off. But I didn't actually change anything in the stylesheet here, just how it was enqueued, so probably better to save that for a separate PR.

public_html/wp-content/plugins/pattern-creator/view/editor.php
```diff
@@ -10,35 +10,50 @@
 
 // @todo Permissions TBD, see https://github.com/WordPress/pattern-directory/issues/30.
```
ryelle: Can you remove this todo comment before merging?

ryelle (approved):
All working correctly now 👍🏻

public_html/wp-content/plugins/pattern-creator/view/editor.php
coreymckrill: Done in 7099169

PR #250, opened 2021-07-01
Add Translation of Patterns via GlotPress

dd32:
This PR ports work from WordPress.com for Pattern Translation to the Pattern Directory.

There are several working parts to this that work together:
 - Patterns are parsed, and static text is extracted from it
 - Those strings are pushed into GlotPress
 - Those strings are pulled from GlotPress and new Patterns are created in a translated state automatically.

Included are some CLI commands used on WordPress.com, which might make little sense on W.org but I've been using to test & debug.

This is still a work in progress, and will need to be squashed on merge.

Testing this inside the Local Development environment is likely to not be viable, due to the specific integrations with translate.wordpress.org

See #223 

Props @lsl, @dd32, @andrewserong, @deBhal, @enejb, @ramonjd, @akirk, @yuliyan

### Screenshots

An example of an "automatically translated" pattern:
Screen Shot 2021-07-01 at 10 57 03 am

1. Use the CLI commands to test everything works as expected
2. Run the cron tasks on WordPress.org
3. Verify posts created as expected


lsl:
This is awesome, thanks for giving this some new life, good to see this finding some use outside of .com!

@otherwpcomdevsnotyetlisted = @andrewserong, @deBhal, @enejb, @ramonjd, @akirk, @yuliyan

Btw there are a decent amount of unittests in bin/tests/isolated available for this code if you need or think it has value pulling over. (might not be worth it at this point!)

dd32:
> @otherwpcomdevsnotyetlisted

Thank you @lsl! Updated the PR with that.

> Btw there are a decent amount of unittests in bin/tests/isolated available for this code if you need or think it has value pulling over. (might not be worth it at this point!)

Any tests related to the Parsers would be of great benefit, I think I've only applied bugfixes (and renamed the classes to be more namespacey) but validating if I broke anything in doing so would be good.

Given the manual review of the patterns that I've been doing, I'm not going to priortise porting those tests today though.

FYI folks This PR is probably mergeable (Other than lack of tests, and utterly failing php lint checks), the code **is running** on my wordpress.org sandbox, and the cron jobs **are being processed on production via it** - so that's a good enough reason to merge and deploy it I think.

ryelle:
I've fixed some of the php linter issues in https://github.com/WordPress/pattern-directory/pull/250/commits/34bb13171eee2fbe19c1321c16232ee03c7bd268, and in combination with the changes in #264 this should pass the lint checks.

ryelle (approved):
Merging #264 & rebasing fixed up the lint issues, and everything looks good here. @dd32 since you've already got this on your sandbox, do you want to do the merge & sync honors? I can grab it if not 🙂

PR #253, opened 2021-07-01
CPT: Update the filter name for block types to fix warning.

### LINKED ISSUE #246: Deprecated notice for `allowed_block_types` breaks editor in local envs with WP 5.8
`wp-admin/post-new.php?post_type=wporg-pattern` is broken in local environments running WP 5.8-RC1 when `WP_DEBUG_DISPLAY` is `true`:

> PHP Deprecated:  `allowed_block_types` is <strong>deprecated</strong> since version 5.8.0! Use `allowed_block_types_all` instead.
> PHP Warning:  Cannot modify header information - headers already sent...

Background on the the deprecation: 
https://make.wordpress.org/core/2021/06/16/block-editor-api-changes-to-support-multiple-admin-screens-in-wp-5-8/

Uses:

https://github.com/WordPress/pattern-directory/blob/ff09c1afe4fad91686b9d41b508b5fc4cdfa7c13/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php#L14

https://github.com/WordPress/pattern-directory/blob/89bf68a873a93adcd4f50e8b7b653843b890a8c3/public_html/wp-content/plugins/pattern-creator/pattern-creator.php#L109

ryelle:
The filter name has changed from `allowed_block_types` to `allowed_block_types_all`, and the second parameter is now the block context, which includes the current post (if it's a post editor). This update changes the filter we're using to account for that.

Fixes #246

### Screenshots

No visual change.

1. Have `WP_DEBUG_DISPLAY` set to true on a local env
2. Try to add a pattern in wp-admin
3. It should work fine, no deprecation notice in the header

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
+++ b/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
@@ -12,6 +12,7 @@
 
 namespace WordPressdotorg\Pattern_Creator;
 use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\POST_TYPE;
+use WP_Block_Editor_Context;
 
 const AUTOSAVE_INTERVAL = 30;
 const IS_EDIT_VAR = 'edit-pattern';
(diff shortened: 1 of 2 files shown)


coreymckrill (commented):
👍 Looks good, fixes the error in the back end block editor. It appears that we're still using the deprecated hook [here](https://github.com/WordPress/pattern-directory/blob/89bf68a873a93adcd4f50e8b7b653843b890a8c3/public_html/wp-content/plugins/pattern-creator/pattern-creator.php#L109).

ryelle:
Ah, good catch. Updated it there too.

coreymckrill (approved):

Issue #188, opened 2021-06-14
Improve loading state to avoid "page jumping"


shaunandrews:
When browsing through categories the page "jumps" around as patterns are loaded. Here's a quick video to show the behavior:

https://user-images.githubusercontent.com/191598/121961221-21dca080-cd35-11eb-8e22-8394f6a70456.mp4

Perhaps we can retain the height of the page with some skeleton-style placeholders while loading:

<img width="970" alt="image" src="https://user-images.githubusercontent.com/191598/121961379-5cded400-cd35-11eb-9bbf-c20427486b17.png">

ryelle:
This is a duplicate of #101 — we haven't implemented the loading state at all yet.

shaunandrews:
Here I am trying to be helpful and just creating more noise. Sorry about that.

PR #254, opened 2021-07-02
Theme: Update Pattern Grid menu loading

### LINKED ISSUE #101: Pattern Grid: Add a loading state

StevenDufresne:
This PR is the last in a series to add loading to the pattern grid and fixes the "jumping" nature of the page. This PR updates the pattern category menu to use the newly introduced loading skeleton and explicitly sets the height to avoid layout shifts on larger displays.

Fixes #101, Related #188

### Screenshots
| Before | After | 
| --- | --- |
|  |     |

1. Visit `/`
2. Observe the menu loading
3. Click on a category
4. Reload the page
5. Expect to see the same experience as seen in the second step.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern-grid-menu.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern-grid-menu.scss
@@ -15,10 +15,19 @@
 		width: calc(100% - #{$gutter-default * 2});
 	}
 
+	.pattern-skeleton__container {
+		flex-wrap: wrap;
+		height: 2.3125rem;
+		align-items: center;
+
+		span {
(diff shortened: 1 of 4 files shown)


ryelle:
I pushed a change to remove the fixed height from the nav menu, and allow for wrapping of the skeleton menu items. Now the height is only set on the skeleton, and uses `rem` to make sure it always matches the nav menu size (even if the user changes the base font size with text zoom). The wrapping is necessary if you increase the font size beyond 120%.

ryelle (approved):
If you think my changes are okay, this is good to merge 👍🏻

PR #264, opened 2021-07-02
Tools: Update PHP CodeSniffer config

ryelle:
See #250 -- There are a handful of failures and places where the phpcs rules are disabled. My goal was to remove `phpcs:disable`/`phpcs:enable` as much as possible.

1. Allow the short array syntax.
2. Update the list of file names allowed to deviate from the hyphenated format - previously this was leftover from Learn, now it includes the parser files in #250.
3. Add `nodeValue` and `parentNode` as valid camelCase properties, also for #250 since these values are the result of `DOMXPath`.
4. Ignore all docs issues in the `pattern-translations` includes.

Those last two I feel the least sure about. For 3, it seems pretty innocuous but also specific to that one plugin. For 4, I feel like the better solution would obviously be to document all these new functions and methods, but they're fairly well named from what I've seen, and I don't want documentation to hold up getting translations merged. The docs rules are still in place for all other PHP files.

`composer run lint` should pass, any syntax/style errors you make should be flagged as expected.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -87,16 +87,25 @@
 		<!-- print_r() is perfectly accepted in some circumstances, like WP_CLI commands. -->
 		<exclude name="WordPress.PHP.DevelopmentFunctions.error_log_print_r" />
 
-		<!-- Allow short ternary pattern -->
+		<!-- Allow short ternary pattern. -->
 		<exclude name="WordPress.PHP.DisallowShortTernary.Found" />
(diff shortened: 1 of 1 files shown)


iandunn (approved):
I personally prefer the long array syntax, for the reasons Core uses it:

https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/

I'm ok if y'all prefer the short one, though.

PR #265, opened 2021-07-05
Replace backtick quotes

### LINKED ISSUE #260: Replace backtick quotes
**Describe the bug**
This line uses backtick instead of single quotes:
https://github.com/WordPress/pattern-directory/blob/841f5a9beb1106bd23a76a6cd0e313ccd07c2f56/public_html/wp-content/plugins/pattern-creator/src/plugins/viewport-header-control.js#L117

This (I can't think of another reason why) makes GlotPress to fail importing the string - instead it imports the textdomain:
https://translate.wordpress.org/projects/meta/pattern-directory/ro/default/?filters%5Boriginal_id%5D=12101839

**Expected behavior**
See `Width (%spx)` string in GP, not `wporg-patterns`

vlad-timotei:
Fixes #260
See [this issue as well](https://github.com/wp-cli/i18n-command/issues/257)

This replaces backquotes with single quotes to allow current version of wp-cli i18n command to correctly extract the string. Also  following the standards of not using template literals in this case.

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/src/plugins/viewport-header-control.js
+++ b/public_html/wp-content/plugins/pattern-creator/src/plugins/viewport-header-control.js
@@ -114,7 +114,7 @@ const ViewportHeaderControl = () => {
 				icon={ null }
 				text={ sprintf(
 					/* translators: %s viewport width as css, ie: 100% */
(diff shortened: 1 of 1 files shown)


ryelle (approved):
Thanks for finding & fixing this!

Issue #255, opened 2021-07-02
Improve Pattern-Thumbnail Rendering/Loading


StevenDufresne:
We use iframes to render previews for patterns in the pattern grid.  

Our current implementation has 2 drawbacks:

- Assets are loaded for every iframe (although they are all the same)
- We display them immediately resulting in numerous layout shifts which freezes the browser

<hr/>

## Screenshots

![Loading](https://d.pr/i/5HdgnR.gif)
![](https://d.pr/i/s4V6i9.png)

## Assets are loaded multiple times
We inject the editor styles into every `iframe` which forces the browser to redownload them for each preview:

**Code**: 
https://github.com/WordPress/pattern-directory/blob/841f5a9beb1106bd23a76a6cd0e313ccd07c2f56/public_html/wp-content/themes/pattern-directory/src/components/pattern-preview/canvas.js#L19

####  Ideas
Could we maybe grab all the styles and inject them directly? 

<hr/>

## We display them immediately resulting in numerous layout shifts which freezes the browser

This is partly worsened by the asset problem mentioned above because there is a lot of requests for assets that come in at different times and cause each `iframe` to update their view at different times.

####  Ideas

1. Only load the iframe if it's visible. (Adding `loading=lazy` will not work in this case)
2. Obfuscate the view until all the assets have loaded in the iframe.
   - I tested this by just adding an overlay and removing it after 2 seconds and it helps. We obviously wouldn't want an arbitrary timeout controlling this.

ryelle:
> Obfuscate the view until all the assets have loaded in the iframe. 

Could we do this by listening to the iframe's `onload` event? I _think_ this would wait for all assets to load before being fired. Maybe we could use `filter: blur(…)` on the thumbnail, to have each pattern still seem unique until the iframe assets load.

> Could we maybe grab all the styles and inject them directly?

One idea I had for this issue https://github.com/WordPress/pattern-directory/issues/143 is the idea of creating our own mini-theme base for displaying patterns - a set of styles that use the bare minimum needed to render nicely, and get injected directly into the preview without the request off to the themes site. But that's a bigger task :)

Does the browser not cache the CSS file after it's been loaded once? it _shouldn't_ be doing requests for each iframe, but I'm not sure if the nature of iframes change how browsers cache files.

StevenDufresne:
> > Obfuscate the view until all the assets have loaded in the iframe.
> 
[...]

Issue #235, opened 2021-06-28
Buttons & social links often show the wrong text/icon color


kjellr:
Buttons & social links often show the wrong text/icon color in the pattern preview. 

## Examples: 

https://wordpress.org/patterns/pattern/introduction-with-image-on-the-left/
https://wordpress.org/patterns/pattern/contact-and-social-links/

Editor|Frontend Preview
---|---
<img width="1105" alt="Screen Shot 2021-06-21 at 8 17 17 AM" src="https://user-images.githubusercontent.com/1202812/122760514-3525c980-d269-11eb-8e4d-bd3272a7ba7a.png">|<img width="1012" alt="Screen Shot 2021-06-21 at 8 17 07 AM" src="https://user-images.githubusercontent.com/1202812/122760518-3656f680-d269-11eb-9c68-7e331a341bc6.png">

Editor|Frontend Preview
---|---
<img width="1105" alt="Screen Shot 2021-06-21 at 8 17 35 AM" src="https://user-images.githubusercontent.com/1202812/122760532-3bb44100-d269-11eb-8a6d-dde5a4bed497.png">|<img width="1020" alt="Screen Shot 2021-06-21 at 8 16 11 AM" src="https://user-images.githubusercontent.com/1202812/122760544-3e169b00-d269-11eb-9129-7f441b46f8a2.png">

coreymckrill:
From https://github.com/WordPress/pattern-directory/pull/266#issuecomment-874811398

> I'm pretty sure that's a gutenberg bug with custom link colors, but haven't tracked it down yet.

@ryelle is [this](https://github.com/WordPress/gutenberg/issues/31990) the issue you were thinking of?

ryelle:
I don't think that's exactly it, I'll see if I can replicate outside of the preview iframe. The issue I see here is this `important` style is overriding the links in the icon, even though the icons have colors set locally. The output HTML looks like this:

```html
<li style="color: #FFFFFF; background-color: #000000; " class="wp-social-link wp-social-link-twitter wp-block-social-link">
  <a href="https://twitter.com" aria-label="Twitter: https://twitter.com" class="wp-block-social-link-anchor"> <...svg...></a>
</li>
...
<style>.wp-elements-60ef5bd1316b8 a{color: #000000 !important;}</style>
```

ryelle:
[...]

PR #266, opened 2021-07-05
Pattern Previews: Use less-opinionated base theme

### LINKED ISSUE #143: Tidy up display of patterns in the front end.
I'm not sure what the long-term goal is for the "theme" used in the pattern previews, but for now I'd recommend we make a couple small changes to improve the appearance of all patterns: 

- Use a system sans-serif font stack throughout. (`-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;`)
- Change the background to white so it's more neutral. 

Before|After
---|---
Screen Shot 2021-06-09 at 8 33 36 AM|Screen Shot 2021-06-09 at 8 34 35 AM

ryelle:
Fixes #143 - Switch the pattern preview theme styles to use Twenty Twenty-One with a white background as the base CSS. This also removes some other CSS dependencies that are not necessary in this context (non-editable). This brings the number of files loaded per each iframe down to two - `wp-block-library` and the theme styles (see #255).

### Screenshots

patterns
pattern-2
pattern-1

1. Check out branch and build with `yarn workspaces run build`, or just look at the screenshots ^
2. View the pattern previews in grid mode
3. They should look as expected - headings are plain sans-serif, no top border. Body text is also sans-serif. Background should be white (unless the pattern uses a background).
4. View single pattern previews
5. They should look as expected, see above

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -90,11 +90,7 @@ function enqueue_assets() {
  * See https://github.com/WordPress/gutenberg/blob/6ad2a433769a4514fc52083e97aa47a0bc9edf07/lib/client-assets.php#L710
  */
 function generate_block_editor_styles_html() {
-	$handles = array(
-		'wp-block-editor',
-		'wp-block-library',
-		'wp-edit-blocks',
-	);
+	$handles = array( 'wp-block-library' );
 
 	$block_registry = \WP_Block_Type_Registry::get_instance();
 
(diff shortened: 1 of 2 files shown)


StevenDufresne (commented):
Speed looks better on load! Good call.

Are we going to assume that some CSS styles will not work or conflict?

The button is `background-color: #fff` on `color: #fff`:

`/pattern/introduction-with-image-on-the-left/`

StevenDufresne (approved):

ryelle:
> Are we going to assume that some CSS styles will not work or conflict?

I'm pretty sure that's a gutenberg bug with custom link colors, but haven't tracked it down yet. It's an issue with the current (Twenty Nineteen) display too, so it's unrelated. It was reported here: #235

kjellr:
Thank you for this change! It looks great.

PR #267, opened 2021-07-05
Theme: Add RTL build step, fix issues in RTL sites

### LINKED ISSUE #257: Add RTL Styling
**Describe the bug**
https://ar.wordpress.org/patterns/ has no CSS Styles.

**To Reproduce**
1. Go to https://ar.wordpress.org/patterns/ or set your locale to an RTL language.

**Expected behavior**
A beautiful website should appear.

ryelle:
Fixes #257 — this ports over the `rtlcss` task from [the `wporg` parent theme](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg/Gruntfile.js#L89), but switches from the `grunt-rtlcss` wrapper to using `rtlcss` in postcss. The version of `rtlcss` in `grunt-rtlcss` is outdated (2.x), while using `rtlcss` as a postcss processor means we can stick with `rtlcss` 3.x (and PostCSS 8.x).

This also ports over the dashicons arrow-swapping code, but it isn't necessary for the site right now - I only added it for consistency with the other sites.

In addition to generating the RTL CSS, I fixed some display issues in RTL

- The iframe alignment needs to use `top right` when the site is RTL
- The drag handles should not be swapped, otherwise the drag direction is reversed

We could use a different image for RTL, since the X lines up well on LTR, but not here. I don't think we have a good way to _translate_ this image, but at least swapping LTR/RTL would be nice.

rtl-ar-modal-img

This left-aligned modal title seems to be a gutenberg bug.

rtl-ar-modal-x

### Screenshots

rtl-ar-single
rtl-ar-grid

1. Check out branch and build with `yarn && yarn workspaces run build` (note the `yarn`, need to install new deps)
2. Switch your site to an RTL language, ex Arabic or Hebrew
3. Navigate through the frontend, it should have style, and nothing should look broken

Note, this doesn't add RTL support to the pattern creator, so that's still a bit janky - can be done post-launch, when we restart work on the creator.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/Gruntfile.js
+++ b/public_html/wp-content/themes/pattern-directory/Gruntfile.js
@@ -23,6 +23,42 @@ module.exports = function ( grunt ) {
 		return files;
 	};
 
+	const rtlcssPluginSwapDashiconArrows = {
+		name: 'swap-dashicons-left-right-arrows',
+		priority: 10,
+		directives: {
+			control: {},
+			value: [],
+		},
+		processors: [
+			{
+				expr: /content/im,
+				action: function ( prop, value ) {
+					if ( value === '"\\f141"' ) {
+						// replace dashicons-arrow-left with -right.
+						value = '"\\f139"';
+					} else if ( value === '"\\f340"' ) {
+						// replace dashicons-arrow-left-alt with -right-alt.
+						value = '"\\f344"';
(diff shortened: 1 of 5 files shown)


coreymckrill (approved):
👍 Looks good to me. I tested using the [RTL Tester](https://wordpress.org/plugins/rtl-tester/) plugin.

Issue #268, opened 2021-07-05
Pattern List: Author archive header navigation


ryelle:
The author archive page should use the same header as categories, and show the current author:

![author](https://user-images.githubusercontent.com/541093/124509286-b0c26300-dd9f-11eb-91a2-93005fc957dc.png)

The category navigation above should also filter the author's patterns, for example clicking "Buttons" should show the patterns in category Buttons by ryelle.

ryelle:
This could use some of the `PatternGridMenu` iterations from #284

ryelle:
Fixed by #305.

PR #269, opened 2021-07-06
Theme: Update main page header & search

StevenDufresne:
This is an alternative to #232 to complete #160.

Differences from designs:
- The search icon for the search bar doesn't match.
  - It uses the current dashicons that are loaded and not Gutenberg's icon.
- The breadcrumb text doesn't match the design.
  - I suggest we follow that up with a different pull request

The search page doesn't have a great 404 view. That is contained in this ticket #116.
    
### Screenshots

| Page  | Large | Small | 
| --- | --- | --- | 
|  Home |  |  |
|  Search |  |  |

1. Visit '/'
2. Search for a keyword
3. Click through the results


public_html/wp-content/themes/pattern-directory/header.php
```diff
@@ -27,9 +27,10 @@
 		<header id="masthead" class="site-header <?php echo is_home() ? 'home' : ''; ?>" role="banner">
 			<div class="site-branding">
 				<?php if ( is_home() ) : ?>
-					<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php echo esc_html_x( 'Pattern Directory', 'Site title', 'wporg-patterns' ); ?></a></h1>
+					<h1 class="site-title"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" rel="home"><?php echo esc_html_x( 'WordPress Patterns', 'Site title', 'wporg-patterns' ); ?></a></h1>
```
ryelle: I know the design says "WordPress Patterns", but the "WordPress" feels redundant. Themes & plugins are just called "Themes" and "Plugins", so maybe this should just be "Patterns".

public_html/wp-content/themes/pattern-directory/header.php
```diff
@@ -58,15 +65,15 @@ class="menu-toggle dashicons dashicons-arrow-down-alt2"
...
-							) );
-							?>
-						</div>
+						<?php
+						wp_nav_menu( array(
+							'theme_location' => 'primary',
+							'menu_id'        => 'primary-menu',
+						) );
+						?>
+
```
ryelle, line 74: The nav is now misaligned on mobile:
Screen Shot 2021-07-07 at 6 40 00 PM

public_html/wp-content/themes/pattern-directory/searchform.php
```diff
@@ -0,0 +1,28 @@
+<?php
+/**
+ * Template for search form.
+ *
+ * @link https://codex.wordpress.org/Template_Hierarchy
+ *
+ * @package WPBBP
```
ryelle:
```suggestion
 * @package WordPressdotorg\Pattern_Directory\Theme
```

public_html/wp-content/themes/pattern-directory/src/components/category-context-bar/messaging.js
```diff
@@ -50,45 +50,16 @@ export const getLoadingMessage = ( categoryName ) => {
...
-export const getAllSearchMessage = ( count, searchTerm ) => {
+export const getSearchMessage = ( count, searchTerm ) => {
 	return createInterpolateElement(
 		sprintf(
 			/* translators: %1$d: number of patterns. %2$s search term.  */
 			_n(
-				'%1$d pattern matching "<b>%2$s</b>".',
-				'%1$d patterns matching "<b>%2$s</b>".',
+				'%1$d pattern found for <b>%2$s</b>',
+				'%1$d patterns found for <b>%2$s</b>',
```
ryelle, line 62: Something's going wrong here, I see the plural string when it shouldn't be:
Screen Shot 2021-07-07 at 6 34 11 PM

Ah, it's because the `resultCount` from HTML is actually a string, `_n` doesn't understand.

public_html/wp-content/themes/pattern-directory/src/components/category-context-bar/index.js
```diff
@@ -1,23 +1,28 @@
...
  */
 import { useRoute } from '../../hooks';
-import { getCategoryFromPath } from '../../utils';
-import { getAllSearchMessage, getDefaultMessage, getDefaultSearchMessage, getLoadingMessage } from './messaging';
+import { getCategoryFromPath, getSearchTermFromPath } from '../../utils';
+import { getDefaultMessage, getLoadingMessage, getSearchMessage } from './messaging';
 import { store as patternStore } from '../../store';
 
-function CategoryContextBar() {
+function CategoryContextBar( { resultCount } ) {
```
ryelle: Rather than passing `resultCount` down from the PHP template, you can return a count in the `useSelect`, like so: `count: getPatternTotalsByQuery( query )` -- that will get the total number of results for the current query, so you can use that instead of `category.count` and `resultCount`. It's also a number, so you won't have the string issue in `_n`.

public_html/wp-content/themes/pattern-directory/src/index.js
```diff
@@ -33,3 +34,9 @@ for ( let i = 0; i < previewContainers.length; i++ ) {
 		container.hidden = false;
 	} );
 }
+
+// Load the patterns search package
+const searchContainer = document.getElementById( 'patterns-search__container' );
+if ( searchContainer ) {
+	render( <PatternsSearch { ...searchContainer.dataset } />, searchContainer );
```
ryelle, line 41: If you use `getPatternTotalsByQuery` to get the count in `CategoryContextBar`, you don't need to pass down `resultCount`, and this doesn't need to be a separate component.

public_html/wp-content/themes/pattern-directory/src/utils/query.js
```diff
@@ -104,3 +104,13 @@ export const getCategoryFromPath = ( path ) => {
 export const getPageFromPath = ( path ) => {
 	return Number( getValueFromPath( path, 'page' ) || 1 );
 };
+
+/**
+ * Retrieve the page from a url path.
+ *
+ * @param {string} path
+ * @return {number} The page number.
```
ryelle:
```suggestion
 * Retrieve the search term from a url path.
 *
 * @param {string} path
 * @return {string} The search term.
```

ryelle (commented):
Overall this looks good! I tested out a bunch of search queries, categories, pagination, and author pages.

I noticed that (understandably) the label doesn't update when the category changes. Maybe this could be managed with a `useEffect` in `QueryMonitor`, just to replace that text…?
Screen Shot 2021-07-07 at 6 32 43 PM

public_html/wp-content/themes/pattern-directory/src/index.js
StevenDufresne: Yeah, that's definitely better. Thanks.

public_html/wp-content/themes/pattern-directory/src/utils/query.js
StevenDufresne: 🤦 . Gracias.

public_html/wp-content/themes/pattern-directory/header.php
StevenDufresne: Can I get some device details? I am unable to reproduce:

| FireFox | Chrome Mobile |  Safari  |
| --- | --- | --- | 
|   |  |  |

public_html/wp-content/themes/pattern-directory/src/components/category-context-bar/index.js
StevenDufresne: Thanks, I didn't realize it wouldn't be converted to a number. Will follow your suggestions. 👍

public_html/wp-content/themes/pattern-directory/header.php
ryelle: It was firefox, but maybe my CSS was cached/not fully rebuilt -- i'll try again tomorrow.

ryelle: Nope, I'm still seeing this after rebuilding and clearing cache. I have a menu set in this location, I see the issue logged in and logged out, on Safari, Chrome, and Firefox.

Screen Shot 2021-07-08 at 8 48 03 AM

public_html/wp-content/themes/pattern-directory/src/components/breadcrumb-monitor/index.js
```diff
@@ -0,0 +1,58 @@
...
+ * Update the breadcrumb text to the category name.
+ *
+ * @param {string} categoryName
+ */
+const setBreadcrumbText = ( categoryName ) => {
+	const breadcrumb = document.getElementById( 'breadcrumb-part' );
+	if ( breadcrumb ) {
+		const split = breadcrumb.innerText.split( ':' );
+
+		breadcrumb.innerText = `${ split[ 0 ] }: ${ categoryName }`;
```
ryelle: I don't think you can rely on this being `:` in every language, so it would be better to replace the whole "section: term" string, using a placeholder for the term, `__( 'Category: %s' )`, `__( 'Author: %s' )`.

ryelle: Replacing the whole string would also prevent this ("Author: Columns"):
Screen Shot 2021-07-08 at 9 05 02 AM

public_html/wp-content/themes/pattern-directory/src/components/patterns/index.js
```diff
@@ -19,6 +21,7 @@ const Patterns = () => {
 	return (
 		<RouteProvider>
 			<QueryMonitor />
+			<BreadcrumbMonitor />
 			<PatternGridMenu />
```
ryelle, line 25: I still think you could get rid of `PatternsSearch`, you'd just need to check for a search value before showing the menu. What do you think?

```js
const query = useSelect( ( select ) => select( patternStore ).getCurrentQuery() );
// Only show the full menu if the query does not have a search term.
const showMenu = ! query?.search;

return (
	<RouteProvider>
		<QueryMonitor />
		<BreadcrumbMonitor />
		{ showMenu ? <PatternGridMenu /> : <CategoryContextBar /> }
```

StevenDufresne: Right, it was a bit of a toss up but I chose a separate component because of the yet to be implemented search error page:



We could pass an empty results views to the pattern grid but historically I have run into more problems by passing components than I have composing components.

public_html/wp-content/themes/pattern-directory/src/components/breadcrumb-monitor/index.js
StevenDufresne: I hadn't considered other languages. Thanks for the notice. I'll update.

In terms of the `Author` problem. That currently happens because there's a bigger problem with how we currently handle urls and how it works with`author/admin/` (it replaces it when you click on a category). I'm hesitant to handle that case in this PR because my gut is telling me we'll probably want a separate component or risk overcomplicating some of the "default" components to handle too many different cases.

public_html/wp-content/themes/pattern-directory/header.php
StevenDufresne: Ah very interesting, once the menu is explicitly set, meaning "Primary" is selected, the markup changes. 
| Not Selected | Selected |
| --- | --- |
|  | 

I decided to widen the CSS to remove potential confusion for developers working locally.

Ref: b89bba3

public_html/wp-content/themes/pattern-directory/src/components/breadcrumb-monitor/index.js
ryelle: Yeah, I wasn't clear -- Author stuff can all be handled in #268. I don't think we'll need a separate component, but the header nav might need some cleanup/refactor since it's all so Category specific right now.

StevenDufresne: Do you think there's anything left here?

ryelle: Nope, breadcrumb nav looks good 👍🏻

public_html/wp-content/themes/pattern-directory/src/components/patterns/index.js
ryelle: I have another PR that I just need to write up for the empty results here, but it's the same empty grid as when there are no items in a category, so I still don't think this needs a separate component -- but I don't think I follow your second comment.

StevenDufresne: When viewing `author/wordpressdotorg/`, clicking on any of the categories changes the URL from `author/wordpressdotorg/` to `pattern-categories/{category}/`. 

We have an issue to address that: [issue](268).

We'll need to figure out a new URL structure.  Maybe: `author/wordpressdotorg/{category}`. 

The `PatternGridMenu`, will probably not work with that new URL structure by default. It will need to be refactored if we want to use it. Refactoring may lead to conditional views, which I don't like too much. I think composing a new `PatternGridMenu` maybe `AuthorPatternGridMenu` to keep the controls simple will be a better approach. 

If we went with that approach, having an `AuthorPatterns` will allow us to compose that page, and then we don't have to include the `BreadcrumbMonitor` because `Author: Wordpress.org` or `Author: Wordpress.org : Gallery` is not a great breadcrumb. It would probably be something like `Author / Wordpress.org / Gallery`.....

.... hope I'm explaining my thoughts well enough.. regardless, I don't have any changes on the PR that I'm waiting on.

ryelle: I think we just disagree in approach. I would rather see `PatternGridMenu` used on the author page too, to create URLs like `/author/wordpressdotorg/gallery`, or even `/author/wordpressdotorg/?pattern-category=gallery` if the rewrite is tricky.

On the breadcrumb, I think it should be `All Patterns / Author: Wordpress.org / Gallery` - but that can be done in `BreadcrumbMonitor`, we don't need to create a special `AuthorBreadcrumbMonitor`.

ryelle (approved):
Approving so we can merge & move forward, but I still think we can get rid of the `PatternsSearch` component.

PR #272, opened 2021-07-07
Theme: Fix drag handle precision and minimum value

### LINKED ISSUE #233: Round the precision of the viewport width control
The current control that shows the viewport width is currently showing numbers with high precision, which I don't think is very useful to the users.

Screenshot 2021-06-28 at 15 33 25

I think we could just round the number with `Math.floor` or `Math.round` to get an number that is easier to read and use.

StevenDufresne:
This PR uses `Math.floor` to round the pixel preview value and `Math.max` to make sure the value never goes below a minimum value. I wasn't sure what the minimum value should be so I chose `320px` which is considered the narrow view and there aren't many devices that should be below that, even though patterns may be used in contexts less than the full device width. 

Alternatively, we could use `280` which appears to be the minimum size the preview window will allow. idk.

Fixes #233

### Screenshots
N/A

1. Visit any pattern
2. Drag the resize handles and expect to not see decimals.
3. Drag the resize handles downwards and expect to not get below `320px`.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/src/components/pattern-preview/drag-handle.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/pattern-preview/drag-handle.js
@@ -8,11 +8,13 @@ import { useDrag } from 'react-use-gesture';
  */
 import { LEFT, RIGHT } from '@wordpress/keycodes';
 
-function DragHandle( { label, className, onDragChange, direction = 'left', ...props } ) {
+function DragHandle( { label, className, onDragChange, onDragEnd, direction = 'left', ...props } ) {
 	const dragGestures = useDrag( ( { delta, dragging } ) => {
 		const multiplier = direction === 'left' ? -2 : 2;
 		if ( dragging ) {
-			onDragChange( delta[ 0 ] * multiplier, 'mouse' );
+			onDragChange( delta[ 0 ] * multiplier );
+		} else {
+			onDragEnd();
 		}
 	} );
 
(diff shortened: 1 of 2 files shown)


ryelle:
This fixes the decimal point issue, but I'm not sure about the resize minimum. The reason it was allowed to go negative was for people overshooting and dragging back out, it would match the mouse behavior.

On trunk:

https://user-images.githubusercontent.com/541093/124807913-4ab71680-df2c-11eb-9503-ac8c7a7bff4e.mov

On this branch, when dragging back out from the min-stop, it starts expanding early.

https://user-images.githubusercontent.com/541093/124807528-d7150980-df2b-11eb-9ca4-86d67429f0e1.mov

Is there a way we can handle both behaviors? this & @dd32's [video](https://github.com/WordPress/pattern-directory/issues/233#issuecomment-871836007)?

As for 280 vs 320, 280 was an arbitrary number, we can keep it at 320.

dd32:
> The reason it was allowed to go negative was for people overshooting and dragging back out, it would match the mouse behavior.

I'm not sure what's possible, but could the negative be kept only during drag, and once the drag ends / resets, the next drag attempt starts at the current width?

I think that would handle both of these scenario's in the most user-friendly manner.

StevenDufresne:
I think I have a solution that doesn't modify the internal value, except for the keyboard-driven events.

See:
734068d
e9d2c1a

StevenDufresne:
> I'm not sure what's possible, but could the negative be kept only during drag, and once the drag ends / resets, the next drag attempt starts at the current width?

That works well for the mouse behavior but this control also has keyboard behavior.

 I chose to pass the source of the event and do it specifically for that source otherwise the "dragEnd" behavior could be hard to understand if it's used for both mouse events and keyboard events. I can be convinced otherwise though :).

ryelle:
I think splitting the display & internal value is the way to go here, and this fixes my video from yesterday, but I still noticed some weirdness. It would be good to reset the value to a min/max when the drag is released, you can see in the video here that I'm still able to make the internal value go negative, but if i release the click and move back to the handle, i need to pull it all the way back to positive.

https://user-images.githubusercontent.com/541093/124930198-1f393800-dfcf-11eb-801f-b13b56803c7c.mov

I also noticed a (hard to video) issue where the keyboard nav gets stuck under 280? I'm not sure exactly what's happening, but if I arrow-left beyond 280, i need to hit right twice before it moves again.

https://user-images.githubusercontent.com/541093/124930534-69221e00-dfcf-11eb-828b-c9bcf4201de3.mov

StevenDufresne:
I think I got it now 🤡.

Ref: 1729bb5

ryelle (approved):
💯  Works great now!

PR #275, opened 2021-07-07
Directory: Generate translatable strings for taxonomy terms that appear in the UI

### LINKED ISSUE #256: Translate Category Names
**Describe the bug**
Category names should be translatable when used via a translated UI such as http://ja.wordpress.org/patterns/

**To Reproduce**
1. Go to https://ja.wordpress.org/patterns/
2. The section titles should be translated

**Expected behavior**
I expected to see a language I could not understand, instead, I saw English.

**Additional context**
https://wordpress.slack.com/archives/C02RP50LK/p1625195316084600

coreymckrill:
A GitHub Action workflow that periodically runs a script that pulls in terms from specific taxonomies via the REST API and outputs them to a file within the repository, and then automatically commits changes to that file. This allows the taxonomy term labels to be included in the pot file that gets sent to GlotPress, but the generated file that contains the translation strings is located in a part of the repo that never gets synced to the production server via SVN.

This is the same method used to translate taxonomy terms [on the Learn site](https://github.com/WordPress/learn/pull/205).

Note that the script only pulls in terms for the Categories and Flag Reasons taxonomies. Keywords is not included because it's not clear that those should be translated, or that they will even remain a taxonomy.

Fixes #256 

You can try out the script that generates the translation file:

1. Run `composer install` to make sure you have the script's Requests dependency
2. Delete the `extra` folder
3. From the repo root run `php bin/i18n.php`
4. The `extra` folder should reappear with the fully formed translation file

### DIFF
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -1,0 +1,34 @@
+name: I18n
+
+on:
+  schedule:
+    - cron: '0 6,18 * * *'
+
+jobs:
+  translation-strings:
+    name: Translation strings
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set PHP version
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.4
+          tools: composer:v2
+
+      - name: Install dependencies
+        run: composer install
+
+      - name: Run translation script
+        run: php ./bin/i18n.php
+
+      - name: Commit and push
+        # Using a specific hash here instead of a tagged version, for risk mitigation, since this action modifies our repo.
+        uses: actions-js/push@4decc2887d2770f29177082be3c8b04d342f5b64
+        with:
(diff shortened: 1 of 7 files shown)


ryelle:
Does GlotPress pull from this github repo, or does it pull from the synced code on trac? Just thinking that we don't have anything automated to sync changes from here to trac, so if this file is changed and built here, does it still need a manual step to get imported into GlotPress?

coreymckrill:
> Does GlotPress pull from this github repo, or does it pull from the synced code on trac?

It pulls [directly](https://dotorg.trac.wordpress.org/browser/wordpress/website/translate/bin/init-repos.sh#L34) from the GitHub repo, so there's no need to sync to Trac, and no need for the generated file to ever be on the production server.

dd32:
A quick glance through the code seems good to me.

Others might disagree, but I don't forsee the Category list growing significantly in size, Keywords however might be more freeform?

ryelle (approved):
Looks good! I noticed the generator script also picked up the new flag reasons too 👍🏻

Issue #161, opened 2021-06-10
Add ordering support when browsing patterns


ryelle:
The pattern grid archives should have an order dropdown in the top right menu area.

I think this should trigger a query string, like `?orderby=date` etc. Featured might just be a category, not an actual ordering parameter, but that's just a suggestion.

<img width="680" alt="Screen Shot 2021-06-10 at 3 57 22 PM" src="https://user-images.githubusercontent.com/541093/121588982-8bdb0a00-ca04-11eb-8aa6-44c2822f7799.png">

<img width="232" alt="Screen Shot 2021-06-10 at 3 59 20 PM" src="https://user-images.githubusercontent.com/541093/121589224-d3619600-ca04-11eb-85ba-76b2d4d79f9d.png">

Same on My Patterns & Favorites, but no Featured.

<img width="821" alt="Screen Shot 2021-06-10 at 3 57 49 PM" src="https://user-images.githubusercontent.com/541093/121589043-9c8b8000-ca04-11eb-8895-f708e3ca197b.png">

StevenDufresne:
@ryelle @shaunandrews

Did we want to launch with the featured category? I feel like it's a nice-to-have because the management of that category adds complexity.

Also, we have a sort by "Date" but it's unidirectional. Would the label "Newest" make more sense?

ryelle:
No, I think it's late to introduce featured for launch, but it should be added shortly after, maybe a new taxonomy like "Browse" in plugins. Not required to fix this issue.

> Also, we have a sort by "Date" but it's unidirectional. Would the label "Newest" make more sense?

That's more understandable to me 👍🏻

Issue #100, opened 2021-05-10
Display "social currency" metrics about patterns


ryelle:
The design includes views and a favorite count on each pattern. 

- Get the view count (or it a "times copied" count), if we can
- Add that into the CPT's API response for each pattern
- Display it in the grid

![pattern](https://user-images.githubusercontent.com/541093/117689979-51bde480-b188-11eb-804e-3ad178958520.png)

coreymckrill:
So maybe the "times copied" count is a postmeta value on the pattern post? Would that only count uses of the "Copy pattern" button on the wp.org pattern directory, or could we bump the counter when someone picks the pattern via the API from within their site?

shaunandrews:
I think it would be good to record values for:
- Views: How many times has a pattern detail page been viewed. It might be nice to break this down based on _where_ the view was registered; In a query loop, a single pattern, and/or from within the editor.
- Copies: How many times a pattern's copy button has been clicked. It might be nice to break this down based on _where_ the copy button was clicked; In a query loop or on the single screen for the pattern.
- Insertions: How many times a pattern was inserted from within the editor.
- Favorite: How many time's a pattern's favorite button has been clicked. It might be nice to break this down based on _where_ the buttons as clicked; In a query loop, on a single screen, or within the editor.

ryelle:
I think we'll get pushback if we try to send back metrics from a user's site (privacy concerns).

Do we need to worry about performance with the idea of incrementing a counter with every copy?

Could we use Jetpack/tracks to offload handling the metrics, or does that open a new can of privacy worms?

StevenDufresne:
Can we start with more simple metrics and expand? When I look at something like dribbble.com, they count favorites and views as the social indicators. If we took the same approach we would only need to track things on wordpress.org. It may help scope this down for the initial launch and if those values aren't good enough, we could ask users what they think is worth tracking.

shaunandrews:
Sure. The designs already show the views and favorites as the primary metrics. I was more thinking of what could be helpful in the future as we iterate on the directory and pattern discovery/exploration.

ryelle:
I added a fake favorite count in #164, but I'd like some help getting the real value. I started with this query:

```
$count = $wpdb->get_var( $wpdb->prepare(
	"SELECT COUNT(*)
		FROM {$wpdb->usermeta}
		WHERE meta_key=%s
		AND meta_value=%d",
	META_KEY,
	$post->ID
) );
```

[...]

PR #276, opened 2021-07-07
Patterns: Show real favorite counts

ryelle:
See #100, follow up to #93/#94. This displays the current number of favorites on a given pattern, using a "cached" value on post meta. The post meta value is updated any time a user favorites or unfavorites that pattern. When adding & removing favorites, the favorite count for the pattern is returned on success, so the value can be updated live after clicking favorite.

I decided to cache/save the value to post meta, since we'll also want to order by favorites in #161.

If there are no issues with this "cache"-ish solution, I also have a bin script that can be run to update the wporg patterns (since people can favorite patterns right now, there might be a few already favorited).

### Screenshots

Screen Shot 2021-07-07 at 6 04 18 PM

1. Optionally have a site with multiple users
2. Favorite some patterns on either/both users
3. The pattern counts should be updated instantly, and should be correct - if you favorite something with two users, it should show "2 ❤️ "

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/favorite.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/favorite.php
@@ -84,3 +84,81 @@ function get_favorites( $user = 0 ) {
 
 	return array_map( 'absint', $favorites );
 }
+
+/**
+ * Get the cached count of how many times this pattern has been favorited.
+ *
+ * @param int|WP_Post $post The pattern to check.
+ * @return integer
+ */
+function get_favorite_count( $post = 0 ) {
+	$post = get_block_pattern( $post );
+	if ( ! $post ) {
+		return false;
+	}
+
+	return absint( get_post_meta( $post->ID, META_KEY, true ) );
+}
+
+/**
+ * Get a count of how many times this pattern has been favorited, directly from the users table.
+ *
+ * @param int|WP_Post $post The pattern to check.
+ * @return integer
+ */
+function get_raw_favorite_count( $post = 0 ) {
+	global $wpdb;
+	$post = get_block_pattern( $post );
+	if ( ! $post ) {
+		return false;
+	}
+	$count = $wpdb->get_var( $wpdb->prepare(
+		"SELECT COUNT(*)
+			FROM {$wpdb->usermeta}
+			WHERE meta_key=%s
+			AND meta_value=%d",
+		META_KEY,
+		$post->ID
+	) );
+
+	return absint( $count );
+}
+
+/**
+ * Update a given post's favorite count cache.
+ *
+ * @param mixed $value The post ID.
+ */
+function update_favorite_cache( $value ) {
+	if ( ! is_numeric( $value ) ) {
+		return;
+	}
+
+	$count = get_raw_favorite_count( $value );
+	update_post_meta( $value, META_KEY, $count );
+}
+
+/**
+ * Trigger the update of favorite count when a user favorites or unfavorites a pattern.
+ *
+ * @param int    $mid         The meta ID.
+ * @param int    $object_id   ID of the object metadata is for.
+ * @param string $meta_key    Metadata key.
+ * @param mixed  $_meta_value Metadata value. Serialized if non-scalar.
+ */
+function trigger_favorite_cache_update( $mid, $object_id, $meta_key, $_meta_value ) {
+	if ( META_KEY !== $meta_key ) {
+		return;
+	}
+
+	// This value can be an array in the delete action, so walk through each unique value and refresh the cache.
+	if ( is_array( $_meta_value ) ) {
+		$_meta_value = array_unique( $_meta_value );
+		array_walk( $_meta_value, __NAMESPACE__ . '\update_favorite_cache' );
+		return;
+	}
+
+	update_favorite_cache( $_meta_value );
+}
+add_action( 'added_user_meta', __NAMESPACE__ . '\trigger_favorite_cache_update', 10, 4 );
+add_action( 'deleted_user_meta', __NAMESPACE__ . '\trigger_favorite_cache_update', 10, 4 );
--- a/public_html/wp-content/plugins/pattern-directory/includes/favorite.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/favorite.php
@@ -3,6 +3,7 @@
 namespace WordPressdotorg\Pattern_Directory\Favorite;
 use function WordPressdotorg\Pattern_Directory\Pattern_Post_Type\get_block_pattern;
 
+// Used for both the post meta (count of favorites) and user meta (list of pattern IDs).
 const META_KEY = 'wporg-pattern-favorites';
 
 /**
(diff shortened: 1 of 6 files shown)


StevenDufresne:
This flow isn't working for me:
1. Visit `/` 
2. Click on a specific pattern
3. Favorite specific pattern
4. Click the back button
5. Notice that the pattern doesn't include the new favorite even though the heart itself is red. I would expect to see the heart icon under the title.

[Example Video](https://d.pr/v/ruFgRg)

ryelle:
Is that Chrome? Firefox just goes back and the pattern appears unfavorited, no new API requests are fired. In Chrome, it looks like the API requests are fired but the patterns response is pulled from disk cache, not the real API values. So it pulls the cached `favorite_count: 0` even though the real value is now `1`.

I'm not sure why `GET /wp-json/wporg/v1/pattern-favorites` request does not pull from cache while `GET /wp-json/wp/v2/wporg-pattern` does.

StevenDufresne:
> Is that Chrome? Firefox just goes back and the pattern appears unfavorited, no new API requests are fired. In Chrome, it looks like the API requests are fired but the patterns response is pulled from disk cache, not the real API values. So it pulls the cached `favorite_count: 0` even though the real value is now `1`.
> 
> I'm not sure why `GET /wp-json/wporg/v1/pattern-favorites` request does not pull from cache while `GET /wp-json/wp/v2/wporg-pattern` does.

Yeah, it is in Chrome. For chrome, we could either "bust" the pattern request with a datetime or update the `Cache-control` headers and add `no-store`. That should force the correct updates?

As for firefox, I'm not sure how to force the request, it's most likely cache control related?

ryelle:
Do you think this is a problem that needs to block the merge of this PR? It's not technically related to the favoriting action, it's just more obvious thanks to Chrome's behavior.

Personally, Firefox's behavior seems expected - hitting back shouldn't refresh the page, it should bring me back to the previous state.

StevenDufresne (commented):
> Do you think this is a problem that needs to block the merge of this PR? It's not technically related to the favoriting action, it's just more obvious thanks to Chrome's behavior.
> 
> Personally, Firefox's behavior seems expected - hitting back shouldn't refresh the page, it should bring me back to the previous state.

I would expect it to always reflect the current state of the database but I don't mind if this comes back as a bug, if it ever does.

@coreymckrill Any feedback on the PHP side?

public_html/wp-content/plugins/pattern-directory/includes/favorite.php
```diff
@@ -84,3 +84,81 @@ function get_favorites( $user = 0 ) {
...
+ * @param int|WP_Post $post The pattern to check.
+ * @return integer
+ */
+function get_favorite_count( $post = 0 ) {
+	$post = get_block_pattern( $post );
+	if ( ! $post ) {
+		return false;
+	}
+
+	return absint( get_post_meta( $post->ID, META_KEY, true ) );
```
coreymckrill, line 101: Since this is using the same constant value for both a user meta key and a post meta key, I think it would be worth explaining that in a comment up where `META_KEY` is defined.

public_html/wp-content/plugins/pattern-directory/includes/favorite.php
```diff
@@ -84,3 +84,81 @@ function get_favorites( $user = 0 ) {
...
+		$post->ID
+	) );
+
+	return absint( $count );
+}
+
+/**
+ * Update a given post's favorite count cache.
+ *
+ * @param mixed $value The post ID.
```
coreymckrill: For clarity and consistency, it might be better for this function's parameter to be named `$post_id` and set its type to `int` in the doc block.

public_html/wp-content/plugins/pattern-directory/includes/favorite.php
```diff
@@ -84,3 +84,81 @@ function get_favorites( $user = 0 ) {
...
+ */
+function update_favorite_cache( $value ) {
+	if ( ! is_numeric( $value ) ) {
+		return;
+	}
+
+	$count = get_raw_favorite_count( $value );
+	update_post_meta( $value, META_KEY, $count );
```
coreymckrill:
```suggestion
	$count = get_raw_favorite_count( $value );
	
	if ( ! is_int( $count ) ) {
		return;
	}

	update_post_meta( $value, META_KEY, $count );
```

I think this is a bit cleaner since `get_raw_favorite_count` is really where you check that you have a valid pattern post.

public_html/wp-content/plugins/pattern-directory/includes/favorite.php
```diff
@@ -84,3 +84,81 @@ function get_favorites( $user = 0 ) {
...
+
+/**
+ * Trigger the update of favorite count when a user favorites or unfavorites a pattern.
+ *
+ * @param int    $mid         The meta ID.
+ * @param int    $object_id   ID of the object metadata is for.
+ * @param string $meta_key    Metadata key.
+ * @param mixed  $_meta_value Metadata value. Serialized if non-scalar.
+ */
+function trigger_favorite_cache_update( $mid, $object_id, $meta_key, $_meta_value ) {
```
coreymckrill: Took me a minute to remember that `$object_id` is actually the user ID here, and `$_meta_value` is the post ID or IDs. Might be helpful to rename these to be more explicit.

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
```diff
@@ -290,7 +291,7 @@ function register_rest_fields() {
 		'favorite_count',
 		array(
 			'get_callback' => function() {
-				return 1;
+				return get_favorite_count( get_the_ID() );
 			},
 
 			'schema' => array(
```
coreymckrill, line 297: Would it be worth adding a `'default' => 0` to the schema here, so that there's always a value for this field even if there are no favorites?

coreymckrill (approved):
Some non-blocking nitpicks, but otherwise PHP looks good and works as expected

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
ryelle: `get_favorite_count` will always return an int, even if there are no favorites, but I'll add it.

public_html/wp-content/plugins/pattern-directory/includes/favorite.php
ryelle: I'd prefer to keep these matching the core behavior - `trigger_favorite_cache_update` is called on every meta value, so `$_meta_value` isn't always the post ID. I updated `$object_id`, but want to leave `$_meta_value`.

public_html/wp-content/plugins/pattern-directory/includes/favorite.php
ryelle: Thanks, I think `update_favorite_cache` is more understandable now 👍🏻

public_html/wp-content/plugins/pattern-directory/includes/favorite.php
ryelle: I updated this to be `post_id`, but the value _could_ technically be anything, so I wouldn't want something to go wrong if someone expected it to be an int.

PR #277, opened 2021-07-08
Report pattern modal: Sort reason terms by slug for custom sort order

coreymckrill:
It makes the most sense for the "Other" reason to be last in the list, but since it was added first, and we might add other reasons later, sorting by the ID makes it hard to keep it there. Sorting by slug will make this easier because we can customize the slug through the UI to be things like `1-inappropriate`, `2-other` etc.

### Screenshots

**Before**

reasons-before

**After**

reasons-after

1. Add some terms to the [Pattern Flag Reason taxonomy](http://localhost:8888/wp-admin/edit-tags.php?taxonomy=wporg-pattern-flag-reason&post_type=wporg-pattern)
2. Open the Report This Pattern modal on the front end and check their order
3. Try modifying the terms' slugs so they will sort differently


ryelle (approved):
Looks good 👍🏻

Issue #244, opened 2021-06-29
In the pattern search API, use the client locale when querying


iandunn:
See #223 / #224

Pulls locale from post meta, see #229 

# Pre-launch

- [x] Use `locale` param when querying patterns (`WP_Query`)
- [x] Add e2e test -- https://meta.trac.wordpress.org/changeset/11082/

# Post-launch

- [ ] After, #28 is done, switch to ES to mitigate performance issues w/ meta queries. Do for front-end too, like #273
- [x] Prioritize search based on locale (ala Learn), rather than simple include/exclude. Do for the API, but also for the front end (related #273)

dd32:
d75ca62a870e111d7d55666cae760fc34d2a2702 causes a bit of a shortcoming, in that patterns without a translation become inaccessible.

Instead, I think we need to prioritise translated patterns but not skip not-localised.

iandunn:
Do you mean that requests for `es_MX` patterns won't include `en_US` "parent" patterns that don't have a `es_MX` "child/translated" variant/fork yet?

That seems good enough for launch, since all the "core/featured" patterns will be translated.

We might want to wait until after #28 too, since ES might make that prioritization easier. On the other hand, Learn already does that kind of sorting w/ `WP_Query`, so we could just copy that 🤔

iandunn:
Removing this from the milestone since the pre-launch tasks are done.

ryelle:
It looks like all the linked PRs have been merged, so I'm going to close this issue.

PR #278, opened 2021-07-10
Limit front end patterns to site locale

### LINKED ISSUE #273: Theme: Query results show translated patterns side by side.
**Describe the bug**
I'm not sure if it's enclosed in another ticket but I'll add this in case. 

Since we now have translated patterns, the search results are showing translated versions of the same pattern making the pattern list a bit useless.

**To Reproduce**
1. Go to `wordpress.org/patterns`
2. View the result list, expect to see the same patterns in different languages.

**Expected behavior**
We should only see patterns matching our locale.

**Screenshots**

 - OS: macOs
 - Browser Chrome
 - Version July 7, latest :)

iandunn:
This makes it so that Rosetta sites like `es-mx.wordpress.org/patterns` will only display patterns with a `es_MX` locale.

In the future, we may want to include all locales, but weight translated ones higher (see https://github.com/WordPress/pattern-directory/issues/244#issuecomment-871985952), but this is good enough for the initial launch.

Fixes #273

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -73,10 +73,11 @@ function enqueue_assets() {
 		wp_add_inline_script(
 			'wporg-pattern-script',
 			sprintf(
-				'var wporgAssetUrl = "%s", wporgSiteUrl = "%s", wporgLoginUrl = "%s";',
+				'var wporgAssetUrl = "%s", wporgSiteUrl = "%s", wporgLoginUrl = "%s", wporgLocale = \'%s\';',
 				esc_url( get_stylesheet_directory_uri() ),
 				esc_url( home_url() ),
-				esc_url( wp_login_url() )
+				esc_url( wp_login_url() ),
+				wp_json_encode( get_locale() )
 			),
 			'before'
 		);
@@ -164,6 +165,9 @@ function pre_get_posts( $query ) {
 	} else if ( ! $query->get( 'pagename' ) && 'post' === $query->get( 'post_type', 'post' ) ) {
 		$query->set( 'post_type', array( POST_TYPE ) );
 		$query->set( 'post_status', array( 'publish' ) );
+
+		$query->set( 'meta_key', 'wpop_locale' );
+		$query->set( 'meta_value', get_locale() );
 	}
 }
 
--- a/public_html/wp-content/themes/pattern-directory/package.json
+++ b/public_html/wp-content/themes/pattern-directory/package.json
@@ -63,7 +63,8 @@
 		"globals": {
 			"wporgAssetUrl": "readonly",
 			"wporgLoginUrl": "readonly",
-			"wporgSiteUrl": "readonly"
+			"wporgSiteUrl": "readonly",
+			"wporgLocale": "readonly"
 		}
 	},
 	"stylelint": {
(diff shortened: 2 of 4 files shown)


iandunn:
This kinda works, but I'm guessing it not exactly the way @StevenDufresne or @ryelle would do it. Do y'all have any thoughts?

public_html/wp-content/themes/pattern-directory/src/components/patterns/index.js
```diff
@@ -16,6 +16,11 @@ import { store as patternStore } from '../../store';
 const Patterns = () => {
 	const query = useSelect( ( select ) => select( patternStore ).getCurrentQuery() );
 
+	if ( query ) {
+		query.locale = wporgLocale;
+		// todo why does it still say "loading all patterns"?
```
StevenDufresne: This is happening because it's stuck in an infinite loop.

Move this code into the `QueryMonitor` control. Probably [here](https://github.com/WordPress/pattern-directory/blob/4d05a3f95e556ff4c64db43a9e920e3c544e558a/public_html/wp-content/themes/pattern-directory/src/components/query-monitor/index.js#L60).

ryelle:
I would add the locale to the query at a lower level, in the API request in the resolver itself:

https://github.com/WordPress/pattern-directory/blob/173cd98d161887536a9d2883b8cfbb38e2d06eee/public_html/wp-content/themes/pattern-directory/src/store/resolvers.js#L49

```diff
- path: addQueryArgs( '/wp/v2/wporg-pattern', query ),
+ path: addQueryArgs( '/wp/v2/wporg-pattern', { ...query, locale: wporgLocale } ),
```

This would affect all pattern-grid queries - the main page, categories, authors, and "My Patterns". The only one I could see that being weird on is My Patterns, but I think we can ignore that for now - pattern creation is not ready for public, so the only user that has a My Patterns page is `wordpressdotorg`. (TBH I also think that this would be the expected behavior, I don't think someone who creates a pattern on `ja.wordpress.org/patterns` would expect to see it on `es.wordpress.org/patterns` - but that can be a discussion later)

Going to a pattern directly uses a different resolver/API request, so any locale pattern can be viewed directly, ex: `http://es.wordpress.org/patterns/pattern/two-columns-of-text-with-images-ja/`

public_html/wp-content/themes/pattern-directory/src/components/patterns/index.js
iandunn: Thanks!

iandunn:
That sounds good, thanks! Tested this on my w.org sandbox and it's looking good.

StevenDufresne:
> I would add the locale to the query at a lower level, in the API request in the resolver itself:
> 
> https://github.com/WordPress/pattern-directory/blob/173cd98d161887536a9d2883b8cfbb38e2d06eee/public_html/wp-content/themes/pattern-directory/src/store/resolvers.js#L49
> 
> ```diff
> - path: addQueryArgs( '/wp/v2/wporg-pattern', query ),
> + path: addQueryArgs( '/wp/v2/wporg-pattern', { ...query, locale: wporgLocale } ),
> ```

Yep, this works wells too. I suggested the `QueryMonitor` in case we planned to add a locale switcher to the interface. If we did so and it was a non-reload-the-page version, we'd want to pull from the query string. But early days for those kinds of thoughts 👍 .

PR #280, opened 2021-07-13
Patterns: Handle empty results and 404 pages

### LINKED ISSUE #116: Pattern Grid: Blank page when no patterns are found

ryelle:
Fixes #116. This adds the "empty state" and "404 page" styles. When a category or query result is empty, the page should say that, then display 6 recent patterns. If the page is not found, it should display the "404" page template. This template is handled by PHP, so we can remove the `bypass_404_page` filter.

The 404 page in the design also has "404 patterns" on it, but we don't have that category set up, so I've left it empty.

### Screenshots

Empty search result (Japanese site):

nothing-ja

A locale with no translated patterns (Korean) will also show the empty state on the homepage, but there are no recent patterns either. Ideally this would be a message to get involved, but that can happen in another PR/post launch.

nothing-ko

The 404 page, for example, `wordpress.org/patterns/error404`:

1. Search for a string that doesn't exist
2. You should see the empty state
3. Go to a page or category that doesn't exist
4. You should see the 404 page

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/404.php
+++ b/public_html/wp-content/themes/pattern-directory/404.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * The template for displaying 404 pages (not found).
+ *
+ * @link https://codex.wordpress.org/Creating_an_Error_404_Page
+ *
+ * @package WordPressdotorg\Pattern_Directory\Theme
+ */
+
+namespace WordPressdotorg\Pattern_Directory\Theme;
+
+get_header(); ?>
+
+	<main id="main" class="site-main col-12" role="main">
+
+		<div class="error-404 not-found">
+			<div class="page-container aligncenter">
+				<header class="page-header">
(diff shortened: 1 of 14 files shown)


StevenDufresne:
Looking at the first screenshot, this seems to work with the branch pre: aa775d8.

When I find nothing on the search page, I don't see my expected view:

StevenDufresne:
Relates to #280.

ryelle:
Ah, because the search uses the separate component and I didn't fix that in the rebase. Will update this there too.

PR #281, opened 2021-07-13
Patterns: Re-enable revisions support for pattern post type

### LINKED ISSUE #279: Patterns: Use revisions to track pattern changes
The revisions were disabled on the pattern post type to disable the Revisions UI in the creator (because it's broken), but we actually need revisions to track changes to patterns.

We should re-enable revisions for now, and find another solution for disabling them in the Pattern Creator.

coreymckrill:
Allows us to track changes to patterns.

In #279 it says that the revisions panel is broken in the front end pattern creator UI, but after enabling revisions, I am not seeing the revisions panel on the front end at all, even though it appears normally in the WP Admin block editor. Perhaps nothing more needs to be done here?

Fixes #279

1. Edit a pattern a few times.
2. Check to see if the revisions panel shows up in the WP Admin block editor.
3. Check to see if the revisions panel shows up in the front end pattern creator.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -55,7 +55,7 @@ function register_post_type_data() {
 			'public'          => true,
 			'show_in_rest'    => true,
 			'rewrite'         => array( 'slug' => 'pattern' ),
-			'supports'        => array( 'title', 'editor', 'author', 'custom-fields' ),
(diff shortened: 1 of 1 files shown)


ryelle (approved):
> it says that the revisions panel is broken in the front end pattern creator UI, but after enabling revisions, I am not seeing the revisions panel on the front end at all

Yep, looks good - the panels displayed in the creator are more controlled now, previously it showed the same "2 Revisions" link which tried (and failed) to go back to wp-admin.

PR #283, opened 2021-07-14
Consistency in very similar strings

### LINKED ISSUE #274: Merge two strings
While translating strings for the pattern directory, I stumbled upon two similar strings that probably could be merged to use exactly the same string:
The description is used to help users of assistive technology understand the content of your pattern.
Ref: https://github.com/WordPress/pattern-directory/blob/trunk/public_html/wp-content/plugins/pattern-directory/src/pattern-post-type/details.js#L52

and
The description is used to help users of assistive technology better understand the contents of your pattern.
https://github.com/WordPress/pattern-directory/blob/trunk/public_html/wp-content/plugins/pattern-creator/src/components/publish-modal/submit.js#L98

vlad-timotei:
Replaces the string to be identical with the one [here](https://github.com/WordPress/pattern-directory/blob/aa775d8caeb2f66946b7bed7ce5300eb18135bcb/public_html/wp-content/plugins/pattern-directory/src/pattern-post-type/details.js#L52).

As @ryelle suggested, this PR uses the shorter string alternative. This makes things much easier for translators too.

Fixes #274. Props @tobifjellner.

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/src/components/publish-modal/submit.js
+++ b/public_html/wp-content/plugins/pattern-creator/src/components/publish-modal/submit.js
@@ -96,7 +96,7 @@ export default function SubmitModal( { onSubmit, onClose } ) {
 						value={ description }
 						placeholder={ __( 'Describe the output of pattern', 'wporg-patterns' ) }
 						help={ __(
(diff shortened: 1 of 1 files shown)


ryelle (approved):
Thanks for fixing this!

PR #284, opened 2021-07-14
Theme: Add favorites page

### LINKED ISSUE #239: Theme: Create page to show current user's favorite patterns
This page should show all patterns the user has saved as favorites.

Favorites

- Filtering by category
- The sort dropdown should have "Last updated" and "Views"

ryelle:
Fixes #239 - Add a page, `/favorites/`, which will show the current user's favorite patterns. This also updates PatternGridMenu a bit to be used on the Favorites page - now we can pass the current query down from Patterns, along with the basePath used for category links.

### Screenshots

All favorites, ex `wordpress.org/patterns/favorites`
all-fav

Favorites in the Button category, ex `wordpress.org/patterns/favorites/pattern-categories/buttons`
buttons-fav

Favorites are localized, so only English patterns appear on the main site, while Japanese patterns appear on `ja.wordpress.org/patterns/favorites` (mirrors how My Patterns will work, can be iterated on post-launch if we decide this is not how it should work)
ja-fav

1. Make sure you have a page with the slug `favorites`
2. Refresh permalinks, build the branch `yarn workspaces run build`
3. Favorite some patterns
3. View the favorites page, `/favorites/`
4. It should show all your favorites
5. Filter by category
6. It should only show those favorites in that category
7. The counts on the context bar should be correct
8. Unfavorite a pattern, it should be removed from the view


coreymckrill (approved):
👍 Works as expected, testing locally.

There is one transition that I found a bit jarring. When I "unfavorite" a pattern from the Favorites page, there's a flash of a loading animation where it's redoing the query or something and it takes over the whole grid:

Screen Capture on 2021-07-14 at 13-08-32

My expectation was that the unfavorited pattern would just disappear and the other grid items would shift over. Maybe that's not possible in this context, but just a thought.

ryelle:
> My expectation was that the unfavorited pattern would just disappear and the other grid items would shift over. Maybe that's not possible in this context, but just a thought.

Hm, that's a good point… it's because the entire query technically changes, and we're requesting a "new" set of pattern IDs. I don't think there's an easy fix to drop in here. @StevenDufresne any ideas for a quick fix? If not, IMO, we can launch with this and make an issue.

public_html/wp-content/themes/pattern-directory/functions.php
```diff
@@ -178,6 +178,7 @@ function pre_get_posts( $query ) {
  */
 function add_rewrite() {
 	add_rewrite_rule( '^my-patterns/[^/]+/?$', 'index.php?pagename=my-patterns', 'top' );
+	add_rewrite_rule( '^favorites/.+/?$', 'index.php?pagename=favorites', 'top' );
```
StevenDufresne: I would expect this URL to include the `my-` convention as well.....`my-favorites`

public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+	if ( ! favorites.length ) {
+		return (
+			<div className="entry-content">
+				<p>{ __( 'You haven’t favorited any patterns yet.', 'wporg-patterns' ) }</p>
+			</div>
+		);
+	}
+
+	const modifiedQuery = { ...query, include: favorites };
```
StevenDufresne, line 45: I was able to get into a state where I had a favorited pattern but the query returned nothing, probably because that pattern was moved to draft or deleted:



This will be possible for the directory.

public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js
```diff
@@ -0,0 +1,59 @@
...
+import { RouteProvider } from '../../hooks';
+
+const MyFavorites = () => {
+	const { favorites, isLoggedIn, query } = useSelect( ( select ) => ( {
+		favorites: select( patternStore ).getFavorites(),
+		isLoggedIn: !! select( coreStore ).getCurrentUser()?.id,
+		query: select( patternStore ).getCurrentQuery(),
+	} ) );
+
+	if ( ! isLoggedIn ) {
```
StevenDufresne, line 25: This is flashing on page load, can we delay its rendering until after we are certain the user is not logged in? Same goes for the `You haven’t favorited any patterns yet` view.

StevenDufresne (commented):
We could probably pop it out of the list safely because we are not managing state across different 'pages', but I think we should probably come back to deal with it.

public_html/wp-content/themes/pattern-directory/functions.php
ryelle: I was on the fence about it, I'll switch it back to `my-favorites`.

public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js
ryelle: There doesn't seem to be a loading state for users, `currentUser` is `{}` whether waiting or not logged in.

public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js
ryelle: Will fix in the empty-state PR after this is merged.

public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js
ryelle: Updated this to use a value provided by the server, so there's no user loading state anymore… but now it shows `You haven’t favorited any patterns yet` until _that_ loads. Will see if I can address that as part of #280.

PR #287, opened 2021-07-15
Theme: Add ordering to patterns.

### LINKED ISSUE #161: Add ordering support when browsing patterns

StevenDufresne:
This PR adds the ability to sort patterns based on 2 dimensions:
- Newest
- Favorited

In the future, we hope to add support for `Featured` categories.

Fixes #161

## Needs Resolution

Currently, block patterns don't start with the meta key: `wporg-pattern-favorites` set. This means that when we run the sorting query it will only return the patterns that have favorites. That may not be the experience we are looking for. In order to correct it, we'll need to set a default value. Should we `register_post_meta` and set the default value to `0`? 

**Example**:

### Screenshots
| Desktop | Mobile |
| --- | --- |
|   |  |

1. Visit `/`
2. Use the dropdown
3. Expect that the results match


public_html/wp-content/themes/pattern-directory/src/components/pattern-order-select/index.js
```diff
@@ -0,0 +1,40 @@
...
+const PatternOrderSelect = ( { options } ) => {
+	const { path, replace } = useRoute();
+	const hideLabel = useViewportMatch( 'medium', '>=' );
+
+	if ( ! options ) {
+		return null;
+	}
+
+	return (
+		<div className="pattern-select-control">
```
ryelle: This should use the same name as the component/folder, so either the class should be `pattern-order-select` or the component should be `pattern-select-control` (but that sounds like you're selecting a pattern).

ryelle (approved):

Issue #282, opened 2021-07-14
Javascript translations not being loaded


dd32:
**Describe the bug**
WordPress.org isn't deploying json translations for use with `wp.i18n` at present, resulting in any JS strings using `wp.i18n` not being translated.

This is not Pattern-directory specific, this is related to the way WordPress.org deploys translation updates. Currently it only does .mo/.po files, not the .json files. Creating this as a tracking ticket since the Pattern Directory is the most affected project.

**To Reproduce**
1. Go to https://ja.wordpress.org/patterns/
2. Note the loading text in English.
3. Hover over a pattern, Note the copy button in English.

**Expected behavior**
Translations to be present.

**Additional context**

See slack https://wordpress.slack.com/archives/C02RP50LK/p1626243894250200

dd32:
Translations should now load for `wp.i18n.*()` translations in the Pattern Directory 🎉

PR #290, opened 2021-07-15
Add Build job to create a "production ready" branch

dd32:
While working on #282 I've run into an issue where the GlotPress string extraction/import doesn't have access to the final location for the strings contained within the javascript.

This is because we import the strings directly from this GitHub repo, but, the build json translation files need to reference the file that is enqueued - `build/index.js`, not `src/components/pattern-preview-actions/copy-success-message.js`.

This PR adds an action that runs the build process and pushes the resulting changes to a `build` branch, which can then be used as the base for the GlotPress string extraction. As a bonus, it can also be used as a pre-built set of files for easier deployment back to WordPress.org. 

Things of note:
 - The i18n building added in #275 could be merged into this instead of committing to the trunk branch. I'm not sure if we'd want to run the fetch-from-wp-json on every commit here though.
 - I wasn't quite sure what I was doing, and as a result, I had to `git add .. --force` some things that `.gitignore` ignores, there's probably a better way, but it might also be a shortcoming of the commit action chosen.
 - package-lock.json isn't committed to the repo, but yarn.lock and composer.lock are. I `rm`'d it to keep the branch pushes to just the relevant changes.
 - I haven't added a `cache` step into the workflow yet, as such the install process takes ~2minutes in the action. Caching that would speed it up significantly.
 - I'm not sure if the build commands run here are the correct ones to generate a production-ready set of build files.

See #282

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,27 @@
+name: Build and push to build branch.
+
+on:
+  push:
+    branches:
+      - trunk
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Build and Push
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install all dependencies
+        run: npm install && yarn
+
+      - name: Build
+        run: yarn workspaces run build
+    
+      - name: Commit and push
+        # Using a specific hash here instead of a tagged version, for risk mitigation, since this action modifies our repo.
+        uses: actions-js/push@4decc2887d2770f29177082be3c8b04d342f5b64
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: build
+          message: Build: ${{ GITHUB_SHA }})
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
         run: |
              rm package-lock.json
              git add public_html/wp-content/themes/pattern-directory/css/* --force
-             git add public_html/wp-content/plugins/*/build --force
+             git add public_html/wp-content/plugins/pattern-*/build --force
     
       - name: Commit and push
         # Using a specific hash here instead of a tagged version, for risk mitigation, since this action modifies our repo.
@@ -33,4 +33,5 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           branch: build
+          force: true
           message: "Build: ${{ github.sha }}"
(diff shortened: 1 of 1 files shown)


dd32:
An example of the result can be seen here:
 - https://github.com/dd32/pattern-directory/runs/3074069965?check_suite_focus=true
 - https://github.com/dd32/pattern-directory/tree/build
 - https://github.com/dd32/pattern-directory/commit/48c17e39a428386dadf30361f80f490147b3c1e9

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Build and Push
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install all dependencies
+        run: |
+             npm install
```
ryelle: You don't need to `npm install`, since this project uses `yarn` - yarn will install all the relevant dependencies. It also won't generate the `package-lock.json`, so you won't need to remove that.

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+             yarn
+             composer install
+
+      - name: Build
+        run: yarn workspaces run build
```
ryelle: This is the correct command to build everything - the plugins are already building a minified, production-ready bundle, but it looks like the theme isn't set up correctly. I fixed that in #293, so once that's merged, this command will generate all production-ready code.

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Ignore .gitignore
+        run: |
+             rm package-lock.json
+             git add public_html/wp-content/themes/pattern-directory/css/* --force
+             git add public_html/wp-content/plugins/pattern-*/build --force
```
ryelle: This misses the theme's JS bundle:
```suggestion
             git add public_html/wp-content/themes/pattern-directory/css/* --force
             git add public_html/wp-content/themes/pattern-directory/build --force
             git add public_html/wp-content/plugins/pattern-*/build --force
```

ryelle (commented):
Looking good so far. You might want to set the node version ([like this](https://github.com/WordPress/pattern-directory/blob/0eb786af0dff3c74266b8c19f6b60c55a1df13b0/.github/workflows/linters.yml#L31-L34)), to make sure it's built using the right tools.

> I'm not sure if we'd want to run the fetch-from-wp-json on every commit here though.

I think it's OK to keep that separate & run on a schedule.

> I wasn't quite sure what I was doing, and as a result, I had to git add .. --force some things that .gitignore ignores, 

I've encountered this on other projects, and I think `--force` really is the best way. 🤷🏻

dd32:
@ryelle Thanks for the review! I've updated all that now, it looks like it's working as it should now, and once #293 is merged all should be good with this one.

PR #291, opened 2021-07-15
Theme: Update data passed from the server to the app

ryelle:
This reorganizes the data passed into the JS app on global variables. All the URLs are now part of an object `wporgPatternsUrl`, the locale stays as `wporgLocale`, and a new variable `wporgPatternsData` is added, which passes in the current user ID. This replaces the user ID passed in on the single pattern container.

See https://github.com/WordPress/pattern-directory/pull/284#discussion_r670674587 for what prompted this. Technically, we can get the current user with `select( coreStore ).getCurrentUser()`, but there is no distinction between the loading state and logged out state. If we provide this value from the server, there is no loading state, so if the ID is 0, they're not logged in.

### Screenshots

There should be no visual changes.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -73,12 +73,30 @@ function enqueue_assets() {
 
 		wp_add_inline_script(
 			'wporg-pattern-script',
+			sprintf( "var wporgLocale = '%s';", wp_json_encode( get_locale() ) ),
+			'before'
+		);
+
+		wp_add_inline_script(
+			'wporg-pattern-script',
 			sprintf(
-				'var wporgAssetUrl = "%s", wporgSiteUrl = "%s", wporgLoginUrl = "%s", wporgLocale = \'%s\';',
-				esc_url( get_stylesheet_directory_uri() ),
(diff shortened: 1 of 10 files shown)


public_html/wp-content/themes/pattern-directory/functions.php
```diff
@@ -71,14 +71,32 @@ function enqueue_assets() {
...
+			'before'
+		);
+
+		wp_add_inline_script(
+			'wporg-pattern-script',
+			sprintf(
+				"var wporgPatternsData = JSON.parse( decodeURIComponent( '%s' ) )",
+				rawurlencode( wp_json_encode( array(
+					'userId' => get_current_user_id(),
+				) ) ),
```
coreymckrill, line 86: Just curious what the purpose of `rawurlencode` is on this one...

ryelle: I thought it was best practice for passing back arrays/objects from PHP to JS, but maybe that's an outdated practice?

coreymckrill: Oh it could be, though I haven't seen it before. But that's why I asked 😄

PR #294, opened 2021-07-15
Stats snapshots for patterns

### LINKED ISSUE #292: Directory: Create periodic snapshots of specific stats for tracking over time
Pulling this issue together from a few other places, mainly [this comment](https://github.com/WordPress/pattern-directory/issues/24#issuecomment-880756943).

The idea is to keep track of a few key stats (listed below) by generating a "snapshot" post containing the stats numbers at that particular moment, very similar to [how it's done](https://github.com/WordPress/five-for-the-future/blob/production/plugins/wporg-5ftf/includes/stats.php) on the 5ftf site.

**Open question**: should snapshots be generated daily, weekly, or monthly? The 5ftf stats are bi-weekly, but things are likely to change more often on the Pattern Directory, so daily might make the most sense here.

Initially, the stats to track are (list subject to change):

* Total number of published patterns
* Total number of unlisted patterns
* Total number of favorites
* Total number of flags, any status (reported patterns)
* Total number of pending flags
* Total number of resolved flags

We could also grab some other related stats, such as:

* Total number of patterns with at least one favorite
* Pattern(s) with the most favorites
* Total number of patterns with at least one flag, any status
* Total number of patterns with pending flags
* Total number of patterns that have reached the flag threshold and have been set to pending status for moderator review

Once frontend pattern creation launches, we might also want to see things like:

* Total number of published patterns for each category
* Total number of users who have at least one published pattern
* User(s) with the most published patterns
* User(s) with the most flagged patterns

coreymckrill:
Adds functions for gathering stats about patterns and saving them to snapshot posts, similar to what is done on the 5ftf site. A cron job is set up to run at 00:00 UTC daily to save a snapshot (don't worry it won't run on your wporg sandbox).

Also adds a simple, work-in-progress page in WP Admin to view the current stats and some basic info about the number of snapshots that have been created. In later PRs we can add things like exporting snapshots to CSV, showing stats on a public frontend page, and/or making pretty charts.

Fixes #292

### Screenshots

stats-page

1. This works best if you load it onto your wporg sandbox, so you can see the real numbers.
2. In WP Admin, find the Stats page under the Block Patterns menu.
3. Try creating, favoriting, or flagging a pattern and see if the numbers change.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/admin-stats.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/admin-stats.php
@@ -24,8 +24,8 @@ function add_subpage() {
 
 	add_submenu_page(
 		$parent_slug,
-		__( 'Pattern Stats', 'revisions-extended' ),
-		__( 'Stats', 'revisions-extended' ),
+		__( 'Pattern Stats', 'wporg-patterns' ),
+		__( 'Stats', 'wporg-patterns' ),
 		$post_type_object->cap->edit_posts,
 		PATTERN_POST_TYPE . '-stats',
 		__NAMESPACE__ . '\render_subpage'
@@ -60,7 +60,7 @@ function render_subpage() {
 				<?php foreach ( $schema['properties'] as $field_name => $field_schema ) : ?>
 					<tr>
 						<td>
-							<abbr title="<?php echo esc_attr( $field_schema['description'] ) ?>">
+							<abbr title="<?php echo esc_attr( $field_schema['description'] ); ?>">
 								<?php echo esc_html( $field_name ); ?>
 							</abbr>
 						</td>
--- a/public_html/wp-content/plugins/pattern-directory/includes/admin-stats.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/admin-stats.php
@@ -42,6 +42,7 @@ function render_subpage() {
 	$data           = get_snapshot_data();
 	$snapshot_query = get_snapshots( array(), true );
 	$snapshots      = $snapshot_query->get_posts();
+	$next_snapshot  = wp_get_scheduled_event( PATTERN_POST_TYPE . '_record_snapshot' );
 
 	?>
 	<style>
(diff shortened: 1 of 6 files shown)


coreymckrill:
^ Just added two more stats that aren't reflected in the screenshot. These split out originals and translations so you can get a better sense of how many unique patterns there actually are.

public_html/wp-content/plugins/pattern-directory/includes/admin-stats.php
```diff
@@ -0,0 +1,136 @@
...
+
+		<table class="widefat striped">
+			<tbody>
+				<?php foreach ( $schema['properties'] as $field_name => $field_schema ) : ?>
+					<tr>
+						<td>
+							<abbr title="<?php echo esc_attr( $field_schema['description'] ); ?>">
+								<?php echo esc_html( $field_name ); ?>
+							</abbr>
+						</td>
```
ryelle: Rather than having the descriptions on hover, could they be added as a middle column? I'm on a large monitor so there is definitely enough space, but I don't think we need to worry much about smaller screens unless it's unusable…
Screen Shot 2021-07-16 at 4 08 46 PM

public_html/wp-content/plugins/pattern-directory/includes/pattern-flag-post-type.php
```diff
@@ -145,3 +145,49 @@ function check_flag_threshold( $post_ID, $post, $update ) {
...
 }
+
+/**
+ * Get a list of post IDs for patterns that have pending flags.
+ *
+ * TODO this isn't used anywhere on the front end, but maybe it should be cached?
+ *
+ * @param array $args Optional. Query args. 'orderby' and/or 'order'.
+ *
+ * @return array
```
ryelle: (Just a suggestion) Do you know what type of items are in the array? I like the `string[]`/`integer[]` format for documenting that.

ryelle (approved):
This is good with or without my description suggestion, but I do think the plain-English is easier to understand than the keys.

Thanks for getting something up and running so quickly!

coreymckrill:
Good suggestions! Updated visuals on the stats table:

updated-stats-table

public_html/wp-content/plugins/pattern-directory/includes/pattern-flag-post-type.php
coreymckrill: Change in 6a12c28

public_html/wp-content/plugins/pattern-directory/includes/admin-stats.php
coreymckrill: b84b2c1

PR #296, opened 2021-07-16
Theme: Fix favorite button styles

StevenDufresne:
This fixes an issue seen on the live site. The favorite count button gets a `box-shadow` and a `background` color when `:active`.

### Screenshots
Remove the problem identified in the image by the red arrow.

Must be tested in a sandbox or on the live site.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern-grid.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern-grid.scss
@@ -107,12 +107,6 @@
 	text-decoration: none;
 }
 
-// Fixing WordPress.org collision.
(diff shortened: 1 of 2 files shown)


ryelle:
It looks like this happens only on the logged out view, because the `span` has the `button` class on it. I think it makes more sense to remove the button class from the span, rather than overriding the CSS.

StevenDufresne:
> It looks like this happens only on the logged out view, because the `span` has the `button` class on it. I think it makes more sense to remove the button class from the span, rather than overriding the CSS.

Yep, that works too. Updated.

ryelle (approved):
I tweaked it again to keep one single variable for all classes, in case we need to add/toggle more classes later.

PR #302, opened 2021-07-19
Theme: Increase enqueue action priority

StevenDufresne:
This PR increases the priority for `wp_enqueue_scripts` to make sure:

```
wp_enqueue_style(
  'wporg-style',
  get_theme_file_uri( '/css/style.css' ),
  array( 'dashicons', 'open-sans' ),
  filemtime( __DIR__ . '/css/style.css' )
);
```

is called from the child theme and not the parent. This approach is also consistent with other child themes ([wporg-plugins](https://github.com/WordPress/wordpress.org/blob/9ed80a9c96f4e66810e7a21a2e27b5ac7c3db4b7/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins/functions.php#L149)).

## Issue

Since this priority was set to `20`, the [wporg 'wporg-style' declaration](https://github.com/WordPress/wordpress.org/blob/9ed80a9c96f4e66810e7a21a2e27b5ac7c3db4b7/wordpress.org/public_html/wp-content/themes/pub/wporg/functions.php#L110) would run and apply the default cache value of `20200403`. Therefore on the live site, the cache was serving older assets... which is why some styles currently appear broken.

@dd32 Has suggested we update the `wporg` code to use `filemtime` as well, which I think is a good idea.

Note: This problem is only seen on the live site because it uses the `s.w.org` content delivery network.

- Open up developer tools (or page source)
- Click on Network
- Expect to find a Unix timestamp for the css/style.css file (Ie: pattern-directory/css/style.css?ver=1626659094).
- You should not see `pattern-directory/css/style.css?ver=20200403`

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -47,7 +47,7 @@ function enqueue_assets() {
 	$suffix       = $script_debug ? '' : '.min';
 
 	// Unregister the parent style so that the version & deps defined here are used.
-	wp_dequeue_style( 'wporg-style' );
+	wp_deregister_style( 'wporg-style' );
 	wp_enqueue_style(
 		'wporg-style',
 		get_theme_file_uri( '/css/style.css' ),
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -6,7 +6,7 @@ use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\POST_TYPE;
 use const WordPressdotorg\Pattern_Directory\Pattern_Flag_Post_Type\POST_TYPE as FLAG_POST_TYPE;
 
 add_action( 'after_setup_theme', __NAMESPACE__ . '\setup' );
-add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
+add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets', 20 );
 add_action( 'wp_head', __NAMESPACE__ . '\generate_block_editor_styles_html' );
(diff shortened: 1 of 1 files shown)


ryelle:
I tweaked this a little, since the priority was added to fix an issue on the logged out New Pattern page (see #249). Instead, I left the priority at 20, but dequeue the style first.

> @dd32 Has suggested we update the wporg code to use filemtime as well, which I think is a good idea.

Would we be able to do that from the parent theme? it would still be grabbing the `filemtime` of the parent theme's `style.css`, so still not accurate. IMO we probably don't need to enqueue a style from the parent at all, I'm pretty sure the child themes all redefine `wporg-style` (spot-checked learn & 5ftf). We just haven't run into this issue yet because they don't change the priority, and the child function runs first.

(We should probably take ^ to a meta ticket…)

StevenDufresne:
> I tweaked this a little, since the priority was added to fix an issue on the logged out New Pattern page (see #249). Instead, I left the priority at 20, but dequeue the style first.
> 
> > @dd32 Has suggested we update the wporg code to use filemtime as well, which I think is a good idea.
> 
> Would we be able to do that from the parent theme? it would still be grabbing the `filemtime` of the parent theme's `style.css`, so still not accurate. 

`get_theme_file_path( '/css/style.css' )` should retrieve the correct file but i don't mind just removing it either.

StevenDufresne:
@ryelle Were you able to get it to work by using `wp_dequeue_style`? I wasn't able to, most likely because of the priority of our `wp_enqueue_scripts`.

My local branch is still showing the following:
`http://localhost:8888/wp-content/themes/pattern-directory/css/style.css?ver=20200403`

ryelle:
Oh weird, I thought it worked but you're right - it should be using `wp_deregister_style`. Try now with the update.

StevenDufresne:
Yep, works as expected now.

ryelle (approved):

PR #303, opened 2021-07-19
Theme: Decode search term for language support 

### LINKED ISSUE #301: Search: Non-ASCII characters garbled
Maybe this is still being worked on, but when non-ASCII characters are used for search, they are displayed with character entity references. This is not the highest priority either – but would be nice if it's fixed at some point.

**To Reproduce**
Search for `画像` on any site (English or Rosetta site in any language). It means "image" or "images" in Japanese.

Depending on which site you are on, you get a different number of matches:

- https://wordpress.org/patterns/search/%E7%94%BB%E5%83%8F/ (76 matches)
- https://ja.wordpress.org/patterns/search/%E7%94%BB%E5%83%8F/ (74 matches)
- https://ar.wordpress.org/patterns/search/%E7%94%BB%E5%83%8F/ (18 matches)

Also, some of the matched patterns in Japanese may not be correct. This "[Three Buttons](https://ja.wordpress.org/patterns/pattern/three-buttons-ja/)" pattern is matched ([English version here](https://wordpress.org/patterns/pattern/three-buttons/)) but I don't think it includes the keyword (it has no image reference).

**Expected behavior**
See "画像" instead of "%E7%94%BB%E5%83%8F" as the searched keyword.
See "not found" result on sites other than ja.wordpress.org.

**Screenshots**
image

StevenDufresne:
This PR fixes an output issue with Non-ASCII search terms.

Fixes #301

### Screenshots

| Before | After |
| --- | --- |
|  |  |

Visit any of the following:
- `/search/%D8%B9%D9%8E%D8%B1%D9%8E/` (Arabic)
- `/search/%E7%94%BB%E5%83%8F` (Japanese)
- `/search/%EC%A1%B0%EC%84%A0%EB%A7%90/` (Korean)

Expect the search term to match the characters in the URL or the search form.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/src/components/category-context-bar/index.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/category-context-bar/index.js
@@ -69,7 +69,7 @@ function CategoryContextBar( { query } ) {
 			setMessage( getDefaultMessage( count, category.name ) );
 		}
 
-		const searchTerm = getSearchTermFromPath( path );
+		const searchTerm = decodeURI( getSearchTermFromPath( path ) );
 		if ( searchTerm.length > 0 ) {
 			setMessage( getSearchMessage( count, searchTerm ) );
 		}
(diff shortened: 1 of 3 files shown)


ryelle (approved):
Looks good, and looks like this also fixes the incorrect results mentioned in #301 👍🏻

PR #304, opened 2021-07-19
Pattern Translations: Retain original pattern's date in translated patterns

### LINKED ISSUE #299: Pattern Translations: Post date changes with every sync, creating lots of `_wp_old_date` postmeta rows
Every time [this function](https://github.com/WordPress/pattern-directory/blob/05c024ff4e138bca15138ee673c089a322cfac70/public_html/wp-content/plugins/pattern-translations/pattern-translations.php#L28) runs for a pattern post that already exists, `wp_insert_post` is setting the publish date to the current date/time, which in turn causes WP to [generate](https://github.com/WordPress/WordPress/blob/7bcca73ed7b1a910a996bdc3b601d1570ecd48f3/wp-includes/post.php#L6866) a `_wp_old_date` postmeta row for the post.

As of today there are nearly 15,000 `_wp_old_date` postmeta rows in the Pattern Directory site related to this.

Ideally, I think, `create_or_update_translated_pattern` would check if the post already exists and then if it does, run `wp_update_post` instead of `wp_insert_post`.

coreymckrill:
This addresses the issue described in #299 by including the original pattern's `post_date` value in the array that is passed to `wp_insert_post` when creating or updating a translated pattern.

This also adds a `wpop_is_translation` boolean postmeta value to the translated pattern post so that it's easier to differentiate it from other kinds of derived patterns that we may have in the future, like remixes.

Fixes #299

Props dd32, coreymckrill

### DIFF
--- a/public_html/wp-content/plugins/pattern-translations/pattern-translations.php
+++ b/public_html/wp-content/plugins/pattern-translations/pattern-translations.php
@@ -37,8 +37,9 @@ function create_or_update_translated_pattern( Pattern $pattern ) {
 		'post_author'  => $pattern->parent ? get_post( $pattern->parent->ID )->post_author : 0,
 		'post_status'  => $pattern->parent ? get_post( $pattern->parent->ID )->post_status : 'pending',
 		'meta_input'   => [
-			'wpop_description' => $pattern->description,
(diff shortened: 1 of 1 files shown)


dd32 (approved):
LGTM

PR #305, opened 2021-07-20
Patterns: Update URL + query logic for consistent navigation

### LINKED ISSUE #258: ?name=* queries do not 404
**Describe the bug**
When visiting a URL such as https://wordpress.org/patterns/?name=random-post-name or https://wordpress.org/patterns/random-post-name/ the request should 404, and use a 404 template.

Requests to 404 patterns do work though: https://wordpress.org/patterns/pattern/random-pattern-doesnt-exist/

The URI for patterns is currently https://wordpress.org/patterns**/pattern/**$pattern_name/, it might be worth removing that extra subdirectory as part of this to make queries simpler.

**To Reproduce**
1. Visit the quoted URIs above

**Expected behavior**
A 404 HTTP status and 404 template.

ryelle:
Fixes #258 by refactoring how the query object is generated from URLs/paths, and how URLs & context messages are generated from the query object. This collects logic around the app for getting filter view parameters into one place, which can convert to/from URLs & query objects.

Introduces `getQueryFromUrl` and `getUrlFromQuery`, so we're no longer getting the query values piecemeal with `getCategoryFromPath` etc. None of the `get*FromPath` functions should be used outside of the selector (the exception is the MyPatterns component). Also, we don't need to build up category paths anymore - we can use `getUrlFromQuery( { ...query, 'pattern-categories': id } ),` and it will build the correct URL, complete with current `orderby` or `author` properties.

Additionally, I refactored the `CategoryContextBar` to show the current author on author archives, plus sub-category navigation. It only shows the username for now, post-launch we can revisit to get the author's display name.

⚠️  I know this is a big change to how site nav works, and if it looks good I'd like to merge it ASAP - it would be great to have the author stuff working for launch. But let me know if anything doesn't look right, and I'll iterate.

### Screenshots

Screen Shot 2021-07-19 at 7 30 51 PM

Screen Shot 2021-07-19 at 7 20 07 PM

Screen Shot 2021-07-19 at 7 20 31 PM

1. I added tests for the new query selectors, those should pass.
2. Try clicking around the site- the patterns should load correctly for all filter views you pick

Some example paths:
- Click on a category, view page 2 or 3, clicking on an author in the category view should bring you to all patterns by that author
- View an author archive, view page 2 or 3, then click on a cateogry - it should show you page 1 of that author's patterns in the chosen category.


coreymckrill (commented):
👍 All works as described, tests pass

> Click on a category, view page 2 or 3, clicking on an author in the category view should bring you to all patterns by that author

Why not "all patterns _in the category_ by that author"? Seems like that would be more consistent with how the author archive works.

ryelle:
> Why not "all patterns in the category by that author"? Seems like that would be more consistent with how the author archive works.

The category nav works more like a filter, so it changes the current query (`/`, `/author/name`, etc). The links in the pattern thumbnails change the whole context - they load a new page, either the single pattern page or author archive. That's how I've been thinking about it anyway, but if that seems unexpected we could change it.

coreymckrill:
Yeah, I think that makes sense. Probably fine to leave it and see what other users think.

Issue #34, opened 2021-01-19
Pattern translations & i18n


ryelle:
Breaking out from #9: "Translations and i18n will need to be solved."

ryelle:
I have in my notes:

- Fork pattern into `[locale].wordpress.org/patterns`?
  - What about when the parent pattern updates?
- GlotPress probably makes more sense, but how should we mark what text needs translation?

ryelle:
A method for remixing patterns to translate content has been described in #224.

iandunn:
What's the difference between this and #224?

ryelle:
I guess it's not specifically, but I saw this as more of a tracking issue for i18n -- #224 has the "how to do it", but we haven't really discussed how to make localized patterns available to people browsing the directory.

PR #306, opened 2021-07-20
Translate term names, and load the correct locale for API & REST-API requests

dd32:
Translate term names, and load the correct locale for API & REST-API requests.

This PR is to ensure that Term names are properly localised in a variety of places:
 - https://ja.wordpress.org/patterns/ - The Term names in the menu should be localised. This is from the following REST API:
 - https://ja.wordpress.org/patterns/wp-json/wp/v2/pattern-categories?_locale=user Should return Japanese names (Note: Authenticated requests will return the users locale (Which we don't use on WordPress.org), unauthenticated will return Japanese)
 - https://api.wordpress.org/patterns/1.0/?locale=ja&categories - An API request made from an API client, should have the `locale` parameter respected

The translation method for the Term names is straight forward, setting the appropriate locale however is not.

The approach taken here to set the locale feels hacky to me, but I wasn't able to find a better way that reliably worked in all 4 use-cases and returned expected consistent results:
 - REST API, Authenticated, API client sending `_locale=user`
 - REST API, Authenticated, API client NOT setting a locale
 - REST API, Unauthenticated.
 - WordPress.org API requests setting the `locale` parameter.

Additionally, pages titles weren't translated:
 - https://ja.wordpress.org/patterns/my-favorites/ displays English text in the Page header, and in the Blue bar.

See #34, #223

1. Apply, visit urls metioned.

### DIFF
--- a/bin/i18n.php
+++ b/bin/i18n.php
@@ -106,6 +106,29 @@ function get_taxonomy_terms( $taxonomy ) {
 }
 
 /**
+ * Get data about pages from a REST API endpoint.
+ *
+ * @return array
+ */
+function get_pages() {
+	$endpoint = ENDPOINT_BASE . 'pages' . '?per_page=100';
+
+	$response = Requests::get( $endpoint );
+
+	if ( 200 !== $response->status_code ) {
+		die( 'Could not retrieve page list.' );
+	}
+
+	$pages = json_decode( $response->body, true );
+
+	if ( ! is_array( $pages ) ) {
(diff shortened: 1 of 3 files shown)


StevenDufresne:
This works on my machine! Tested with both `ja` and `ru`.

dd32:
@StevenDufresne pointed out that there's also some page titles that aren't translated, so I've added those to this PR too.

ryelle (approved):

PR #307, opened 2021-07-20
Prioritise results by requested locale, but return Locale and default (en_US) patterns.

dd32:
Currently the Pattern directory limits the results to the given locale, this is not ideal in all situations.

Consider:
 - https://ru.wordpress.org/patterns/ currently has no patterns, as none are translated
 - https://cs.wordpress.org/patterns/ currently only has a single partially translated pattern.
 - Results inside Gutenberg reflect the above.
 - https://api.wordpress.org/patterns/1.0/?locale=foo_bar returns no results. The `foo_bar` locale is admittedly not used on any sites, but it's not uncommon to see API requests with it set to `en` or `es` which are equally invalid `WP_Locale` values.

This PR changes the logic from only returning results in the matching locale, to returning in both the local and in en_US, with the localised results first.

No attempt at deduplication is done here, both the Translated and english results would be returned.
This also has the benefit that searching an English phrase would result in matching an english pattern, and not just returning an empty data set if it's not included in the locale translated patterns. Searching for a German phrase on an English or French site would however not return a matching German pattern.

See #244

1. Apply patch
2. Visit URLs mentioned. Perform searches.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -497,14 +497,34 @@ function filter_patterns_collection_params( $query_params ) {
 function filter_patterns_rest_query( $args, $request ) {
 	$locale = $request->get_param( 'locale' );
 
-	// Limit results to the requested locale.
-	// Workaround for https://core.trac.wordpress.org/ticket/47194.
+	// Prioritise results in the requested locale.
+	// Does not limit to only the requested locale, so as to provide results when no translations
+	// exist for the locale, or we do not recognise the locale.
 	if ( $locale ) {
-		$args['meta_query'][] = array(
+		$args['meta_query']['wpop_locale'] = array(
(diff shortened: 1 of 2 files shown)


dd32:
I also note that this introduces a query alteration which will require writing some manual ElasticSearch rules when time to implement that comes along.

iandunn:
>  it's not uncommon to see API requests with it set to `en` or `es` which are equally invalid WP_Locale values

Some recent WPCOM discussion around that is at `D64053-code`

iandunn:
> No attempt at deduplication is done here, both the Translated and english results would be returned.

That's not ideal for `en_*` locales, since e.g., the `en_AU` and `en_US` versions of `About me cards` will both appear.

https://en-au.wordpress.org/patterns/search/about/

We can probably de-dupe via `post_parent`, but that can be an iteration in another PR. I'll open an issue.

iandunn (approved):
LGTM. I'll go ahead and merge/deploy since it's early in Australia and this'll improve results for folks upgrading to 5.8.

PR #315, opened 2021-07-20
Pattern Translations: Copy viewport width from parent to translated post

coreymckrill:
Since only the text is changing, the viewport width value set on the original (if it exists) should also be used with the translations.

### DIFF
--- a/public_html/wp-content/plugins/pattern-translations/pattern-translations.php
+++ b/public_html/wp-content/plugins/pattern-translations/pattern-translations.php
@@ -32,21 +32,30 @@ if ( defined( 'WP_CLI' ) ) {
  * Creates or updates a localised pattern.
  */
 function create_or_update_translated_pattern( Pattern $pattern ) {
+	$parent = get_post( $pattern->parent->ID );
+
+	$meta_input = array(
+		'wpop_locale'         => $pattern->locale,
+		'wpop_is_translation' => true,
+	);
+	if ( $pattern->description ) {
+		$meta_input['wpop_description'] = $pattern->description;
+	}
+	if ( $parent->wpop_viewport_width ) {
+		$meta_input['wpop_viewport_width'] = $parent->wpop_viewport_width;
+	}
+
 	$args = [
 		'ID'           => $pattern->ID,
 		'post_type'    => POST_TYPE,
 		'post_title'   => $pattern->title,
 		'post_name'    => $pattern->ID ? $pattern->name : ( $pattern->name . '-' . $pattern->locale ), // TODO: Translate the slug?
-		'post_date'    => $pattern->parent ? get_post( $pattern->parent->ID )->post_date : '',
+		'post_date'    => $pattern->parent ? $parent->post_date : '',
 		'post_content' => $pattern->html,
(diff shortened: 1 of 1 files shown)


dd32 (approved):
LGTM

public_html/wp-content/plugins/pattern-translations/pattern-translations.php
```diff
@@ -32,21 +32,24 @@
  * Creates or updates a localised pattern.
  */
 function create_or_update_translated_pattern( Pattern $pattern ) {
+	$parent = get_post( $pattern->parent->ID );
```
dd32: This will throw a notice if the parent isn't available, but this function probably shouldn't be called with that input anyway.

public_html/wp-content/plugins/pattern-translations/pattern-translations.php
```diff
@@ -32,21 +32,24 @@
...
 		'post_title'   => $pattern->title,
 		'post_name'    => $pattern->ID ? $pattern->name : ( $pattern->name . '-' . $pattern->locale ), // TODO: Translate the slug?
-		'post_date'    => $pattern->parent ? get_post( $pattern->parent->ID )->post_date : '',
+		'post_date'    => $pattern->parent ? $parent->post_date : '',
 		'post_content' => $pattern->html,
 		'post_parent'  => $pattern->parent ? $pattern->parent->ID : 0,
-		'post_author'  => $pattern->parent ? get_post( $pattern->parent->ID )->post_author : 0,
-		'post_status'  => $pattern->parent ? get_post( $pattern->parent->ID )->post_status : 'pending',
-		'meta_input'   => [
+		'post_author'  => $pattern->parent ? $parent->post_author : 0,
```
dd32: I'm tempted to suggest we could just change these to `$parent->post_author ?? 0` rather than the ` a ? b : c` operator.

public_html/wp-content/plugins/pattern-translations/pattern-translations.php
```diff
@@ -32,21 +32,24 @@
...
+		'post_status'  => $pattern->parent ? $parent->post_status : 'pending',
+		'meta_input'   => array(
```
dd32: If updating this to `array()` the entire block should be changed to `array()` rather than using `[]`.

I used `[]` as a) it's my personal preference, and b) All of the library code that this plugin pulled in used that syntax.

public_html/wp-content/plugins/pattern-translations/pattern-translations.php
coreymckrill: Ah good call. Changed in 880ea43

public_html/wp-content/plugins/pattern-translations/pattern-translations.php
coreymckrill: Good idea. Changed in 880ea43

public_html/wp-content/plugins/pattern-translations/pattern-translations.php
coreymckrill: Bah, that was a copy pasta mistake. Did not intend to change the syntax style. Restored in 9f7bf9a

PR #322, opened 2021-07-22
Theme: Space out context bar on search page.

StevenDufresne:
This PR updates the pattern search in order to add padding to push the context category bar down.

### Screenshots

| Before | After |
| --- | --- |
|  |  |

1. Search for a pattern on `/`
2. Expect to see spacing between the breadcrumbs and the category context bar.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_context-bar.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_context-bar.scss
@@ -1,5 +1,5 @@
 .context-bar {
-	margin: $gutter-default $gutter-default 0;
+	margin: 0 $gutter-default;
 	background: $color-gray-light-300;
 	border-radius: 2px;
 	font-size: 0.8125rem;
@@ -12,8 +12,12 @@
 		justify-content: space-between;
 	}
 
+	@media only screen and ( min-width: $breakpoint-medium ) {
+		margin: 0 $gutter-default;
+	}
+
 	@media only screen and ( min-width: $breakpoint-large ) {
-		margin: $gutter-default auto;
+		margin: 0 auto;
 		max-width: $size__site-main;
 	}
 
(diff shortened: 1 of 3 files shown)


public_html/wp-content/themes/pattern-directory/css/components/_patterns-search.scss
```diff
@@ -0,0 +1,3 @@
+.patterns-search__container {
+	margin-top: 1.5rem;
```
ryelle: You should be able to just add this margin to `.context-bar`, and margin collapsing means it'll still be fine when there's a menu.

StevenDufresne: Yep. I tend to lean towards margin on containers and not components. I don't know how many times I've needed to overwrite the margin-bottom on the[ Gutenberg base control](https://github.com/WordPress/gutenberg/blob/ece5560abd945279530bd58f80f74f06b904a293/packages/components/src/base-control/styles/base-control-styles.js#L18) already :)

ryelle: Technically we want the margin on the component though, since ContextBar will always need space above it (whether it's between the menu & context bar or just spacing above the context bar).

If you'd rather put space on the container, can you put space on `site-main`, or give all the wrappers a new class? Really I just don't want the search page to have a special new style, when this does apply everywhere.

StevenDufresne: The easiest and cleanest way I could find apart from adding classes to the main container is to target the specific page: `.search .site-header`.

Adding it to the component is tricky for 2 reasons:
1. The mobile layouts differ and there would need context-specific adding/removing to get it to work which would then again be specific to the page, or specific to the lack of menu.
2. The component shrinks its height to 0 to account for the transition, adding margin disturbs the layout when it's not active

Also, adding it to the `site-header` for non-homepage also doesn't work because the mobile version of the category navigation requires that there is no margin-bottom on the `site-header`, but it is required for the search page. So you need to conditionally add the margin which is again, not ideal.

ryelle:
@StevenDufresne I tried my approach in https://github.com/WordPress/pattern-directory/pull/338, which also fixed an issue on the homepage - it was a little more involved, but I think it's a more stable fix for the spacing issues.

ryelle:
Fixed with #338

PR #323, opened 2021-07-22
Pattern Stats: Add CSV export for snapshots

coreymckrill:
Adds a form on the WP Admin stats screen for exporting a date range of stats snapshots to a CSV file.

Brings over the `Export_CSV` class from the WordCamp codebase.

### Screenshots

snapshot-export

1. Load this PR onto your sandbox so you can try exporting real data.
2. Go to the Stats screen, select a date range, and click the Export button.
3. Check the resulting spreadsheet to make sure the data looks legit.
4. See if you can get the form to spit out an `error.txt` file instead of a `.csv` file 😄


public_html/wp-content/plugins/pattern-directory/includes/class-export-csv.php
```diff
@@ -0,0 +1,341 @@
+<?php
+
+namespace WordCamp\Utilities;
```
ryelle, line 3: I'm assuming this is intentionally not renamed to `WordPressdotorg\Pattern_Directory\…` since it's a cloned file, is that right?

public_html/wp-content/plugins/pattern-directory/includes/stats.php
```diff
@@ -58,85 +58,71 @@ function get_meta_field_schema() {
 			'count-patterns'                 => array(
 				'description' => __( 'The total number of pattern posts.', 'wporg-patterns' ),
 				'type'        => 'integer',
-				'default'     => 0,
```
ryelle, line 61: Why remove the defaults?

ryelle (approved):
Works great, I was able to make a quick chart of favorites activity with the export i downloaded 💯  

Screen Shot 2021-07-26 at 12 49 07 PM

I changed the min date on the inputs in inspector to get the error file, and tried to access the page as a lower-capability user & couldn't get in. It all looks good to me 👍🏻

public_html/wp-content/plugins/pattern-directory/includes/class-export-csv.php
coreymckrill: Yep, trying to keep the file exactly the same in both places. We also do it with the meetup api client file. Ideally we'd move files like these into a public directory that's shared by both WordCamp and WordPress.org, but that's a whole separate can of worms 😅

public_html/wp-content/plugins/pattern-directory/includes/stats.php
coreymckrill: If the stat schema changes at some point, and there are new stats that we didn't capture in previous snapshots, having defaults could be misleading. When we export a snapshot that's missing a field from the schema, and that field has a defined default, it will appear that the snapshot just has the default value. But if no default is defined, that snapshot will just have an empty value for that field.

ryelle: Ah, that makes sense. thanks for explaining 👍🏻

PR #337, opened 2021-07-26
Theme: Update the search form styles to match new designs.

### LINKED ISSUE #328: Search box focus styling
When the search box has `:focus` it's border is hard to see. Here's how it looks on the homepage and on an interior page:

image

image

I'd expect the focus border to stand out more from the blue background and to not bisect the input and submit button.

image

Adding a darker (#0A4B78), single-pixel border between the input shape and the focus outline helps along with using a more contrasting color (#72AEE6) for the focus outline.

StevenDufresne:
This PR makes the search form look like the screenshots on #328.

I shared the input style for the button. Thoughts?

Fixes #328

### Screenshots
|  Before | After  |
| --- | ---|
|  |  |
|   |  |

1. Visit `/`.
2. Add focus to the search form.
3. Expect it to look like designs.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern-search.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern-search.scss
@@ -56,7 +56,6 @@
 	}
 
 	&:focus {
-
 		@include focus_styles;
 	}
 }
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern-search.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern-search.scss
@@ -14,7 +14,7 @@
 	input[type="search"] {
 		flex-grow: 1;
 		margin: 0;
-		padding: 0.5rem;
+		padding: 0.5rem 2rem 0.5rem 0.5rem;
 		border: 1px solid transparent;
 		border-radius: 2px;
 		box-shadow: none;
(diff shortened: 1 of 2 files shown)


ryelle:
The search runs into the icon now:

Screen Shot 2021-07-26 at 2 23 12 PM

Maybe using `:focus-within` would help?

public_html/wp-content/themes/pattern-directory/css/components/_pattern-search.scss
```diff
@@ -1,20 +1,28 @@
+@mixin focus_styles {
```
ryelle, line 1: If we do want this to be a consistent focus style, maybe this mixin should be in `settings/`, so it can be used across the site? Or is this style meant just for the search?

StevenDufresne: I've only seen it on search so I was hesitant for it to be shared. The styling seems to be designed specifically for display on a blue background but would probably also work everywhere. ❓

StevenDufresne:
> The search runs into the icon now:
> 
> Screen Shot 2021-07-26 at 2 23 12 PM
> 
> Maybe using `:focus-within` would help?

Oh, that's neat. I missed the birth of `:focus-within`, good to know! 

I think just adding padding to the right, keeping it `position: absolute`, will make the styles a bit more portable, avoiding potential line height issues down the road.

Updated 303392d.

public_html/wp-content/themes/pattern-directory/css/components/_pattern-search.scss
ryelle: I should have directed that question at @shaunandrews 🙂  Is this focus style just for the header search, or would it be repeated anywhere else on the site?

ryelle (approved):
Looks good 👍🏻

PR #338, opened 2021-07-26
Navigation Menu: Fix spacing around menu and message

ryelle:
Alternative to #322 - In trunk, the context bar is pushed against the header on search. There is also a double-margin on the homepage (when no message is displayed). This PR fixes both issues by tweaking some margins on the Navigation Layout & ContextBar, and removing the ContextBar when the message is empty (if/when we bring back context links we'll need to update this logic).

### Screenshots

Double-margin below the menu:

| Before | After |
|------|------|
| Screen Shot 2021-07-26 at 12 26 14 PM | Screen Shot 2021-07-26 at 12 21 09 PM |

No space on search page:

| Before | After |
|------|------|
| Screen Shot 2021-07-26 at 12 27 39 PM | Screen Shot 2021-07-26 at 12 27 52 PM |

Margins are correct on small screens too
Screen Shot 2021-07-26 at 12 21 45 PM

Screen Shot 2021-07-26 at 12 21 53 PM

1. On the homepage, `/`, there should be the same amount of space above and below the menu.
1. Search for a pattern
2. There should be space above the grey "patterns found" message
3. Try on different screen sizes, the spacing should remain consistent.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/src/components/context-bar/index.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/context-bar/index.js
@@ -8,7 +8,7 @@ import useDeepCompareEffect from 'use-deep-compare-effect';
  * WordPress dependencies
  */
 import { Spinner } from '@wordpress/components';
-import { useRef, useState } from '@wordpress/element';
+import { useState } from '@wordpress/element';
 import { useSelect } from '@wordpress/data';
 
 /**
@@ -36,7 +36,6 @@ function ContextBar( props ) {
 		title: '',
 		links: [],
 	} );
-	const innerRef = useRef( null );
 
 	const { author, category, count, isLoadingPatterns, query } = useSelect(
 		( select ) => {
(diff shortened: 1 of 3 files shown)


public_html/wp-content/themes/pattern-directory/src/components/context-bar/index.js
```diff
@@ -73,18 +72,13 @@ function ContextBar( props ) {
...
 	const classes = classnames( {
 		'context-bar__spinner': true,
 		'context-bar__spinner--is-hidden': ! isLoadingPatterns,
 	} );
 
-	return (
-		<header className="context-bar" style={ { height: `${ height }px` } }>
+	return ! message ? null : (
+		<header className="context-bar">
 			<div ref={ innerRef }>
```
StevenDufresne: We'll want to get rid of this `ref`, it's no longer needed.

I think that `div` could also be removed and the styles applied to the header, but I'm not 100% on that.

StevenDufresne (approved):
If we want to do away with the animation, then yes this change works well.

PR #342, opened 2021-07-28
Pattern preview: Show scaled down large-screen preview on smaller screens

### LINKED ISSUE #122: Single Patterns: Live resizable preview
Moving some conversation from #121 here so the framework there can be merged. This is all about the resizable preview.

First, which handle design should we use?

Screen Shot 2021-06-04 at 2 59 32 PM Screen Shot 2021-06-04 at 2 59 42 PM

I made the assumption that we shouldn't show the size dropdown at small screens since there weren't size levels to resize to, but the handles are still available. But now I think we should drop the handles at the 480px breakpoint, keep all sizes in dropdown, and scale down the preview to fit the screen. So it will be tiny if someone on a 400px screen tries to preview at 1200 wide, but that's what they should expect.

@shaunandrews, does that match what you expect to happen on smaller screens?

ryelle:
Fixes #122. Currently, you can only preview a pattern at your native viewport width or smaller. If you're on a phone, you can't see what the pattern would look like on a monitor. This PR updates the logic for the dropdown and drag handles, and scales down the pattern preview so it's visible on any screen size.

Now, the size dropdown is always visible, while the drag handles disappear below 480px.

⚠️  This is still in progress, but I'm having trouble with it. @StevenDufresne maybe you can help? Feel free to comment or commit.

I've swapped out the forked `iframe` component to use [the `SandBox` component](https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/sandbox) from `@wordpress/components`. This seems more in line with our intention, and could replace our forked component (maybe). It loads in the given HTML, then resets the height and width on the `iframe` so that the content is fully visible. This is great normally, because we want the whole pattern to be visible.

To make the scaling work, I'm taking the requested width (full 1200, default 960, etc) and the container width (basically viewport width), and dividing container / request, so if you request 1200px preview on a 600px phone, it should scale by 0.5. Then I pass that to the CSS via custom properties (so that we can override the set width on the iframe, since we can't pass down those styles).

This is where the problem is. The width is correctly scaled down, so ex, in "full 1200 preview" on a 600px screen, a pattern is rendered at 1200w x 800h, the iframe scales down 0.5. A max-width on the container constrains it to 600px width, but the height is still 800px, so there's a huge empty space below the preview. It seems like the container still expects the content to be 1200x800, and the max-width prevents the horizontal overflow, but nothing catches the height.

### Screenshots

Looks great when the preview size is smaller the the viewport!

Screen Shot 2021-07-28 at 17 16 36

Screen Shot 2021-07-28 at 17 16 48

1. Build the branch
2. View any single pattern at various screen sizes
3. Try previewing it at different widths using the dropdown
4. The preview should show the whole pattern, but not have extra space around it

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern-preview.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern-preview.scss
@@ -20,11 +20,11 @@
 	max-width: 100vw;
 	min-width: 320px;
 
-	.pattern-preview__viewport-iframe {
+	.components-sandbox {
 		background: $color-white;
-		border: 1px solid $color-gray-light-400;
+		border: 0;
+		box-shadow: 0 0 0 1px $color-gray-light-400;
 		vertical-align: middle;
-		max-width: 100vw;
 	}
 
 	&:focus-within .pattern-preview__resize-help {
@@ -90,5 +90,17 @@
 
 	&:focus {
 		box-shadow: 0 1px 0 #0073aa, 0 0 2px 1px #33b3db;
+	}
+}
+
+.pattern-preview__canvas {
+	margin: 0 auto;
+	overflow: hidden;
+
+	.components-sandbox {
+		width: var(--wporg-pattern-preview--width) !important;
+		max-width: none !important;
+		transform: scale(var(--wporg-pattern-preview--scale)) !important;
+		transform-origin: top left;
 	}
 }
--- a/public_html/wp-content/themes/pattern-directory/src/components/pattern-preview/canvas.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/pattern-preview/canvas.js
@@ -1,29 +1,64 @@
 /**
- * Internal dependencies
+ * WordPress dependencies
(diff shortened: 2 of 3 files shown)


StevenDufresne:
Hm... that's a tricky problem to solve using `transforms` and it appears like we don't have direct control of the `iframes` height (not 100% on that). You would probably have more success manually implementing the scale using width and height.

To be honest, when viewing different device previews it's pretty common to fix the dimensions and mimic the device as opposed to using drag handles. Are we overcomplicating this? What about fixing the height/width and then making the pre-determined viewports responsive?

ryelle:
You can ignore the drag handles, they're not relevant to the issue. You're right that we don't have control of the iframe's height, it auto-expands to fit the content (which is what we want it to do). We're not doing "device previews", just a few set widths, so there is no height to constrain it to.

If we absolutely have to, we could go back to a fixed height for the container, but I was hoping we could come up with another solution.

ryelle:
Closing in favor of a solution using #380.

PR #344, opened 2021-07-29
Patterns by Author: Show the author's display name instead of username.

### LINKED ISSUE #311: Pattern Author Archive: Show user's display name
Currently, when on an archive view, the header says "Patterns by **authorname**", where authorname is the value from the URL (username). Instead, this should be "Patterns by **WordPress.org**", using the user's display name.

Screen Shot 2021-07-20 at 3 24 41 PM

ryelle:
Fixes #311. Currently, the author archive breadcrumb and context bar use the author's username (ex, `wordpressdotorg`), but it would be better to show the display name. This PR injects the current author name into the `wporgPatternsData` global, setting it when on the author page. This avoids needing to create a separate endpoint for user information, since moving to an author archive will always load a new page from the server.

If/when we move to more SPA-style control, we can revisit this.

### Screenshots

| Before | After |
|--------|------|
| Screen Shot 2021-07-29 at 4 07 55 PM | Screen Shot 2021-07-29 at 3 47 49 PM |

1. Set up some patterns with different authors
2. Set the author display names to something custom, optionally with non-english characters
3. View an author archive
4. The blue bar should read "Author: Display Name"
5. The grey bar should read "X patterns by Display Name", or "X [category] patterns by Display Name"

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -85,6 +85,7 @@ function enqueue_assets() {
 				"var wporgPatternsData = JSON.parse( decodeURIComponent( '%s' ) )",
 				rawurlencode( wp_json_encode( array(
 					'userId' => get_current_user_id(),
+					'currentAuthorName' => esc_html( get_the_author_meta( 'display_name' ) ),
 				) ) ),
 			),
 			'before'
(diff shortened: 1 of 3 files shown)


StevenDufresne (approved):
I'm fine with this!

One thing to mention, on the 'all' view the author is bold, on a specific category the category is bold and the author is not. Not sure if that was the intention... just pointing it out.

ryelle:
> One thing to mention, on the 'all' view the author is bold, on a specific category the category is bold and the author is not. Not sure if that was the intention... just pointing it out.

That's intentional - there was no design for that case, and it seemed like it was competing for attention to have both bold. I figured you already know you're on the author page from the breadcrumb, so we can highlight the category.

PR #346, opened 2021-08-04
Theme: Only load iframe assets when pattern thumbnail is in view.

StevenDufresne:
Loading pattern thumbnails is expensive on page load. This PR implements lazy loading of iframe assets only when the pattern is visible in the viewport.

See #255

### Benchmark
Environment: 
- Chrome Lighthouse (Mobile)
- Localhost
- Incognito mode.

 | Vital | Improvement |
| --- | --- |
 |Time to Interact  | 32%  | 
| First Contentful Paint | 0 |  
| Speed Index |  12% |
| Largest Contentful Paint |  3% |
| Total Blocking Time | 34% |

_Based on the average of 3 test runs before & after._

### Screenshots
[Screencast](https://d.pr/i/63mhH3)

1. Visit `/`
2. As you scroll down, the visible pattern thumbnails should load their assets.
3. Click different categories and expect all the assets for pattern thumbnails to be loaded.


ryelle:
I don't see as much of a consistent improvement in Lighthouse, but this does reduce the initial page load for most browsers (assuming most will see the first two rows when they load the page). I saw about 25-30% smaller download size while testing on desktop sizes, and up to 70% on a small screen. Further work could be reducing the image sizes more, I opened #349 with one idea.

ryelle (approved):

PR #357, opened 2021-09-01
Fix root README path in Pattern Creator README

zackkrida:
Tiny path typo :)


ryelle (approved):
Thanks for the 🗑️ pick-up!

PR #361, opened 2021-09-30
Fix typo on empty Favorites screen.

shaunandrews:
image

The empty favorites screen has copy that mentions the ability to "make it as a favorite." This is a typo; It should say "mark it as a favorite." This originated from my Figma file, so kudos for the dedication to pixel perfection. ;)


ryelle (approved):
Hah, blame "copy as text" 🙂  Thanks for the fix!

PR #373, opened 2021-10-28
Pattern Creation: Prevent uploads to the media library

### LINKED ISSUE #37: Create a library of media and only allow that to be used in patterns
We have #23 for the idea of setting up asset placeholders (cover block, images, etc) where the end user would add in their own image, but it's also likely that pattern authors will want to add their own assets as part of the pattern (a cover block with a highly designed background image).

- Should we allow asset (image, video, audio, …?) uploads to w.org through the creator?
- Maybe only allow the "Insert from URL", with users uploading media elsewhere?
- Safelist the allowed URLs?
- What if the content at the URL changes/disappears?

Checking for inappropriate material & licensing will be tracked in #22, but could also be part of the upload step here

ryelle:
Fixes #37 — ~There are over 100 images uploaded to the Pattern Directory, these have been verified CC0 and can be used in patterns. This PR removes the ability for users to upload to the site, but leaves the ability to open the media library to insert already-uploaded images.~

This PR is laying the groundwork for #375. It disables uploads for non-admins, and sets up a button to open the media modal. In #375, this will be used to open our Openverse modal.

⚠️  This only applies to non-admins, admin & superadmin users can still upload media so that we can manage the media library, in case it's necessary.

❓ Specifically requesting reviews because I'm touching capabilities, if there's a better way to achieve this let me know.

### Screenshots

When media blocks are inserted, they now just have "Pattern Media Library" as an option, for example the Gallery block:

custom-gallery

Media & Text:

Screen Shot 2021-10-28 at 2 20 41 PM

In the Media & Text block, only "Open Media Library" is available to replace the image.

custom-replace-flow

The Media Library itself still has the "Upload files" tab, but if you try to upload you get a permissions error:

Screen Shot 2021-10-28 at 2 21 51 PM

1. Try as various user levels (still need to be a user on the site to create patterns)
2. Create a new pattern or edit an existing one
3. Add any media block: Image, Gallery, Video, etc
4. You should **not** be able to insert via URL
5. You should **not** be able to upload your own file
6. You **should** be able to insert existing media files into the pattern
7. Play around with the pattern & blocks more - saving, editing, etc should all still work

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/package.json
+++ b/public_html/wp-content/plugins/pattern-creator/package.json
@@ -35,6 +35,7 @@
 		"@wordpress/editor": "12.0.6",
 		"@wordpress/element": "4.0.4",
 		"@wordpress/format-library": "3.0.9",
+		"@wordpress/hooks": "3.2.2",
 		"@wordpress/i18n": "4.2.4",
 		"@wordpress/icons": "6.1.1",
 		"@wordpress/interface": "4.1.5",
--- a/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
+++ b/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
@@ -197,3 +197,17 @@ function show_admin_bar( $show_admin_bar ) {
 }
 // Priority needs to be over 1000 to override `logged-out-admin-bar`.
 add_filter( 'show_admin_bar', __NAMESPACE__ . '\show_admin_bar', 1001 );
+
+/**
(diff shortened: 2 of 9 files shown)


shaunandrews:
Could we use the primary button styling and update the language to "Choose {mediaType}" or, if not possible (or specified), just "Media Library"?

image

ryelle:
Updated to use the block label/media type for the title, and primary button style for the button.

Screen Shot 2021-11-11 at 1 39 53 PM

PR #382, opened 2021-11-24
Openverse: Add endpoint to proxy search requests with authentication

ryelle:
See #375 — This PR adds an API endpoint to proxy the Openverse requests through WP, so that we can add authentication & some light caching. The OV API uses OAuth to bypass rate limiting, so we need to be able to generate and use an oauth token when making requests. Additionally, the requests are cached for an hour, to prevent excessive use of the OV API & make things a little faster for folks making repeated searches.

This will need two new secrets for the consumer ID & secret.

I use [Insomnia](https://insomnia.rest/) to test API changes, but you can also request via curl.

Get the secrets, search for "Pattern Directory Openverse Oauth2". Set these in a new file `mu-plugins/1-sandbox.php` (since 0-sandbox is tracked in git).

```php
<?php

define( 'PATTERN_OV_OAUTH_ID', '…' );
define( 'PATTERN_OV_OAUTH_SECRET', '…' );
```

1. Set up an application password for your user to use with Basic Auth
2. Create a GET request to `http://localhost:8888/wp-json/wporg/v1/openverse/search`, with the parameter `search` & any value you want to search for
3. It should succeed. The response is an array of `{ title, url, thumbnail }`, and you should see two headers `X-WP-Total`, `X-WP-TotalPages`.
4. You can check that it created an auth token by running `yarn wp-env run cli "wp option get ov-oauth-token"`
5. Run the request again and it should return faster (locally cached now)
6. Try different requests, edge cases, etc

Here's a command for curl, you'll just want to add your auth info (or change `get_items_permissions_check`):

```bash
curl --request GET \
  --url 'http://localhost:8888/wp-json/wporg/v1/openverse/search?search=Blue%20Sky' \
  --header 'Authorization: Basic [user:app password]'
```

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
+++ b/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
@@ -86,9 +86,27 @@ class Openverse_Client {
 	 */
 	public function get_oauth_token() {
 		$token = get_option( self::TOKEN_OPTION_KEY, array() );
-		if ( $this->is_valid_token( $token ) ) {
+		$token_life = $this->is_valid_token( $token );
+		if ( $token_life >= 5 * MINUTE_IN_SECONDS ) {
 			return $token['access_token'];
 		}
+		
+		$being_refreshed = get_option( self::TOKEN_OPTION_KEY . '_refresh' );
+		if ( $being_refreshed > time() - MINUTE_IN_SECONDS ) {
+			// Return stale token if possible.
+			if ( $token_life ) {
+				return $token['access_token'];
+			}
+
+			// Token has expired, and we're still refreshing the token, which means it's failed multiple times in a row.
+			return new WP_Error(
+				'invalid-ov-token-refresh',
+				'The token refresh process is locked, and we have no almost-expiring tokens.'
+			);
+		}
+
+		// Lock token refresh
+		update_option( self::TOKEN_OPTION_KEY . '_refresh', time() );
 
 		$response = wp_remote_post( $this->url . '/v1/auth_tokens/token/', array(
 			'body' => array(
--- a/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
+++ b/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
@@ -123,7 +123,7 @@ class Openverse_Client {
 		if ( 200 !== $response_code ) {
 			return new WP_Error(
 				'invalid-ov-token-request',
-				sprintf( __( 'The token generation request failed with a %s error.', 'wporg-patterns' ), $response_code )
+				sprintf( 'The token generation request failed with a %s error.', $response_code )
 			);
 		}
 
--- a/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
+++ b/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
@@ -137,9 +137,10 @@ class Openverse_Client {
 			return new WP_Error( 'invalid-ov-token-response', __( 'The token generation response is malformed.', 'wporg-patterns' ) );
 		}
 
-		// Invalidate the token 5 minutes early, just in case.
-		$token['expires_at'] = time() + $token['expires_in'] - 5 * MINUTE_IN_SECONDS;
+		// Invalidate the token a minute early, ensures that any requests around the "expiry" succeed as long as the request is made within a minute of the token validation being run.
+		$token['expires_at'] = time() + $token['expires_in'] - MINUTE_IN_SECONDS;
 		update_option( self::TOKEN_OPTION_KEY, $token );
+		delete_option( self::TOKEN_OPTION_KEY . '_refresh' );
 
 		return $token['access_token'];
 	}
--- a/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
+++ b/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
@@ -216,7 +216,21 @@ class Openverse_Client {
 		if ( false === $results ) {
 			$results = $this->fetch_results();
 			if ( is_wp_error( $results ) ) {
-				return $results;
+				// During development, return the upstream error.
+				if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+					return $results;
+				}
+			
+				// Log the error
+				trigger_error( $results->get_error_code(), E_USER_WARNING );
+
+				// Set a short timeout to avoid hammering the API during outages.
+				set_transient( $cache_key, [], 0.5 * MINUTE_IN_SECONDS );
+		
+				return new WP_Error(
+					'search-request-failed',
+					__( 'The search API provider is currently experiencing an error. Try again in a few seconds', 'wporg-patterns' ) // Publicly facing human-readable error.
+				);
 			}
 			set_transient( $cache_key, $results, $ttl );
 		}
--- a/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
+++ b/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
@@ -76,7 +76,7 @@ class Openverse_Client {
 		if ( time() >= $token['expires_at'] ) {
 			return false;
 		}
-		return true;
+		return time() - $token['expires_at'];
 	}
 
 	/**
--- a/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
+++ b/public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
@@ -1,0 +1,207 @@
+<?php
+/**
+ * Class Openverse_Client
+ *
+ * @package WordPressdotorg\Pattern_Creator
+ */
+class Openverse_Client {
+	/**
+	 * @var string The option key where the cache is saved in the database.
+	 */
+	const CACHE_KEY = 'openverse_search';
+
+	/**
+	 * @var string The option key where the cache is saved in the database.
+	 */
+	const TOKEN_OPTION_KEY = 'ov-oauth-token';
+
+	/**
+	 * @var array<string, string> The set of parameters for this request.
+	 */
+	protected $params = array();
+
+	/**
+	 * @var string The base URL for requests.
+	 */
+	protected $url = 'https://api.openverse.engineering';
+
+	/**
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
```diff
@@ -0,0 +1,207 @@
...
+			return $this->params[ $key ];
+		}
+		return $default;
+	}
+
+	/**
+	 * Check if token exists and has not expired.
+	 *
+	 * @param array $token Token object.
+	 * @return bool
```
dd32:
```suggestion
	 * @return bool|int False upon failure, seconds remaining for token on success.
```

public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
```diff
@@ -0,0 +1,207 @@
...
+		if ( ! isset( $token['access_token'] ) || ! $token['access_token'] ) {
+			return false;
+		}
+		if ( ! isset( $token['expires_at'] ) ) {
+			return false;
+		}
+		if ( time() >= $token['expires_at'] ) {
+			return false;
+		}
+		return true;
```
dd32:
```suggestion
		return time() - $token['expires_at'];
```

public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
```diff
@@ -0,0 +1,207 @@
...
+	/**
+	 * Get (or generate) a valid oauth token.
+	 *
+	 * @return string|WP_Error A valid token or WP_Error if there was an error.
+	 */
+	public function get_oauth_token() {
+		$token = get_option( self::TOKEN_OPTION_KEY, array() );
+		if ( $this->is_valid_token( $token ) ) {
+			return $token['access_token'];
+		}
```
dd32:
```suggestion
		$token_life = $this->is_valid_token( $token );
		if ( $token_life >= 5 * MINUTE_IN_SECONDS ) {
			return $token['access_token'];
		}
		
		$being_refreshed = get_option( self::TOKEN_OPTION_KEY . '_refresh' );
		if ( $being_refreshed > time() - MINUTE_IN_SECONDS ) {
			// Return stale token if possible.
			if ( $token_life ) {
				return $token['access_token'];
			}

			// Token has expired, and we're still refreshing the token, which means it's failed multiple times in a row.
			return new WP_Error(
				'invalid-ov-token-refresh',
				'The token refresh process is locked, and we have no almost-expiring tokens.'
			);
		}

		// Lock token refresh
		update_option( self::TOKEN_OPTION_KEY . '_refresh', time() );
```

public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
```diff
@@ -0,0 +1,207 @@
...
+
+		$token = json_decode( $body, true ); // Second param returns decoded value as assoc array.
+		if ( null === $token ) {
+			return new WP_Error( 'invalid-ov-token-response', __( 'The token generation response is malformed.', 'wporg-patterns' ) );
+		}
+
+		// Invalidate the token 5 minutes early, just in case.
+		$token['expires_at'] = time() + $token['expires_in'] - 5 * MINUTE_IN_SECONDS;
+		update_option( self::TOKEN_OPTION_KEY, $token );
```
dd32:
```suggestion
		// Invalidate the token a minute early, ensures that any requests around the "expiry" succeed as long as the request is made within a minute of the token validation being run.
		$token['expires_at'] = time() + $token['expires_in'] - MINUTE_IN_SECONDS;
		update_option( self::TOKEN_OPTION_KEY, $token );
		delete_option( self::TOKEN_OPTION_KEY . '_refresh' );
```

public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
```diff
@@ -0,0 +1,207 @@
...
+		$ttl = HOUR_IN_SECONDS;
+
+		$results = get_transient( $cache_key );
+		if ( false === $results ) {
+			$results = $this->fetch_results();
+			if ( is_wp_error( $results ) ) {
+				return $results;
+			}
+			set_transient( $cache_key, $results, $ttl );
+		}
```
dd32:
```suggestion
		$results = get_transient( $cache_key );
		if ( false === $results ) {
			$results = $this->fetch_results();
			if ( is_wp_error( $results ) ) {
				// During development, return the upstream error.
				if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
					return $results;
				}
			
				// Log the error
				trigger_error( $results->get_error_code(), E_USER_WARNING );

				// Set a short timeout to avoid hammering the API during outages.
				set_transient( $cache_key, [], 0.5 * MINUTE_IN_SECONDS );
		
				return new WP_Error(
					'search-request-failed',
					__( 'The search API provider is currently experiencing an error. Try again in a few seconds', 'wporg-patterns' ) // Publicly facing human-readable error.
				);
			}
			set_transient( $cache_key, $results, $ttl );
		}
```

public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
```diff
@@ -0,0 +1,207 @@
...
+		);
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		$response_code = wp_remote_retrieve_response_code( $response );
+		if ( 200 !== $response_code ) {
+			return new WP_Error(
+				'invalid-openverse-request',
+				sprintf( __( 'The Openverse API request failed with a %s error.', 'wporg-patterns' ), $response_code )
```
dd32:
```suggestion
				sprintf(  'The Openverse API request failed with a %s error.', $response_code )
```

public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
```diff
@@ -0,0 +1,207 @@
...
+		}
+
+		$body = wp_remote_retrieve_body( $response );
+		if ( is_wp_error( $body ) ) {
+			return $body;
+		}
+
+		$data = json_decode( $body );
+		if ( null === $data ) {
+			return new WP_Error( 'invalid-openverse-response', __( 'The Openverse API response is malformed.', 'wporg-patterns' ) );
```
dd32:
```suggestion
			return new WP_Error( 'invalid-openverse-response', 'The Openverse API response is malformed.' );
```

public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
```diff
@@ -0,0 +1,207 @@
...
+		) );
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		$response_code = wp_remote_retrieve_response_code( $response );
+		if ( 200 !== $response_code ) {
+			return new WP_Error(
+				'invalid-ov-token-request',
+				sprintf( __( 'The token generation request failed with a %s error.', 'wporg-patterns' ), $response_code )
```
dd32:
```suggestion
				sprintf( 'The token generation request failed with a %s error.', $response_code )
```

public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
```diff
@@ -0,0 +1,207 @@
...
+		}
+
+		$body = wp_remote_retrieve_body( $response );
+		if ( is_wp_error( $body ) ) {
+			return $body;
+		}
+
+		$token = json_decode( $body, true ); // Second param returns decoded value as assoc array.
+		if ( null === $token ) {
+			return new WP_Error( 'invalid-ov-token-response', __( 'The token generation response is malformed.', 'wporg-patterns' ) );
```
dd32:
```suggestion
			return new WP_Error( 'invalid-ov-token-response', 'The token generation response is malformed.' );
```

public_html/wp-content/plugins/pattern-creator/includes/openverse-client.php
```diff
@@ -0,0 +1,207 @@
...
+				'license' => 'cc0',
+				'q' => $this->get_param( 'search' ),
+				'page' => $this->get_param( 'page' ),
+				'page_size' => $this->get_param( 'per_page' ),
+			),
+			$this->url . '/v1/images',
+		);
+		$response = wp_remote_get(
+			$url,
+			array(
```
dd32, line 171:
```suggestion
			array(
				'timeout' => 15,
```

dd32 (commented):
The suggestions in this review are a little much, but I've made the changes with the idea that this API is likely to get abused heavily at times, and as I don't know how long the oauth tokens last, I've assumed that it's likely that requests will come in during the token refresh window.

These changes in short are:
1. Allows requests to be made using the "stale"/almost-expiring token while it's being refreshed
2. Limits the token refreshing to happening with hopefully 1, but maybe more, processes. I'm unsure what would happen if multiple requests refreshed the token at the same time (Would one be invalidated by the other? etc) but mostly to avoid causing all requests to request new tokens during the refresh window.
3. Avoids more issues with the OV api being unavailable (API error, or network issue), by not constantly hitting the upstream API in those cases for the same search (imagine a refresh type scenario)
4. Adds a more friendly human-facing error in the event of a search API HTTP failure in production, and avoids adding the errors as translation strings (when they ideally should not be shown to end-users - and those debugging kind of probably need english errors rather than Arabic)
5. Bumps the timeout from 5s to 15s as network & load issues can cause varying latency, it'd be better to wait a few extra seconds than abort while receiving a large response.

Other than that, I see no issues in the rest of the PR or way of doing it.

ryelle:
Thanks for the improvements! Definitely worth making it less prone to abuse now, rather than after something happens 🙂 

The tokens should stay valid for 10 hours, but there's no guarantee (that's not a documented value). This updated refresh flow makes a lot of sense. And even if an invalid token is used in the `being_refreshed` stage, it should work - it will be subject to rate limits, but _hopefully_ that's only happening for max a few minutes every 10 hours.

> I'm unsure what would happen if multiple requests refreshed the token at the same time

Seems like generating new tokens does not invalidate old tokens, so it should be OK if someone's using the previous token while the refresh happens.

> avoids adding the errors as translation strings

Good point - I've gotten conflicting advice on this across projects in the past (where other devs will be the ones seeing the errors), but untranslated makes sense for these.

PR #396, opened 2022-02-03
Patterns: Update permissions to allow pattern creation and editing by all logged-in users.

### LINKED ISSUE #377: Pattern Creation: Allow anyone with WordPress.org user account to create patterns
Right now, you need to be logged into the site to create patterns, but it should be open to any WordPress.org user.

Any user should be able to create patterns, view the `my-patterns` page, edit their own patterns, but not edit anyone else's patterns.

ryelle:
Fixes #377 — Anyone with a wp.org account can now create, edit, view, and delete their own patterns.

The permissions that were already granted in #249 don't actually cover everything required by the editor.

1. The CPT endpoint checks against `current_user_can( 'read_post', $post->ID )` ([source](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L1606)), which fails for wporg users who aren't members of the site, because they don't have `read` permission.
2. Delete permissions are necessary to delete patterns.

I don't see anything unexpected from adding the `read` perm on the frontend, but I'm requesting review just to make sure there isn't something I'm missing.

1. Log into a user account that isn't a member of w.org/patterns
2. View [wordpress.org/patterns](https://wordpress.org/patterns/), you should have an admin bar with the + New button
3. The only item in that dropdown should be Block Pattern, click it
4. The editor should open without console errors
5. Add a pattern, save it however you want (draft or submit, it will go to pending, not publish)
6. It should save successfully
7. Go to [wordpress.org/patterns/my-patterns](https://wordpress.org/patterns/my-patterns/), you should see the pattern you just created
8. Open it in the editor (either using the 🔽  menu or the Edit link in the admin bar)
9. It should load

You can try editing things you don't have access to, it should prevent you. There are a few issues with permissions + the Query Loop block, but I think the solution there will be mocking, see #23.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -649,6 +649,8 @@ function set_pattern_caps( $user_caps ) {
 	);
 	$cap_map = (array) get_post_type_capabilities( (object) $cap_args );
 
+	// Users should have the same permissions for patterns as posts, for example,
+	// if they have `edit_posts`, they should be granted `edit_patterns`, and so on.
 	foreach ( $user_caps as $cap => $bool ) {
 		if ( $bool && isset( $cap_map[ $cap ] ) ) {
 			$user_caps[ $cap_map[ $cap ] ] = true;
@@ -657,9 +659,13 @@ function set_pattern_caps( $user_caps ) {
 
 	// Set caps to allow for front end pattern creation.
 	if ( is_user_logged_in() && ! is_admin() ) {
-		$user_caps['publish_patterns']        = true;
-		$user_caps['edit_patterns']           = true;
-		$user_caps['edit_published_patterns'] = true;
+		$user_caps['read']                       = true;
+		$user_caps['publish_patterns']           = true;
+		$user_caps['edit_patterns']              = true;
+		$user_caps['edit_published_patterns']    = true;
+		$user_caps['delete_patterns']            = true;
+		$user_caps['delete_published_patterns']  = true;
+		// Note that `edit_others_patterns` & `delete_others_patterns` are separate capabilities.
 	}
 
 	return $user_caps;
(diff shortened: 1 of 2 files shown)


coreymckrill (approved):
👍 Works as expected on wporg sandbox. I was able to create, edit, and delete a pattern. I tried to edit someone else's pattern by putting its ID into the URL, but it showed me an error message.

I don't really see an issue with adding the `read` capability. If I try to go to w.org/patterns/wp-admin, it redirects me to my wporg profile page.

dd32:
> Delete permissions are necessary to delete patterns.

What's the intention of allowing people to delete published patterns?

It seems that maybe we wouldn't want people deleting published patterns, as it provides a poor UX for the users of the pattern. "Yesterday I searched and installed XYZ but I can't find it today".

ryelle:
They'll only be able to delete their own patterns, as part of managing their contributions, if they decide they don't want the pattern public anymore. It won't break anyone's existing content to have a pattern removed, though I see what you mean about it possibly being confusing to a user if the pattern they favorited disappears.

In any case, the button to delete already exists, this just gives the users permissions to do it. If we want to remove deletion of published patterns, we'll need to remove the button then too.

dd32 (approved):
That's fair, I guess I was just trying to avoid a potential left-pad situation :) which can't really happen, as even copy-pasting between sites doesn't rely upon the central block directory.

I see no harm in granting these permissions on the front-end / rest-api.

The worst case I can see here, is that it might be possible to set a future date for publishing via an API request, which might not be intended? (That might be handled via a filter elsewhere though).

This does grant access to https://wordpress.org/patterns/wp-json/wp/v2/wporg-pattern?_context=edit&_embed=1  which lists the raw content for each pattern, but that's not secret. Nor are any of the details it exposes for Terms or Users.
(edit: You'll need to add `&_wpnonce=...` to the above URL to access it, grab the nonce from the pagesource of w.org/patterns.)

ryelle:
> The worst case I can see here, is that it might be possible to set a future date for publishing via an API request, which might not be intended?

Yeah, good point - i don't think _that's_ a problem, but since this gives access to the API, it would be possible for someone to override the current default status (pending) and publish a pattern. I think that should be handled in the validation step though, so I'll do that in another PR before merging this one (eventually the default step will be published, so i don't want to remove the permissions here).

In the last two commits here, I fixed & added some tests to the capability tests for a user "not on the current site", to better reflect the WP.org environment.

ryelle:
> I think that should be handled in the validation step though, so I'll do that in another PR before merging this one

Done in 232ed1274ace6bda45fb55b17807f74c36e54d6c.

Issue #22, opened 2020-11-16
Validate submitted patterns


tellyworth:
Patterns need to be checked for validity during the submission process. Probably different levels of validation at different points:

### Early:
* Is it even a pattern? (Contains blocks, not random binary, etc)
* Does it meet some minimum size/length? (At least one block with non-default attributes or content?)
* Does it contain JS or inline CSS?
* Do the blocks exist?
* Is the markup minimally valid?

### Later:
* Does it use permitted blocks? (Core only, or an allow/deny list?)
* Does it contain forbidden links or keywords? (Spam etc)
* Does it have sufficient metadata?

### Last:
* Does it have the same name or content as another published block?
* Does the content meet community standards? (Offensive language or images)
* Does it pass a license check (if applicable for images or other media)?
* Is the author following appropriate guidelines?

I imagine the Early checks would be applied before saving a pattern at all. Later checks would need to pass in order to submit a pattern for publication. And the Last checks would be made at least in part by humans, before approving a pattern for inclusion in the public directory.

If we assume that ["private" patterns](https://github.com/WordPress/pattern-directory/issues/16#issuecomment-725845843) will eventually become a thing, then the Early checks might be sufficient for saving a pattern for private use only.

carolinan:
Some of the points here aligns closely with the need to validate the block markup of full site editing themes when they are submitted to the directory.

dingo-d:
My 2 cents: If you want to have a fast validation that doesn't depend on a human being since that's slow (coming from experience with theme reviews), why not explore the option to implement machine learning algorithms for `Offensive language or images`?

It's not like these aren't out there ([link](https://sightengine.com/), [link (archived)](https://github.com/yahoo/open_nsfw), [link](https://github.com/GantMan/nsfw_model)). I'm sure there are people who would be up to the task of using some of these and programing some kind of validation tool that would speed up the process.

ryelle:
I started some validation in #38, pulling out my comments so they're not lost in the PR:

**Does it contain JS or inline CSS?**
These should be stripped out by `wp_kses` for users, but we'll want to check that when we figure out permissions in general, #30

**Is the markup minimally valid?**
[...]

PR #406, opened 2022-02-23
Add spam checking on pattern submissions

dd32:
This PR adds some anti-spam checks to pattern submissions:
 - Akismet
 - Basic paragraph-only detection (ie. if a submission is only paragraphs, it's likely to require manual review / be an automated bot)
 - A keyword to use for testing the spam process.

In order to simplify the text being checked, I leveraged the Pattern Translator codebase to extract all strings from the submitted content, rather than simply passing the HTML to Akismet, to avoid any issues related to HTML.

See #22

1. Apply PR
2. Create a Spam pattern. (Either by including `PatternDirectorySpamTest` within the content, or the pattern only containing `core/paragraph` blocks)
3. Verify that the post is pending review.

You can also try running `yarn test:php` however I've been unable to run phpunit locally due to `wp-env` problems, so I'm not able to check the unit test additions actually work.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
@@ -1,7 +1,7 @@
 <?php
 
 namespace WordPressdotorg\Pattern_Directory\Pattern_Validation;
-use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\{POST_TYPE,UNLISTED_STATUS,SPAM_STATUS};
+use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\{ POST_TYPE, UNLISTED_STATUS, SPAM_STATUS };
 
 use WordPressdotorg\Pattern_Translations\Pattern as Translations_Pattern;
 use WordPressdotorg\Pattern_Translations\PatternParser as Translations_PatternParser;
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
@@ -1,7 +1,7 @@
 <?php
 
 namespace WordPressdotorg\Pattern_Directory\Pattern_Validation;
-use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\POST_TYPE;
+use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\{POST_TYPE,UNLISTED_STATUS,SPAM_STATUS};
 
 use WordPressdotorg\Pattern_Translations\Pattern as Translations_Pattern;
 use WordPressdotorg\Pattern_Translations\PatternParser as Translations_PatternParser;
@@ -180,7 +180,7 @@ function validate_status( $prepared_post, $request ) {
 	$current_status = get_post_status( $prepared_post->ID );
 
 	// Drafts or unlisted patterns are OK.
-	if ( in_array( $target_status, [ 'draft', 'auto-draft', 'unlisted' ] ) ) {
+	if ( in_array( $target_status, [ 'draft', 'auto-draft', UNLISTED_STATUS ] ) ) {
 		return $prepared_post;
 	}
 
@@ -190,7 +190,7 @@ function validate_status( $prepared_post, $request ) {
 	}
 
 	$default_status = get_option( 'wporg-pattern-default_status', 'publish' );
-	$valid_states   = array_unique( array( 'pending', $default_status ) );
+	$valid_states   = array_unique( array( 'pending', SPAM_STATUS, $default_status ) );
 
 	// Make sure the target status is the expected status (publish or pending).
 	if ( ! in_array( $target_status, $valid_states, true ) ) {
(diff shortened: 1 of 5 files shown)


public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
```diff
@@ -200,5 +204,141 @@ function validate_status( $prepared_post, $request ) {
...
+			'comment_author_url'   => '',
+			'comment_content'      => $combined_strings,
+			'comment_content_raw'  => $content,
+			'permalink'            => get_permalink( $post ),
+		);
+
+		$akismet = \Akismet::rest_auto_check_comment( $akismet_payload );
+		if ( is_wp_error( $akismet ) ) {
+			$akismet = array( 'akismet_result' => 'discard' );
+		}
```
ryelle, line 318: Is this saying treat it as spam if Akismet errors?

public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
```diff
@@ -200,5 +204,141 @@ function validate_status( $prepared_post, $request ) {
...
+	}
+
+	// Testing keyword. Case-sensitive.
+	if ( ! $is_spam && str_contains( $combined_strings, 'PatternDirectorySpamTest' ) ) {
+		$is_spam = true;
+	}
+
+	// If it's been detected as spam, flag it as pending-review.
+	if ( $is_spam ) {
+		$prepared_post->post_status = 'pending';
```
ryelle: Do you think we could also attach a note using the Internal Notes plugin? Otherwise it's not clear if this was pending on purpose (the default status is `pending`) vs spammed into pending.

ryelle (commented):
The tests pass locally, and [on the PR itself](https://github.com/WordPress/pattern-directory/runs/5300816056?check_suite_focus=true). Manually tested as well. I haven't been able to trigger Akismet spamming a pattern, but the other two checks work as expected.

public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
dd32: No.

This is some uglyness from the Akismet plugin API not being designed to be used outside of comment filters.

If the "Discard the worst of the worst spam" is enabled, Akismet will return a WP_Error instead of changing the status to spam when it thinks it's being called from the REST API (that's why `rest_auto_check_comment` is called). If it thinks it's being called outside of the rest-api then it redirects and/or die's.

public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
dd32: Ah, I had thought the plan was for auto-publish, not always going to pending.
Would it make more sense to add a different post_status for `spam` then?

Doesn't look like Internal Notes is in the local environment, so I didn't realise it was being used.

dd32: >  I haven't been able to trigger Akismet spamming a pattern

I think about half of the seemingly-spam submissions to w.org/patterns triggered it (I copy pasted them over to local as a test), based on those posts I then added the `only-paragraphs` check.

I find it's harder to trigger Akismet with content than one would expect.. Looking at a public blogs spam comments should also pull up a few strings it'll trigger on.

dd32: > Ah, I had thought the plan was for auto-publish, not always going to pending.

Clarification on that, I had assumed this due to everyone having `publish_pattern` capabilities from abb07a078e3041b14abef37ed1fa77f25f66618e.  Upon reviewing this, it seems like this method will prevent `publish`, so that's fine.

dd32: > Would it make more sense to add a different post_status for spam then?

I've explored that in: https://github.com/dd32/pattern-directory/compare/add/pattern-validation...dd32:add/spam-status

ryelle: > Ah, I had thought the plan was for auto-publish, not always going to pending

It's an option, I don't know which we'll launch with. I think the plan is to eventually auto-publish, but be pending at launch to add a little bit of a manual quality check.

> Doesn't look like Internal Notes is in the local environment, so I didn't realise it was being used.

Oh oops, it should get added to the composer deps.

public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
```diff
@@ -200,5 +204,158 @@ function validate_status( $prepared_post, $request ) {
...
+		$pattern->locale      = get_locale();
+
+		$parser  = new Translations_PatternParser( $pattern );
+		$strings = $parser->to_strings();
+	}
+
+	// Combine strings for ease of use.
+	$combined_strings = implode( "\n", $strings );
+	$combined_links   = implode( "\n", $links );
+	$combined         = $combined_strings . "\n" . $combined_links;
```
ryelle: I don't see `$combined` used, should this be sent to akismet instead of just `$combined_strings`?

public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
ryelle: > I find it's harder to trigger Akismet with content than one would expect.. Looking at a public blogs spam comments should also pull up a few strings it'll trigger on.

Finally got it to trigger, though I had to change the `comment_author` to a spammy name. The note added looks good 👍🏻

ryelle: > Would it make more sense to add a different post_status for spam then?

I like `add/spam-status`, I think that would make it much easier to separate out spam especially with everything pending at launch. Since you've already got it in a separate branch, would you rather make a separate PR for it once this lands, or combine it all?

public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
dd32: Yeah, `$combined` isn't being used, nor is `$links` / `$combined_links`.

I had added those with the intention of using them, but never did.. 

I'll pull those out for now.

public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
dd32: > I like add/spam-status, I think that would make it much easier to separate out spam especially with everything pending at launch.

Sounds good, I'll merge it into this PR.

> Since you've already got it in a separate branch

It was a branch forked from this branch, so I could quickly merge it into this branch if we went ahead with it.

PR #414, opened 2022-02-28
Remove custom layout declarations.

### LINKED ISSUE #389: Unable to set full/wide width for a pattern
**Describe the bug**
Within the /new-pattern/ experience, I am unable to set a pattern align attribute to wide/full. Is this intentional? 

For [this pattern](https://wordpress.org/patterns/?post_type=wporg-pattern&p=9244), I believe I built it in the post.php view, and copy/pasted it in (which maintained the fullwidth value). 

**Screenshot**
This is a test pattern

StevenDufresne:
Fixes #389 

This PR removes the custom Layout object that modifies internal Gutenberg functionality to allow custom block alignments.
 
Gutenberg supports full width alignments when `add_theme_support('align-wide')` is added to a theme. We have that included in our `functions.php` which is why the full-width alignments show up when adding blocks on a new post/page. However, `new-pattern` uses our own import of the `Editor/BlockEditor` and therefore behaves differently. The problem here was solved in two steps:
- Remove the custom layout passed into the Block editor
- Set `supportsLayout` to `false` for the block editor. 

### Considerations
- If we allow users to set widths that are not supported in their themes, is that a problem? No, the alignment tool will not show anything selected when a user toggles the control and once the user updates the pattern on their site, it will replace the pattern's default.
- Is turning off `supportsLayout` going to cause problems in the future since it's a forward-leaning feature? 
  - I don't think so, but not sure. But since it's only affecting how patterns are created, we can remove it in the future if it causes problems. 

### Screenshots
**Enabled on Image Block**

**Enabled on Group Block**

1. Visit `new-pattern`
2. Add an image block
3. Verify you have all the alignments shown in screenshot.
4. Save pattern with a "full" alignment
5. Copy your pattern and expect to see that alignment attached.

**Reading/Code References**
- Where Gutenberg determines alignment support: [Link](https://github.com/WordPress/gutenberg/blob/32ecd97e988174b6401289c44200135c224ca104/packages/block-editor/src/components/block-alignment-control/use-available-alignments.js#L16)
- The default layout configurations: [Link](https://github.com/WordPress/gutenberg/blob/bf405a3f1280cade6b9f480c973261b987d5fd53/packages/block-editor/src/layouts/flow.js)

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/src/components/block-editor/index.js
+++ b/public_html/wp-content/plugins/pattern-creator/src/components/block-editor/index.js
@@ -26,12 +26,6 @@ import { useMergeRefs } from '@wordpress/compose';
 import { POST_TYPE, store as patternStore } from '../../store';
 import { SidebarInspectorFill } from '../sidebar';
 
-const LAYOUT = {
-	type: 'default',
-	// At the root level of the site editor, no alignments should be allowed.
-	alignments: [],
-};
-
 export default function BlockEditor( { setIsInserterOpen } ) {
 	const { settings, deviceType } = useSelect(
 		( select ) => {
@@ -73,10 +67,7 @@ export default function BlockEditor( { setIsInserterOpen } ) {
 					contentRef={ mergedRefs }
 					name="editor-canvas"
 				>
-					<BlockList
-						className="pattern-block-editor__block-list wp-site-blocks"
-						__experimentalLayout={ LAYOUT }
-					/>
+					<BlockList className="pattern-block-editor__block-list wp-site-blocks" />
 				</Iframe>
 			</BlockTools>
 		</BlockEditorProvider>
--- a/public_html/wp-content/plugins/pattern-creator/src/components/editor/index.js
+++ b/public_html/wp-content/plugins/pattern-creator/src/components/editor/index.js
@@ -78,6 +78,7 @@ function Editor( { initialSettings, onError, postId } ) {
 		const result = {
 			...settings,
 			fullscreenMode: true,
+			supportsLayout: false,
 			__experimentalLocalAutosaveInterval: 30,
 			__experimentalSetIsInserterOpened: setIsInserterOpened,
 		};


ryelle:
Can you briefly explain what `supportsLayout` does? I don't really understand why setting that to false doesn't turn off the wide/full layouts.

StevenDufresne:
> Can you briefly explain what `supportsLayout` does? I don't really understand why setting that to false doesn't turn off the wide/full layouts.

Leaving it `true`, removes the wide/full layouts.

To be honest, I can't say that I understand the code flow 100%, but here is some evidence:
- The Editor used for posts/pages have `supportsLayout` set to `false`.
- The [`contentWide` and `wideSize` variables](https://github.com/WordPress/gutenberg/blob/d989ce3d96bdbabc39b6630c23cdcc52b7f7bcca/packages/block-editor/src/layouts/flow.js#L178) are always undefined when it's `true`.
  - I think it's related to our theme not having a[ theme.json ](https://make.wordpress.org/core/2021/06/29/on-layout-and-content-width-in-wordpress-5-8/) with the specific properties... but not 100%.
-  When it's `true` we [fall into this block.
](https://github.com/WordPress/gutenberg/blob/d989ce3d96bdbabc39b6630c23cdcc52b7f7bcca/packages/block-editor/src/components/block-alignment-control/use-available-alignments.js#L36)
    - The `layoutAlignments` is missing the "full" alignments because `contentSize` and `wideSize` are not set.
    - Ie `['none', 'left', 'center', 'right',].filter( ( { name: alignmentName } ) => ['none', 'left', 'center', 'right', 'wide', 'full' ].includes( alignmentName ) )`;

Ref:[ PR that introduced this logic.](https://github.com/WordPress/gutenberg/pull/33359)

I decided to stop digging deeper into the code since the page/post editors have it `false` and I currently don't have an elegant way to test gutenberg and a custom editor. If we are not comfortable with setting it to false, i can dig in a bit deeper.

ryelle:
Okay, I _think_ I get it, that it's about the layout controls not specifically about the wide/full alignments, but has the effect of enabling wide/full since we don't set layout values in a `theme.json`. I think leaving it `false` makes sense for now. We might run into issues if people want to submit site layout patterns, but we can make an issue & figure that out if/when it comes up.

ryelle (approved):

PR #419, opened 2022-03-02
Pattern Preview: Update styles to match editor.

### LINKED ISSUE #416: Alignment & spacing is not passed to the frontend preview
Pulled out of #388. When styles are applied using the custom `wp-container-*` classes, these are not applied to the preview content. This is most obvious if you try to add & center a button:

Editor:

Frontend preview:

The root of the issue is that the inline `style` tags are added on a hook, either `wp_footer` or `wp_enqueue_scripts` ([source](https://github.com/WordPress/gutenberg/blob/3efbac2056858211aeda54e2ba8751817d9aaee7/lib/compat/wordpress-5.9/script-loader.php#L63)), rather than output directly inline, so the API response does not contain the styles (see https://github.com/WordPress/gutenberg/issues/35376#issuecomment-975156782, https://github.com/WordPress/gutenberg/issues/38167).

ryelle:
Fixes #416 — The pattern content rendered in the preview frame is pulled from the API, and is only the post content. On regular page loads, there are extra styles output in the footer for alignment, spacing, and link color (if applicable). These were not being added to the preview, causing differences between editor & frontend. Additionally, the default preset classes were not loaded, so background colors, etc, were not applied.

1. To fix the extra `style` tag issues, I've copied over the gutenberg code to render these styles, and am injecting them directly into the content. This should only be a temporary fix, until the Style Engine work is done, since one of the bugs mentioned is this issue with API responses.
2. The preset classes were not loaded (due to [unregistering `global-styles` in wporg-mu-plugins](https://github.com/WordPress/wporg-mu-plugins/blob/7674f5dbf3c0d326008ad1d4884946e2b7ee4a60/mu-plugins/blocks/global-header-footer/blocks.php#L258)), so those are loaded in when generating the styles to inject into the preview iframe. They're immediately unregistered to avoid overwriting the properties for the header & footer (see https://github.com/WordPress/wporg-mu-plugins/pull/166).

### Screenshots

| Editor | Before | After |
|--------|--------|-------|
|  |  |  |

1. Create a new pattern
2. Add a group block and give it a background color or gradient
3. Add more content inside, a Row or Buttons would work
4. Change the default alignment of items, for example, to center
5. Once you have a customized pattern, save it
6. Preview it on the frontend
7. It should match the editor view

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -5,6 +5,8 @@ namespace WordPressdotorg\Pattern_Directory\Theme;
 use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\POST_TYPE;
 use const WordPressdotorg\Pattern_Directory\Pattern_Flag_Post_Type\POST_TYPE as FLAG_POST_TYPE;
 
+require_once __DIR__ . '/includes/inline-styles.php';
+
 add_action( 'after_setup_theme', __NAMESPACE__ . '\setup' );
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets', 20 );
 add_action( 'wp_head', __NAMESPACE__ . '\generate_block_editor_styles_html' );
@@ -115,7 +117,8 @@ function enqueue_assets() {
  * See https://github.com/WordPress/gutenberg/blob/6ad2a433769a4514fc52083e97aa47a0bc9edf07/lib/client-assets.php#L710
  */
 function generate_block_editor_styles_html() {
-	$handles = array( 'wp-block-library' );
+	wp_enqueue_global_styles();
+	$handles = array( 'wp-block-library', 'global-styles' );
 
 	$block_registry = \WP_Block_Type_Registry::get_instance();
 
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -149,6 +149,9 @@ function generate_block_editor_styles_html() {
 		),
 		'before'
 	);
+
+	wp_dequeue_style( 'global-styles' );
+	wp_deregister_style( 'global-styles' );
 }
 
 /**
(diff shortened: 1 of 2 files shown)


StevenDufresne:
I was able to create a discrepancy with this block:
```
<!-- wp:group {"align":"wide","gradient":"cool-to-warm-spectrum"} -->
<div class="wp-block-group alignwide has-cool-to-warm-spectrum-gradient-background has-background"><!-- wp:buttons {"align":"wide","layout":{"type":"flex","justifyContent":"right"}} -->
<div class="wp-block-buttons alignwide"><!-- wp:button {"backgroundColor":"luminous-vivid-amber","textColor":"black","width":50,"align":"center","style":{"spacing":{"padding":{"top":"20px","right":"60px","bottom":"20px","left":"60px"}},"border":{"radius":"40px"}},"className":"is-style-fill"} -->
<div class="wp-block-button aligncenter has-custom-width wp-block-button__width-50 is-style-fill"><a class="wp-block-button__link has-black-color has-luminous-vivid-amber-background-color has-text-color has-background" style="border-radius:40px;padding-top:20px;padding-right:60px;padding-bottom:20px;padding-left:60px">Button 1</a></div>
<!-- /wp:button --></div>
<!-- /wp:buttons --></div>
<!-- /wp:group -->
```

| Editor | Preview |
| --- | --- |
|  |   |

ryelle:
@StevenDufresne It looks like this is a bug from Gutenberg — I can reproduce the same issue across themes in the post editor on another test site, a style applied to buttons with `aligncenter` overrides the custom width applied. How did you add the `"align":"center"` attribute on to the button?

StevenDufresne (approved):

StevenDufresne:
> @StevenDufresne It looks like this is a bug from Gutenberg — I can reproduce the same issue across themes in the post editor on another test site, a style applied to buttons with `aligncenter` overrides the custom width applied. How did you add the `"align":"center"` attribute on to the button?

I can't remember specifically, but everything was done through the interface. I can try to retrace my steps if you would like.

ryelle:
> I can't remember specifically, but everything was done through the interface. I can try to retrace my steps if you would like.

Hm, don't worry about it for now— we know it's not a bug with the creator, so if you did manage to reproduce it, it should be reported up to gutenberg.

PR #420, opened 2022-03-02
Pattern Creator: Add `theme.json` to theme to enable appearance tools.

ryelle:
This PR adds a `theme.json` file to enable [appearanceTools](https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-json/#opt-in-into-ui-controls), which enables all the possible UI controls for typography, colors, spacing, etc. It also enables the custom link color, since `add_theme_support( 'experimental-link-color' );` doesn't seem to work.

Just adding a `theme.json` doesn't trigger `wp_is_block_theme`, so we don't need to worry about any conflicts or issues with the global header.

⚠️  There's a bug where the theme style overrides the link color when the container has a background, I'm not sure if this is a theme issue or not since I can't customize the link color on TT1 [#55028-core](https://core.trac.wordpress.org/ticket/55028).

### Screenshots

New settings available in sidebar:

Screen Shot 2022-03-02 at 3 08 13 PM

1. Create a new pattern
2. See all the UI options available to you
3. Use them, they should appear on the frontend correctly

Note that this PR builds on #419 so that the styles are injected as expected.

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -28,13 +28,8 @@ add_action( 'template_redirect', __NAMESPACE__ . '\rewrite_urls' );
  * as indicating support for post thumbnails.
  */
 function setup() {
-	add_theme_support( 'post-thumbnails' );
-
 	// Add gutenberg styling supports.
 	add_theme_support( 'align-wide' );
-	add_theme_support( 'custom-spacing' );
-	add_theme_support( 'custom-line-height' );
-	add_theme_support( 'experimental-link-color' );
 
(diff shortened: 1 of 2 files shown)


StevenDufresne (approved):
Yep, it looks good. The one thing to note is that none of those inputs were showing until I enabled them. I would be surprised if many users are able to find them tucked behind the "..." vertical menu.

ryelle:
> The one thing to note is that none of those inputs were showing until I enabled them. I would be surprised if many users are able to find them tucked behind the "..." vertical menu.

Good point - that is how it works in gutenberg, so anyone used to working in GB should know to check the dots, but I'll see if I can get them to default on in the pattern creator (in another PR)

PR #422, opened 2022-03-03
Creator: Fix padding in editor

StevenDufresne:
Introducing Twenty Twenty styles into the editor in #417 created some small regressions.

- The padding around the content in the editor is gone
- The first "block" jumps upwards after first interaction.

This is the result of the Twenty Twenty theme overriding margin and padding on some of Gutenberg's [classic.scss](https://github.com/WordPress/gutenberg/blob/b6c79fed01b0a18802f9603176746c92cde1b02d/packages/edit-post/src/classic.scss#L23
) styles.

These styles are meant to be a fallback for themes that currently don't use `theme.json` layouts. See comment in css file for more information.

## Current Solution
Copy the Gutenberg styles to our stylesheet.  

#### Re-arrange CSS imports, move Gutenberg after Twenty Twenty one
This would most likely introduce other layout issues.

#### Control Via theme.json
Try this after #420 is merged

## Remaining Issues
- The Block list appender ([+]) still jumps up higher than expected because it's absolutely positioned. In the default editor it isn't noticeable because the title styles are different.

### Screenshots
| Before | After |
| --- | --- |
|  |    | 
|  |    | 

1. Go to `new-pattern`
2. Expect to see around the content
3. Resize the window, expect the padding persists.

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/src/components/block-editor/style.scss
+++ b/public_html/wp-content/plugins/pattern-creator/src/components/block-editor/style.scss
@@ -25,3 +25,11 @@
 .pattern-visual-editor > [name="editor-canvas"] {
 	margin-top: 0 !important;
 }
+
+
+// Restore Gutenberg Styles that are overwritten by twenty twenty one
+html :where(.wp-block) {
+	margin-bottom: 28px;
+	margin-top: 28px;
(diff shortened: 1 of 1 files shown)


StevenDufresne:
Closing in favor of #428's approach.

PR #426, opened 2022-03-09
Creator: Move initial editor settings into initialize function

ryelle:
This straightens out the editor settings somewhat by synchronously setting them (`updateSettings`) in `reinitializeEditor`. It also cleans up where the preference checks happen - previously there were some in `<Editor />` and some in the `getSettings` selector; now building the whole settings object happens in the selector.

There should be no visual or functional changes. The code should be more readable. Make sure the editor still works :)


StevenDufresne (approved):
This is a nice improvement and will make the code more readable. 👍

PR #427, opened 2022-03-09
Add Internal Notes plugin to composer dependencies

### LINKED ISSUE #411: Tools: Add Internal Notes plugin to composer dependencies
The [Internal Notes plugin](https://github.com/WordPress/wporg-internal-notes/) is used on the Patterns site, but is not installed in the local environment. Other plugin dependencies are installed with composer, so Internal Notes should get added that way too.

coreymckrill:
Now that the Internal Notes plugin has a [build branch](https://github.com/WordPress/wporg-internal-notes/pull/8), we can add that as a composer dependency without adding to the Pattern Directory's own overhead for building stuff when setting up the local dev environment.

Fixes #411 

1. Run `composer install`
2. The `wporg-internal-notes` folder should now appear inside `wp-content`. It should be the built version, with no development-related files.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,6 +55,10 @@
 		{
 			"type": "vcs",
 			"url": "git@github.com:WordPress/wporg-mu-plugins.git"
+		},
+		{
+			"type": "vcs",
+			"url": "git@github.com:WordPress/wporg-internal-notes.git"
 		}
 	],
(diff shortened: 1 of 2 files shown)


ryelle (approved):
👍🏻  Installed the notes plugin as expected

PR #428, opened 2022-03-09
Creator: Fix layout & alignment options

ryelle:
This is an assortment of trial and error right now, trying to get the alignment options back onto top-level blocks _and_ working so that default alignment is not edge to edge/full width.

Adding `__experimentalLayout` to `BlockList` & `layout` to `theme.json` enables wide/full alignment & uses the 800px/1000px values:
Screen Shot 2022-03-09 at 6 45 17 PM

Align wide behaves as expected:
Screen Shot 2022-03-09 at 6 42 51 PM

"None" alignment still displays as full width in the editor _and_ the frontend.
Screen Shot 2022-03-09 at 6 47 01 PM

The toolbar often overlaps the first block, I'm not sure why, I was hoping figuring out the alignment would also fix whatever style conflict is happening here. Might be related to #422 
Screen Shot 2022-03-09 at 6 49 25 PM

@StevenDufresne This is the layout stuff I mentioned in DM - I don't really have anything coherent right now, because it's still not really doing what I want, but if you have any comments I'd appreciate it.

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
+++ b/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
@@ -222,7 +222,12 @@ function rest_api_init() {
 add_action( 'rest_api_init', __NAMESPACE__ . '\rest_api_init' );
 
 /**
- * Add Twenty Twenty-One styles, set the editor background to white, and add the correct layout settings.
+ * Filter editor settings to add extra styles to the Pattern Creator editor.
+ *
+ * This adds `link` & `style` tags to be loaded into the editor's iframe.
+ * - Load Twenty Twenty-One styles for a theme preview
+ * - Set the editor background to white for a cleaner preview
+ * - Add layout styles for the pattern container so that alignments work
  *
  * @param array $settings Default editor settings.
  * @return array Updated settings.
--- a/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
+++ b/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
@@ -155,6 +155,8 @@ function pattern_creator_init() {
 	wp_enqueue_script( 'wp-format-library' );
 	wp_enqueue_style( 'wp-edit-site' );
 	wp_enqueue_style( 'wp-format-library' );
+	// Load layout and margin styles.
+	wp_enqueue_style( 'wp-editor-classic-layout-styles' );
 	wp_enqueue_media();
 }
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\pattern_creator_init', 20 );
--- a/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
+++ b/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
@@ -220,3 +220,27 @@ function rest_api_init() {
 	$controller->register_routes();
 }
 add_action( 'rest_api_init', __NAMESPACE__ . '\rest_api_init' );
+
+/**
+ * Add Twenty Twenty-One styles, set the editor background to white, and add the correct layout settings.
+ *
+ * @param array $settings Default editor settings.
+ * @return array Updated settings.
+ */
+function add_theme_styles_to_editor( $settings ) {
+	if ( ! isset( $settings['__unstableResolvedAssets']['styles'] ) ) {
+		return $settings;
+	}
+
+	$block_gap = wp_get_global_styles( array( 'spacing', 'blockGap' ) );
+	$layout = wp_get_global_settings( array( 'layout' ) );
+	$style = gutenberg_get_layout_style( '.pattern-block-editor__block-list.is-root-container', $layout, true, $block_gap );
+
+	$settings['__unstableResolvedAssets']['styles'] .=
+		'\n<link rel="stylesheet" id="theme-styles" href="https://wp-themes.com/wp-content/themes/twentytwentyone/style.css" media="all" />'; //phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedStylesheet
+	$settings['__unstableResolvedAssets']['styles'] .=
+		'\n<style>body.editor-styles-wrapper { background-color: white; }' . $style . '</style>';
+
+	return $settings;
+}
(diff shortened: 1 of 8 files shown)


public_html/wp-content/plugins/pattern-creator/pattern-creator.php
```diff
@@ -155,6 +155,8 @@ function pattern_creator_init() {
 	wp_enqueue_script( 'wp-format-library' );
 	wp_enqueue_style( 'wp-edit-site' );
 	wp_enqueue_style( 'wp-format-library' );
+	// Load layout and margin styles.
+	wp_enqueue_style( 'wp-editor-classic-layout-styles' );
```
StevenDufresne, line 159: TIL 👍

public_html/wp-content/plugins/pattern-creator/src/components/block-editor/index.js
```diff
@@ -73,7 +75,10 @@ export default function BlockEditor( { setIsInserterOpen } ) {
 					contentRef={ mergedRefs }
 					name="editor-canvas"
 				>
-					<BlockList className="pattern-block-editor__block-list wp-site-blocks" />
+					<BlockList
+						className="pattern-block-editor__block-list wp-site-blocks"
+						__experimentalLayout={ layout }
```
StevenDufresne, line 74: Passing the layout setting like this is solving a few issues I was seeing when I was passing the old object we copied from Gutenberg, so that's good. 👍

StevenDufresne:
> The toolbar often overlaps the first block, I'm not sure why, I was hoping figuring out the alignment would also fix whatever style conflict is happening here. Might be related to https://github.com/WordPress/pattern-directory/pull/422

I don't think it's related and this PR will replace #422. I'm 100% certain which overlap you are speaking about specifically, I also see overlap in the default page/post editors but i may be looking at the wrong spots.

<hr>

The alignment looks right to me (i think?)

ryelle:
In https://github.com/WordPress/pattern-directory/pull/428/commits/d92b859b399bd93d5771698dbce7be83789cd025 I added CSS for the alignments. In the pattern creator (editor), I was able to use `gutenberg_get_layout_style`, which renders those styles for all the blocks with layout. ~For the frontend previews, I needed to add the same CSS to the iframe, which causes some duplication, but I think that's the best solution for launch.~ Actually using `generate_block_editor_styles_html` so that we can use `gutenberg_get_layout_style` here too.

Preview thumbnails - the preview window width is now 1200 to show the difference between default/wide/full blocks. Previously it was 800, so all blocks appeared "full"



Single pattern preview:


Editor:


>  I also see overlap in the default page/post editors but i may be looking at the wrong spots.

Hm, I don't see the overlap in Posts, the toolbar is correctly above the content. In the Patterns screenshot it's overlapping the first two lines of the paragraph.

| Pattern | Post |
|----|----|
|  |  |

public_html/wp-content/plugins/pattern-creator/pattern-creator.php
```diff
@@ -218,3 +220,27 @@ function rest_api_init() {
 	$controller->register_routes();
 }
 add_action( 'rest_api_init', __NAMESPACE__ . '\rest_api_init' );
+
+/**
+ * Add Twenty Twenty-One styles, set the editor background to white, and add the correct layout settings.
+ *
+ * @param array $settings Default editor settings.
+ * @return array Updated settings.
+ */
```
StevenDufresne, line 234: This is going to be hard to understand in a few months after a context switch, or for someone new. Is there a way to over-communicate what's happening?

If we change from Twenty Twenty-One to another theme in the future, does this need to be removed?

public_html/wp-content/plugins/pattern-creator/src/index.js
```diff
@@ -24,13 +25,16 @@ apiFetch.use( filterEndpoints );
...
+ * @param {Object}  settings        Editor settings.
+ * @param {number}  settings.postId ID of the current post.
  */
-export function reinitializeEditor( target, settings ) {
+export function reinitializeEditor( target, { postId, ...settings } ) {
 	unmountComponentAtNode( target );
 	const reboot = reinitializeEditor.bind( null, target, settings );
-	render( <Editor initialSettings={ settings } onError={ reboot } />, target );
+
+	dispatch( patternStore ).updateSettings( settings );
```
StevenDufresne, line 38: Why do we need to `dispatch` here? Does the order matter?

ryelle: It's been removed from `<Editor />`, so this should only run on the initialization of the editor (or re-initialization if necessary), and since it's a synchronous call, it happens before rendering, and sets up the settings without needing a `useEffect`. 

See https://github.com/WordPress/gutenberg/blob/a233dc45f1316ab05764840eebcb94373400add4/packages/edit-site/src/index.js#L57-L58. I'll add a comment explaining this too.

public_html/wp-content/plugins/pattern-creator/pattern-creator.php
ryelle: I added some extra documentation - https://github.com/WordPress/pattern-directory/pull/428/commits/d7445b1d0e1766d18308075c1ec69639b858cf29 & https://github.com/WordPress/pattern-directory/pull/428/commits/3b9731f179e6a18b494ba33c3d0405f2b52df998 - hopefully that clears it up a bit.

> If we change from Twenty Twenty-One to another theme in the future, does this need to be removed?

If we want to change the theme used in the previews, it would need to be updated here & in the [theme Iframe component](https://github.com/WordPress/pattern-directory/blob/b73e61da2dcb998182e82b517e3916269d1e3fe2/public_html/wp-content/themes/pattern-directory/src/components/iframe/index.js#L94). We'll always want to pass _some_ styles though, so this wouldn't be removed entirely.

ryelle: Also, the layout styles will probably always be needed, unless the `wporg-patterns` theme itself switches to a block-based theme (I think… if I'm reading the layout logic correctly).

ryelle:
I'm going to merge this now, but if there are any other concerns we can follow up on those in other issues/PRs.

PR #430, opened 2022-03-10
Fix Spinner does not stop when favorites is an empty list

### LINKED ISSUE #362: Favorites: Empty screen loading state never completes
The empty Favorites screen (/patterns/favorites/) include a navigation for filtering an empty list. Clicking on one of the links displays the loading message, which never goes away because there's nothing _to_ load.

image

The navigation and sorting dropdown should not be displayed when there are no favorites.

outdoor2kode:
If a user didn't add any pattern to its favorite list, when switching to other categories, the loading spinner would keep spinning even if the loading is over.

### Root cause
**Details**
This is the logic that decides whether to stop showing the spinner.
https://github.com/WordPress/pattern-directory/blob/21e69c4986286c8a0db749999b649b58947456b7/public_html/wp-content/themes/pattern-directory/src/components/context-bar/index.js#L115
Its logic comes from this function:
https://github.com/WordPress/pattern-directory/blob/21e69c4986286c8a0db749999b649b58947456b7/public_html/wp-content/themes/pattern-directory/src/components/context-bar/index.js#L69
https://github.com/WordPress/pattern-directory/blob/21e69c4986286c8a0db749999b649b58947456b7/public_html/wp-content/themes/pattern-directory/src/store/selectors.js#L20-L24
`isLoading` would always be `true` when the favorites list is empty (i.e., You won't see any patterns under Favorites `All`)
Once there's a pattern in the favorites list, `isLoading` would become `false` (explained later), and the problem is solved then.

The reason `isLoading` is always `true` in this case is that the `queryString` could not be found in `state.patterns.queries`.
Why so? That's because we didn't call `getPatternsByQuery( modifiedQuery )`
https://github.com/WordPress/pattern-directory/blob/21e69c4986286c8a0db749999b649b58947456b7/public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js#L37
when the fav list is empty, which causes that the `queryString` isn't stored via dispatching the`loadPatterns`.
And this is also why `isLoading` would become `false` once there's a pattern in the favorites list - the `getPatternsByQuery( modifiedQuery )` is called.
https://github.com/WordPress/pattern-directory/blob/21e69c4986286c8a0db749999b649b58947456b7/public_html/wp-content/themes/pattern-directory/src/store/resolvers.js#L53-L58

So here comes the first solution

**Sol 1** 
Call `getPatternsByQuery( modifiedQuery )` no matter if the `favorites.length===0` or not
```js
const temp = getPatternsByQuery( modifiedQuery );
const posts = favorites.length ? temp : [];
```
However, though it solved the `isLoading` is always `true` problem, it would bring another issue.
Since we don't have any items in the favorites list for the moment, the `include` in `modifiedQuery` would be`[]`, and when this `modifiedQuery` is sent as payload through the `getPatternsByQuery( modifiedQuery )`, the `include=[]` would be filtered out in advance in `addQueryArgs` before sending to the endpoint `'/wp/v2/wporg-pattern'`.
https://github.com/WordPress/pattern-directory/blob/21e69c4986286c8a0db749999b649b58947456b7/public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js#L35
https://github.com/WordPress/pattern-directory/blob/21e69c4986286c8a0db749999b649b58947456b7/public_html/wp-content/themes/pattern-directory/src/store/resolvers.js#L48-L49
Usually, for a normal case, it would include something like `include[0]=732` in the payload when sending to that endpoint and gets the response with the pattern data of patternId 732.
And since in the case mentioned above, `include=[]` is filtered out from payload, the endpoint response would still return with `our suggested favorite patterns` for a certain category and the patterns would be stored in `state.patterns.queries`.
image
image
image
Since there are favorite patterns returned by endpoint and stored into state, it would make the number of patterns in a category wrong.
https://github.com/WordPress/pattern-directory/blob/21e69c4986286c8a0db749999b649b58947456b7/public_html/wp-content/themes/pattern-directory/src/components/context-bar/index.js#L74
https://github.com/WordPress/pattern-directory/blob/21e69c4986286c8a0db749999b649b58947456b7/public_html/wp-content/themes/pattern-directory/src/store/selectors.js#L60-L63
image

This issue could be solved by making it still send `include` in payload with value like `[-1]` to make it return `[]`, which is definitely a bad idea and confusing solution that needs some comments. 

Or we can modify the logic of the endpoint and the `addQueryArgs` instead to make it send something legit when `include=[]` from client to the endpoint and have some corresponding process to it. And this is where I found it difficult for the moment to locate where and how the query string params `pattern-categories` and `include` are processed. Might need more studies and some clarifying for my doubts on my side if we choose this solution at the end of the day.

**Sol 2** 
Add a condition check for `query.include` in `isLoadingPatternByQuery`.
image
It's not that ideal as it's kind of specifically for a certain case and might need some comments.
But it probably is safe for the moment to my knowledge as the `include=[favorite patternIds]` is only used and included by `MyFavorites`.

**Sol 3** 
[Pass isEmpty as isFavoritesEmpty to ContextBar](https://github.com/WordPress/pattern-directory/commit/960a80353e2b978647645057d30c52d32204bb1e)
It's also not that ideal as far as I'm concerned as it's a bit not clean and consistent.
But the code could speak itself.

<hr>

Fixes #362 

### Screencasts
Before:
https://d.pr/v/pTI4lI
After:
https://d.pr/v/pdkMqt

1. Go to Favorites -> All
2. Empty the favorites list
3. Switch to other categories
4. Spinner should not be seen

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js
@@ -55,7 +55,12 @@ const MyFavorites = () => {
 			<QueryMonitor />
 			<div ref={ ref }>
 				{ isLoggedIn && (
-					<PatternGridMenu basePath="/favorites/" query={ query } onNavigation={ onNavigation } />
+					<PatternGridMenu
+						basePath="/favorites/"
+						query={ query }
+						onNavigation={ onNavigation }
+						isEmpty={ isEmpty }
+					/>
 				) }
 			</div>
 			{ ! isLoggedIn || isEmpty ? (
--- a/public_html/wp-content/themes/pattern-directory/src/components/context-bar/index.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/context-bar/index.js
@@ -112,7 +112,7 @@ function ContextBar( props ) {
 
 	const spinnerClassName = classnames( {
 		'context-bar__spinner': true,
-		'context-bar__spinner--is-hidden': ! isLoadingPatterns,
+		'context-bar__spinner--is-hidden': ! isLoadingPatterns || props.isEmpty,
 	} );
 
 	return (


public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js
```diff
@@ -55,7 +55,12 @@ const MyFavorites = () => {
 			<QueryMonitor />
 			<div ref={ ref }>
 				{ isLoggedIn && (
-					<PatternGridMenu basePath="/favorites/" query={ query } onNavigation={ onNavigation } />
+					<PatternGridMenu
+						basePath="/favorites/"
+						query={ query }
+						onNavigation={ onNavigation }
+						isFavoritesEmpty={ isEmpty }
```
ryelle: Is there a reason this prop name should reference "favorites"? I think we could use `isEmpty={ isEmpty }`, in case we have other cases in the future with empty queries.

ryelle (commented):
Thanks for the detailed write-up! I think the solution you landed on is the best balance of simplicity & functionality 🙂 

I have one bit of feedback on the prop name, but this is working well in my testing 👍🏻

StevenDufresne:
Well explained!

I'm a bit hesitant introducing `favorite` specific logic in the `ContextBar` seeing that it's used in contexts where favorites don't exist, at least in the same way. Additionally if we were to drop the `favorite` from the prop, it isn't apparent as to why loading should stop because something is empty.

Instead of passing the props down into the `ContextBar` is there a way to tell if all requests have resolved?

public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js
outdoor2kode: I think the reason is related to Steve's concern in his comment - the idea of `favorites` doesn't exist in `ContextBar`.

So when I passed the `favorites` down from `MyFavorites` to `ContextBar`, I thought it would be a bit confused to only see `isEmpty` in `ContextBar` as if I were a newcomer it might take some time to figure out where the `isEmtpy` comes from or what it refers to and therefore I made it `isFavoritesEmpty` in `ContextBar`.

That being said, I was thinking that the `isEmpty` inside the components `MyFavorites` would only indicate one thing which is that the favorites list is empty. I couldn't come up with other cases for the moment that something else is also empty and would use the same variable `isEmpty` as well. Could you share some thoughts on this if there are any, thanks!

https://github.com/WordPress/pattern-directory/blob/21e69c4986286c8a0db749999b649b58947456b7/public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js#L37-L41

And if it does exist cases that would use the same `isEmpty` to indicate their empty, do you think it makes sense to keep it as `isFavoritesEmpty` for the time being for the reason mentioned above -  the idea of `favorites` doesn't exist in `ContextBar`, and then we can change it to something else once certain cases really happen sometime in the future.

outdoor2kode:
> Well explained!
> 
> I'm a bit hesitant introducing `favorite` specific logic in the `ContextBar` seeing that it's used in contexts where favorites don't exist, at least in the same way. Additionally if we were to drop the `favorite` from the prop, it isn't apparent as to why loading should stop because something is empty.
> 
> Instead of passing the props down into the `ContextBar` is there a way to tell if all requests have resolved?

I'm thinking that `isLoadingPatternsByQuery` from the selector is kind of the way to tell if all requests have been resolved. It would work correctly if we revised the logic with the **Sol 1** mentioned in the PR description except for the number of favorites patterns would go wrong as the `include=[]` is removed from the request and the endpoint doesn't response as expected.

About not passing props down to `ContextBar` to make it looks weird in it, I'm considering that maybe we could make `ContextBar` simpler and general by getting rid of `useSelect` inside it. Let its parent components pass down the data and decide its behavior.

Some perks of the change:
1. Some selectors will only be called one time.
2. The data is controlled and stored in the parent component. - know where to find the data and how it's retrieved.
3. We could still pass down props with the name like `isFavoritesEmpty` and the `ContextBar` would only consider it just as its required props say, `isLoading`. - the idea of `favorite` would not exist in `ContextBar` then.

Drawback:
1. Parent component would be messier.

StevenDufresne:
> I'm thinking that isLoadingPatternsByQuery from the selector is kind of the way to tell if all requests have been resolved. It would work correctly if we revised the logic with the Sol 1 mentioned in the PR description except for the number of favorites patterns would go wrong as the include=[] is removed from the request and the endpoint doesn't response as expected.

Yes, but I was also thinking something within [ `apiFetch`](https://github.com/WordPress/gutenberg/blob/cecd7828d1adc8ba5c3a33b1e38ebcad4131c22f/packages/data-controls/README.md), does it track it somehow? Probably not.


> About not passing props down to `ContextBar` to make it looks weird in it, I'm considering that maybe we could make `ContextBar` simpler and general by getting rid of `useSelect` inside it. Let its parent components pass down the data and decide its behavior.

Yep, that's sounds fine. If we do use a Parent component do we need to make any other changes? Will [`PatternGridMenu`](https://github.com/WordPress/pattern-directory/blob/26499c15913d96e4b9b6e182ee9573472d49e7a3/public_html/wp-content/themes/pattern-directory/src/components/pattern-grid-menu/index.js#L70) which is shared still work or will that need some refactoring?

outdoor2kode:
> Yep, that's sounds fine. If we do use a Parent component do we need to make any other changes? Will [`PatternGridMenu`](https://github.com/WordPress/pattern-directory/blob/26499c15913d96e4b9b6e182ee9573472d49e7a3/public_html/wp-content/themes/pattern-directory/src/components/pattern-grid-menu/index.js#L70) which is shared still work or will that need some refactoring?

I'm tempted to consider [`PatternGridMenu`](https://github.com/WordPress/pattern-directory/blob/26499c15913d96e4b9b6e182ee9573472d49e7a3/public_html/wp-content/themes/pattern-directory/src/components/pattern-grid-menu/index.js#L70) as a general and common component as well, and the changes would be required on both [MyFavorites](https://github.com/WordPress/pattern-directory/blob/26499c15913d96e4b9b6e182ee9573472d49e7a3/public_html/wp-content/themes/pattern-directory/src/components/my-favorites/index.js#L19) and [Patterns](https://github.com/WordPress/pattern-directory/blob/26499c15913d96e4b9b6e182ee9573472d49e7a3/public_html/wp-content/themes/pattern-directory/src/components/pattern/index.js#L20) where I think are the suitable places to fetch components' related data and then pass down. (i.e., All required favorite data is fetched in Myfavorites, and required pattern data is fetched in Pattern)

PR #432, opened 2022-03-11
Add logging for patterns and flags

### LINKED ISSUE #57: Log changes to patterns
There are probably several reasons why we'd want to be able to audit changes made to a pattern. One is so we can see who removed a pattern from the directory, when they did it, and why, if they include a reason.

A log entry should have a date, user name, type of change, and an optional comment. It might look something like this:

| Date | User | Action | Comment |
| --- | --- | --- | --- |
| 2021-03-30 11:05:05 | matt | Submitted | |
| 2021-03-30 11:05:45 | patternbot | Flagged for manual review | Possible copyright issue |
| 2021-03-30 15:21:03 | admin | Approved | Looks ok |
| 2021-03-30 15:22:02 | someotheruser | Flagged for manual review | Inappropriate content |
| 2021-03-30 16:38:55 | admin | Removed | Content is inappropriate |

Log entries could be added both by automated processes (by "patternbot" or something) or by user-driven actions. In some cases the comment field could get populated automatically with pre-determined values. In other cases, like when a moderator is manually changing the status of a pattern, a textarea field should be available to add a reason, which would then get added to the log's comment field.

coreymckrill:
This enables the [new logging feature](https://github.com/WordPress/wporg-internal-notes/pull/9) in the Internal Notes plugin for the Pattern post type, and adds some custom logging for when a pattern gets flagged and when flags are managed by moderators.

This wraps up #57, albeit in a slightly different way than originally imagined. There are probably still some workflow improvements that could be made around reviewing flags, but this is an MVP, and we can iterate once we have a better idea of how pattern moderation is going to play out.

Fixes #57

1. Create a new pattern. Then edit the pattern.
2. View the pattern on the front end. Click the "Report this pattern" button and submit a flag.
3. In WP Admin, go to Block Pattern > All Flags. Mark the flag as resolved.
4. Now view the Edit screen in WP Admin for the pattern. Open the Notes sidebar. All of your previous activity should be recorded in log entries.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/bootstrap.php
+++ b/public_html/wp-content/plugins/pattern-directory/bootstrap.php
@@ -10,6 +10,7 @@ namespace WordPressdotorg\Pattern_Directory;
 
 require_once __DIR__ . '/includes/class-rest-flags-controller.php';
 require_once __DIR__ . '/includes/class-rest-favorite-controller.php';
+require_once __DIR__ . '/includes/logging.php';
 require_once __DIR__ . '/includes/pattern-post-type.php';
 require_once __DIR__ . '/includes/pattern-flag-post-type.php';
 require_once __DIR__ . '/includes/pattern-validation.php';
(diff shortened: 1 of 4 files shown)


ryelle (approved):
Looks good, I see all the post change logs & the flag messages 👍🏻

PR #436, opened 2022-03-17
Add a setting to change the flag threshold

### LINKED ISSUE #425: Pattern flags: Make the "unpublish" threshold a setting field
The "unpublish" threshold is the number of flags a pattern can receive before its status is set back to "pending" so that it will no longer appear in the pattern directory until a moderator has reviewed it. This number is currently [hard-coded](https://github.com/WordPress/pattern-directory/blob/d7f27feb58fcdcf3f9db796b7851f6b2da72988c/public_html/wp-content/plugins/pattern-directory/includes/pattern-flag-post-type.php#L114-L147) at `3` (I told @angelasjin it was `10` but I was mistaken).

Having this number be adjustable from WP Admin would make it easier for moderators to respond if a situation arises where either too many patterns are getting auto-unpublished or too few are. The setting could live on the same page where the "initial status" setting is.

coreymckrill:
Adds a setting field to the existing Block Pattern Settings page to adjust the number of times a pattern can be reported before it is automatically unpublished.

This also moves the code that creates the Block Pattern Settings page from the pattern-creator plugin over to the pattern-directory plugin, as discussed [here](https://github.com/WordPress/pattern-directory/issues/425#issuecomment-1069766718).

Fixes #425 

### Screenshot

pattern-settings

1. Change the flag threshold setting to `1`. View a published pattern on the front end and report it. The pattern should now be set to "pending".
2. Change the flag threshold back to `5`. Now if you report a pattern, it won't be set to "pending" (until 4 other users also report it).


public_html/wp-content/plugins/pattern-directory/includes/pattern-flag-post-type.php
```diff
@@ -136,12 +136,12 @@ function check_flag_threshold( $post_ID, $post, $update ) {
 		'post_status' => PENDING_STATUS,
 	) );
 
-	$threshold = 3;
+	$threshold = absint( get_option( 'wporg-pattern-flag_threshold' ) );
```
ryelle:
```suggestion
	$threshold = absint( get_option( 'wporg-pattern-flag_threshold', 5 ) );
```

I'm not sure if the default is set by the register_setting, but just in case :)

ryelle (approved):
Works great, and thanks for cleaning up the files 😁

Issue #437, opened 2022-03-17
Declined patterns always show default reason


ryelle:
Declined patterns have a "learn more" button in the notice, which opens a modal with more details. This is supposed to have a reason based on the `wpop_unlisted_reason`, chosen when the pattern is unlisted, but it uses [the reason slug](https://github.com/WordPress/pattern-directory/blob/6cea384136a1948150ba75b714ea4405e394bc41/public_html/wp-content/themes/pattern-directory/src/components/pattern/status-notice.js#L8) while the actual meta value is the term ID.

This means all unlisted patterns are showing the default reason:

<img width="388" alt="Screen Shot 2022-03-17 at 4 01 50 PM" src="https://user-images.githubusercontent.com/541093/158893824-88503914-7daf-47d1-99ab-8efc0b60c692.png">

See #446, the method of saving reason might change

PR #438, opened 2022-03-17
Pattern Moderation: Add quick links & bulk edit actions to manage pattern status

ryelle:
When managing submitted patterns, there's no good way to quickly publish pending patterns. Similarly, we might want to mark patterns as spam, or quickly unlist spam patterns. Using the "quick edit" has the side-effect of changing the author, and opening the pattern in the editor takes a while. This PR uses the same logic as the flags row-actions, combining row-action handlers with the bulk-edit dropdown, so we can use both for faster moderation.

The updates to row actions are:

- Remove quick-edit, since it has the author-change effect
- Add "Publish" if the pattern is pending (new submission) or spam (in case it's incorrectly flagged)
- Add "Unlist" if the pattern is spam. If the pattern needs to be unlisted for another reason (broken, copyright, etc) the Unlist button in the editor should be used so that a reason can be added.
- Add "Spam" if the pattern is not already spam.

### Screenshots

On the frontend, since these unlisted patters have no "reason", they get the default reason of "Additional review has been requested for this pattern." (Actually, there's a deeper issue here, but it's not related - see #437).

0. Add a new reason to the reasons taxonomy, `Spam`, with slug `4-spam`, and update the Other's slug to `5-other`
1. View the Block Patterns list table
2. Update some patterns by clicking the row actions & bulk actions
3. They should behave as expected
4. If you check the post meta for a pattern unlisted from bulk actions, the `wpop_unlisted_reason` should be the term ID of your new Spam term

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/admin-patterns.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/admin-patterns.php
@@ -20,6 +20,9 @@ add_action( 'manage_posts_extra_tablenav', __NAMESPACE__ . '\pattern_list_table_
 add_filter( 'views_edit-' . PATTERN, __NAMESPACE__ . '\pattern_list_table_views' );
 add_action( 'pre_get_posts', __NAMESPACE__ . '\handle_pattern_list_table_views' );
 add_filter( 'display_post_states', __NAMESPACE__ . '\display_post_states', 10, 2 );
+add_filter( 'post_row_actions', __NAMESPACE__ . '\add_row_actions', 10, 2 );
+add_filter( 'bulk_actions-edit-' . PATTERN, __NAMESPACE__ . '\add_bulk_actions' );
+add_filter( 'handle_bulk_actions-edit-' . PATTERN, __NAMESPACE__ . '\handle_bulk_actions', 10, 3 );
 
 /**
  * Modify the patterns list table columns.
@@ -212,7 +215,7 @@ function pattern_list_table_styles( $which ) {
 			text-align: center;
 		}
 
-		.column-language {
+		td.column-language {
 			display: flex;
 			flex-direction: column;
 		}
@@ -226,6 +229,16 @@ function pattern_list_table_styles( $which ) {
 		.language-context-link {
 			font-size: 0.85em;
 			line-height: 1.3;
+		}
+
+		.row-actions span.spam a,
+		.row-actions span.unlist a {
+			color: #bd8600;
+		}
+
+		.row-actions span.spam a:hover,
+		.row-actions span.unlist a:hover {
+			color: #996800;
 		}
 	</style>
 	<?php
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/pattern-directory/includes/admin-patterns.php
```diff
@@ -391,3 +405,147 @@ function display_post_states( $post_states, $post ) {
...
+		return $actions;
+	}
+
+	$saved_actions = array_intersect_key( $actions, array_fill_keys( array( 'trash', 'untrash', 'delete' ), true ) );
+	$actions       = array_intersect_key( $actions, array_fill_keys( array( 'edit', 'view' ), true ) );
+
+	$edit_url = add_query_arg( 'post_type', PATTERN, 'edit.php' );
+	$title = _draft_or_post_title();
+
+	if ( 'pending' === $post->post_status || SPAM_STATUS === $post->post_status ) {
```
coreymckrill: For consistency with using constants, `'pending'` could be `PENDING_STATUS` instead. (You'd also have to import it at the top.)

public_html/wp-content/plugins/pattern-directory/includes/admin-patterns.php
```diff
@@ -391,3 +405,147 @@ function display_post_states( $post_states, $post ) {
...
+		$spam_url = add_query_arg(
```
coreymckrill:
```suggestion
		$publish_url = add_query_arg(
```

public_html/wp-content/plugins/pattern-directory/includes/admin-patterns.php
```diff
@@ -391,3 +405,147 @@ function display_post_states( $post_states, $post ) {
...
+			array(
+				'action' => 'publish',
+				'post'   => array( $post->ID ),
+			),
+			wp_nonce_url( $edit_url, 'bulk-posts' )
+		);
+
+		$actions['publish'] = sprintf(
+			'<a href="%s" aria-label="%s">%s</a>',
+			$spam_url,
```
coreymckrill:
```suggestion
			$publish_url,
```

coreymckrill (commented):
👍 Works on local, although I'm getting this when I use one of the post row actions: `Notice: Array to string conversion in /var/www/html/wp-content/plugins/wporg-internal-notes/includes/logging.php on line 69`. I haven't been able to figure out yet why that's happening. It might not be related to this PR...

One comment I have is that when a pattern is marked as "possible spam" it's not clear if it's still visible in the directory. I'm guessing no, since "Publish" is the way you unspam it. But since you can also mark a spammed pattern as "unlisted", it's a bit confusing.

Also, is there any scenario where you'd want to mark something as "not spam" but have it go to Pending instead of Publish?

ryelle:
> Works on local, although I'm getting this when I use one of the post row actions: Notice: Array to string conversion in /var/www/html/wp-content/plugins/wporg-internal-notes/includes/logging.php on line 69. I haven't been able to figure out yet why that's happening. It might not be related to this PR...

Yeah, it's an issue with the Internal Notes - I added an issue with some more info: https://github.com/WordPress/wporg-internal-notes/issues/10

> One comment I have is that when a pattern is marked as "possible spam" it's not clear if it's still visible in the directory. I'm guessing no, since "Publish" is the way you unspam it. But since you can also mark a spammed pattern as "unlisted", it's a bit confusing.

It's not visible in the directory, but it is visible to the pattern author as "Pending". Once it's unlisted, it's "Declined" in the author's pattern list. I don't think of the "possible spam" status as a place things should stay, ideally they'd either be unspammed or officially declined. Do you think there's a reason to keep them in this status?

> is there any scenario where you'd want to mark something as "not spam" but have it go to Pending instead of Publish?

I don't think so— if we get to this state, the pattern author thought it was done enough to submit. Pending would just need to be reviewed, probably by the same person reviewing the spam.

coreymckrill:
Ok, confirmed that the PHP notice is a problem in the Internal Notes plugin, has nothing to do with this 😅

coreymckrill:
> I don't think of the "possible spam" status as a place things should stay, ideally they'd either be unspammed or officially declined. Do you think there's a reason to keep them in this status?

🤔 I guess not. Unless we wanted to have some kind of statistics about how many patterns are spam. But there might be other ways to do that besides using post status.

coreymckrill (approved):

ryelle:
> Unless we wanted to have some kind of statistics about how many patterns are spam. But there might be other ways to do that besides using post status.

Things that are unlisted with the reason "spam" should get that value.

PR #439, opened 2022-03-17
Bin: Add a script to run spam checks over the patterns

ryelle:
For a little while, patterns that were submitted as pending were not getting the spam checks applied. I wanted to quickly check the handful of patterns that were submitted, but there's no easy way to manually run the spam check. This adds a bin script (so it requires a sandbox) to run the checks over all patterns, with a few query params.

For example, to run over all draft patterns:

```
$ php ./bin/check-spam.php --all --post_status=draft
Dry run, will not update any patterns.
19 patterns checked, 3 found to be spam.
```

That command won't make any changes, it just reports totals. Adding `--verbose` gives more details:

```
$ php ./bin/check-spam.php --all --post_status=draft --verbose
Dry run, will not update any patterns.
1: Not spam.
2: Not spam.
3: Spam found: Only contains Paragraph blocks.
...
19 patterns checked, 3 found to be spam.
```

Still no changes. To update the status & flag as possible spam, run with `--apply`:

```
$ php ./bin/check-spam.php --all --post_status=draft --verbose
Dry run, will not update any patterns.
1: Not spam.
2: Not spam.
3: Spam found: Only contains Paragraph blocks.
3: Post status updated.
...
19 patterns checked, 3 found to be spam.
```

Other possible arguments are `--post=<ID>` (instead of `--all`, will override it), or `--per_page=<count>`.

1. Run the script
2. It should correctly detect spam based on the validation
3. If you apply the changes, it should update the status of any spammy pattern
4. Try submitting a spam pattern via the editor (use the keyword `PatternDirectorySpamTest`) 
5. It should still flag that as spam

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/bin/check-spam.php
+++ b/public_html/wp-content/plugins/pattern-directory/bin/check-spam.php
@@ -1,0 +1,113 @@
+<?php
+
+namespace WordPressdotorg\Pattern_Directory;
+
+use function WordPressdotorg\Pattern_Directory\Pattern_Validation\check_for_spam;
+use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\{ POST_TYPE, SPAM_STATUS };
+
+// This script should only be called in a CLI environment.
+if ( 'cli' != php_sapi_name() ) {
+	die();
+}
+
+$opts = getopt( '', array( 'post:', 'url:', 'abspath:', 'post_status:', 'per_page:', 'all', 'apply', 'verbose' ) );
+
+if ( empty( $opts['url'] ) ) {
+	$opts['url'] = 'https://wordpress.org/patterns/';
+}
+
+if ( empty( $opts['abspath'] ) && false !== strpos( __DIR__, 'wp-content' ) ) {
+	$opts['abspath'] = substr( __DIR__, 0, strpos( __DIR__, 'wp-content' ) );
+}
+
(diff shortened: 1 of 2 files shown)


dd32 (approved):
While I've not run it, reading the diff, I don't see anything that sticks out to me as no-good.

coreymckrill:
If a pattern gets flagged as spam, but then is determined during a manual review to _not_ be spam and set back to publish, is there anything to prevent it from getting re-spammed in the next scan?

ryelle:
> If a pattern gets flagged as spam, but then is determined during a manual review to not be spam and set back to publish, is there anything to prevent it from getting re-spammed in the next scan?

Hm, I guess not 🤔 

I don't really see this being run on published patterns though - the immediate use case for me was that I wanted to run it over the ~10 pending patterns that were submitted between opening the pattern creator & adding the auto-check to pending patterns. Honestly I'm not sure it will _need_ to be run again, but I'd written it so I thought I'd commit it. I could swap out [the default status](https://github.com/WordPress/pattern-directory/blob/d44ac6dce9226c70bd80a461d027e8bce5b81c4d/public_html/wp-content/plugins/pattern-directory/bin/check-spam.php#L23) to `pending` & `draft` if that would make more sense?

coreymckrill:
Yeah, keeping it away from published patterns seems like a good idea, but probably not a big deal either way.

> Honestly I'm not sure it will need to be run again

Famous last words 😄

PR #448, opened 2022-03-21
Pattern moderation: Store unlisted reason as tax term instead of postmeta

### LINKED ISSUE #446: Pattern unlisted reason saved as meta value
The reasons are a taxonomy though, so we could put them into the tax terms instead. I think this wasn't done because the meta value method was already in place when the taxonomy was created. Might help with #437 .

coreymckrill:
Changes how a reason for unlisting a pattern is stored, from postmeta to taxonomy term, using the same registered taxonomy that the Pattern Flag post type uses.

Fixes #446 

1. Edit a pattern in WP Admin. There should now be a sidebar section for the Flag Reasons taxonomy.
2. Unlist the pattern and fill out the modal form. Upon completing the form, the chosen reason should get a check next to it in the Flag Reasons section of the sidebar.
3. (If you try to re-publish the pattern here, you might get an error saying the status must be set to `pending-review`. After refreshing, the pattern will indeed be set to spam instead of publish. Haven't figured this out yet.)
4. From the Patterns list table, try unlisting one or more patterns using the row actions or bulk edit. Afterward, if you edit one of those patterns, the Flag Reason should be set to spam.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -432,8 +432,8 @@ function register_post_statuses() {
 /**
  * Do things when certain status transitions happen.
  *
- * @param string $new_status
- * @param string $old_status
+ * @param string   $new_status
+ * @param string   $old_status
  * @param \WP_Post $post
  *
  * @return void
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -5,7 +5,7 @@ namespace WordPressdotorg\Pattern_Directory\Pattern_Post_Type;
 use Error, WP_Block_Type_Registry;
 use function WordPressdotorg\Locales\{ get_locales, get_locales_with_english_names, get_locales_with_native_names };
 use function WordPressdotorg\Pattern_Directory\Favorite\get_favorite_count;
-use const WordPressdotorg\Pattern_Directory\Pattern_Flag_Post_Type\TAX_TYPE;
+use const WordPressdotorg\Pattern_Directory\Pattern_Flag_Post_Type\TAX_TYPE as FLAG_REASON;
 
 const POST_TYPE       = 'wporg-pattern';
 const UNLISTED_STATUS = 'unlisted';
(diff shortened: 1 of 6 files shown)


ryelle:
> If you try to re-publish the pattern here, you might get an error saying the status must be set to pending-review. After refreshing, the pattern will indeed be set to spam instead of publish. Haven't figured this out yet.

Ah, that's from the pattern status validation, to prevent submissions from being set to the wrong status. Maybe unhook that validation in wp-admin? Though why it _does_ update the status despite erroring is unexpected.

https://github.com/WordPress/pattern-directory/blob/3461470a6f7620bba996899cafe6af506bae59a0/public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php#L11

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
```diff
@@ -5,6 +5,7 @@
 use Error, WP_Block_Type_Registry;
 use function WordPressdotorg\Locales\{ get_locales, get_locales_with_english_names, get_locales_with_native_names };
 use function WordPressdotorg\Pattern_Directory\Favorite\get_favorite_count;
+use const WordPressdotorg\Pattern_Directory\Pattern_Flag_Post_Type\TAX_TYPE;
```
ryelle: Could this be renamed on import so it's easier to remember which taxonomy it is?

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
```diff
@@ -427,6 +429,26 @@ function register_post_statuses() {
...
+ *
+ * @return void
+ */
+function status_transitions( $new_status, $old_status, $post ) {
+	if ( POST_TYPE !== get_post_type( $post ) ) {
+		return;
+	}
+
+	// If a pattern gets relisted, remove the reason that it was originally unlisted.
+	if ( UNLISTED_STATUS === $old_status && in_array( $new_status, array( 'publish', 'pending' ), true ) ) {
```
ryelle: I think you could say `&& UNLISTED_STATUS !== $new_status` here, since even if it moves to draft or possible spam, we don't need to keep the unlisted reason.

ryelle (approved):
Works great - I tried unlisting a pattern and it assigned the correct term; I then published the pattern and it cleared the term 👍🏻 

I just have two small suggestions :)

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
coreymckrill: Sure, changed it to `FLAG_REASON` in 3ce923e

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
coreymckrill: 👍 3ce923e

PR #451, opened 2022-03-22
Send email notifications to pattern authors when certain status changes happen

### LINKED ISSUE #58: Email notifications about pattern changes
When the status of a pattern changes (#56), and possibly in some other cases, the pattern author should receive an email notification about the change. This serves to ensure they are informed when e.g. a pattern is removed from the directory by a moderator for some reason, and also to alert them to possible unauthorized access to their account.

coreymckrill:
Hooks into `transition_post_status` to detect when pattern status changes happen, which will mostly be due to moderators and flags.

Fixes #58 

1. Upload this changeset to your wporg sandbox so you can test whether the emails actually get sent.
2. Create a test pattern.
3. If it's clearly test content, it might get immediately marked as spam (like mine did). In which case you should receive an email telling you as much.
4. From the Patterns list table, try switching the pattern's status between publish / spam / unlisted. Each time you should get an email.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/notifications.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/notifications.php
@@ -15,8 +15,8 @@ add_action( 'transition_post_status', __NAMESPACE__ . '\monitor_post_status_tran
 /**
  * Hook into post status transitions to detect changes that warrant a notification.
  *
- * @param string $new_status
- * @param string $old_status
+ * @param string   $new_status
+ * @param string   $old_status
  * @param \WP_Post $post
  *
  * @return void
@@ -195,7 +195,15 @@ If you would like to resubmit your pattern, please make sure it follows the guid
 	send_email( $email, $subject, $message );
 }
 
-
+/**
+ * Wrapper for wp_mail.
+ *
+ * @param string $to
+ * @param string $subject
+ * @param string $message
+ *
+ * @return void
+ */
 function send_email( $to, $subject, $message ) {
 	$message = html_entity_decode( $message );
 
--- a/public_html/wp-content/plugins/pattern-directory/includes/notifications.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/notifications.php
@@ -2,6 +2,7 @@
 
 namespace WordPressdotorg\Pattern_Directory\Notifications;
 
+use function WordPressdotorg\Pattern_Directory\Pattern_Flag_Post_Type\get_default_reason_description;
 use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\{ POST_TYPE as PATTERN, UNLISTED_STATUS, SPAM_STATUS };
 use const WordPressdotorg\Pattern_Directory\Pattern_Flag_Post_Type\{ POST_TYPE as FLAG, TAX_TYPE as REASON, PENDING_STATUS };
 
@@ -103,13 +104,14 @@ function notify_pattern_flagged( $post ) {
 		switch_to_locale( $locale );
 	}
 
-	$reasons = array();
 	$flags = get_posts( array(
 		'post_type' => FLAG,
 		'post_parent' => $post->ID,
 		'post_status' => PENDING_STATUS,
 	) );
+	$reason = '';
 	if ( ! empty( $flags ) ) {
+		$reasons = array();
 		foreach ( $flags as $flag ) {
 			$terms = get_the_terms( $flag, REASON );
 			if ( is_array( $terms ) ) {
@@ -118,16 +120,20 @@ function notify_pattern_flagged( $post ) {
 		}
 		$reasons = array_map(
 			function( \WP_Term $reason ) {
-				return $reason->name;
+				return wp_strip_all_tags( $reason->description );
 			},
 			$reasons
 		);
 		$reasons = array_unique( $reasons );
-		// translators: used between list items, there is a space after the comma.
-		$reasons = implode( __( ', ', 'wporg-patterns' ), $reasons );
+		$reason = trim( implode( "\n", $reasons ) );
 	} else {
 		// If it doesn't have flags, it must have gotten here by getting marked as spam.
-		$reasons = __( 'Spam', 'wporg-patterns' );
+		$spam_term = get_term_by( 'slug', '4-spam' );
+		$reason = wp_strip_all_tags( $spam_term->description );
+	}
+
+	if ( ! $reason ) {
+		$reason = get_default_reason_description();
 	}
 
 	$subject = esc_html__( 'Pattern being reviewed', 'wporg-patterns' );
@@ -142,7 +148,7 @@ Thanks for submitting your pattern. Unfortunately, your pattern, %1$s, has been 
 
 Your pattern has been unpublished from the Block Pattern Directory at this time, and will receive further review. If the pattern meets the guidelines, we will re-publish it to the Block Pattern Directory. Thanks for your patience with us volunteer reviewers!', 'wporg-patterns' ),
 		esc_html( html_entity_decode( $pattern_title ) ),
-		esc_html( html_entity_decode( $reasons ) )
+		esc_html( html_entity_decode( $reason ) )
 	);
 
 	if ( $locale ) {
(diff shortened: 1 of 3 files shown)


coreymckrill:
Looks like the failing tests are because the test user doesn't have an email address set... will look into it tomorrow.

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
```diff
@@ -0,0 +1,234 @@
...
+	if ( ! empty( $flags ) ) {
+		foreach ( $flags as $flag ) {
+			$terms = get_the_terms( $flag, REASON );
+			if ( is_array( $terms ) ) {
+				$reasons += $terms;
+			}
+		}
+		$reasons = array_map(
+			function( \WP_Term $reason ) {
+				return $reason->name;
```
ryelle: The description might be a more user-friendly thing to show (see #453), but those are full sentences. You could `implode` them with line breaks, maybe?

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
```diff
@@ -0,0 +1,234 @@
...
+
+	$email  = $author->user_email;
+	$locale = get_user_locale( $author );
+
+	$pattern_title = get_the_title( $post );
+
+	$reasons = get_the_terms( $post, REASON );
+	if ( ! empty( $reasons ) ) {
+		$reason_term = reset( $reasons );
+		$reason = $reason_term->name;
```
ryelle: Same comment as above, you can use the `description` here for a better reason.

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
```diff
@@ -0,0 +1,234 @@
...
+Your pattern, %1$s, has been unlisted from the Block Pattern Directory due to the following reason:
+
+%2$s
+
+If you would like to resubmit your pattern, please make sure it follows the guidelines:
+
+%3$s', 'wporg-patterns' ),
+		esc_html( html_entity_decode( $pattern_title ) ),
+		esc_html( html_entity_decode( $reason ) ),
+		'https://wordpress.org/patterns/' // TODO: get the actual URL.
```
ryelle: The URL to the guidelines is just the About page.
```suggestion
		'https://wordpress.org/patterns/about/'
```

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
```diff
@@ -0,0 +1,234 @@
...
+
+Your pattern, %1$s, has been unlisted from the Block Pattern Directory due to the following reason:
+
+%2$s
+
+If you would like to resubmit your pattern, please make sure it follows the guidelines:
+
+%3$s', 'wporg-patterns' ),
+		esc_html( html_entity_decode( $pattern_title ) ),
+		esc_html( html_entity_decode( $reason ) ),
```
ryelle: If you do switch to use the term `description`, it can have HTML in it (see the non-english description). That should be fine if you strip_tags though, if you want to do plaintext emails only.

ryelle (commented):
Some notes about the description field, but otherwise this is working well. I tested on my sandbox like suggested, and got the emails for spam, unlisting, and publish 👍🏻

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
coreymckrill: Good idea: a1dee09

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
coreymckrill: a1dee09

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
coreymckrill: a1dee09

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
```diff
@@ -0,0 +1,243 @@
...
+			function( \WP_Term $reason ) {
+				return wp_strip_all_tags( $reason->description );
+			},
+			$reasons
+		);
+		$reasons = array_unique( $reasons );
+		$reason = trim( implode( "\n", $reasons ) );
+	} else {
+		// If it doesn't have flags, it must have gotten here by getting marked as spam.
+		$spam_term = get_term_by( 'slug', '4-spam' );
```
ryelle: I think you need the tax here. in my test, a spammed pattern sent the default reason, not the spam reason.

```suggestion
		$spam_term = get_term_by( 'slug', '4-spam', REASON );
```

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
```diff
@@ -0,0 +1,243 @@
...
+		$reason_term = reset( $reasons );
+		$reason = wp_strip_all_tags( $reason_term->description );
+	}
+
+	if ( ! $reason ) {
+		$reason = get_default_reason_description();
+	}
+
+	if ( $locale ) {
+		switch_to_locale( $locale );
```
ryelle: Should this `switch_to_locale` happen before we get the term description, so that it can be sent in the user's language too?

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
```diff
@@ -0,0 +1,243 @@
...
+ * Wrapper for wp_mail.
+ *
+ * @param string $to
+ * @param string $subject
+ * @param string $message
+ *
+ * @return void
+ */
+function send_email( $to, $subject, $message ) {
+	$message = html_entity_decode( $message );
```
ryelle: Something is still encoding the apostrophe in the message— the email I got says:

`This pattern doesn&#039;t meet the guidelines for the pattern directory.`

ryelle (commented):
Emails are sending out the descriptions now, and the one with HTML in it still looks good. There's still an issue with encoding on the default message.

When I unlisted a pattern, I got two emails - the first had the default message, then the second was correct.

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
coreymckrill: Bah, I had already committed this fix, but I forgot to push it.

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
coreymckrill: Good catch! ab445b9

public_html/wp-content/plugins/pattern-directory/includes/notifications.php
ocean90: To decode quotes you have to set the `ENT_QUOTES` flag.

```suggestion
	$message = html_entity_decode( $message, ENT_QUOTES );
```

(In PHP 8.1 the default was changed from `ENT_COMPAT` to `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401`.)

coreymckrill: Today I learned. Thanks @ocean90!

coreymckrill:
> When I unlisted a pattern, I got two emails - the first had the default message, then the second was correct.

Turns out this is because I didn't have a check in the `transition_post_status` action to ensure the status was actually changing, so it was getting triggered twice: once for the changed status and once for (I'm guessing) adding the reason term. Unfortunately, with that check [in place](https://github.com/WordPress/pattern-directory/pull/451/commits/6cae9e05415bf656790b8b15a29f38e390a51130), the only one that gets sent is with the default message. So I probably need to rethink which action hook is used to trigger these notifications.

ryelle:
> Unfortunately, with that check [in place](https://github.com/WordPress/pattern-directory/pull/451/commits/6cae9e05415bf656790b8b15a29f38e390a51130), the only one that gets sent is with the default message.

Do you think it's worth merging this for the other emails, and figuring it out after, or would you prefer to hold off until you've figured out the right trigger?

coreymckrill:
I'd be happy to merge this now and open a new issue to get this fixed later.

Issue #371, opened 2021-10-20
Pattern Moderation: Show a user-facing reason for why a pattern was declined


ryelle:
In the modal to learn more if your pattern is unlisted (declined), there should be a string explaining why the pattern was removed. For example:

<img width="498" alt="Screen Shot 2021-10-20 at 2 55 01 PM" src="https://user-images.githubusercontent.com/541093/138154326-bd60824c-9571-4ccf-88c7-1c1bd9433f0b.png">

ryelle:
For copy review, maybe @angelasjin can help, or point me to someone who can?

> WordPress.org has chosen to remove your pattern from the directory for the following reason:

Should this be passive voice instead?

For the reason strings, we have the following "buckets" of reasoning, which can show a generic user-facing explanation. Here are some proposals.

**Rude, crude, or inappropriate**

> This pattern contains content deemed inappropriate for a general audience.

**Copyrighted or trademark issue**

> This pattern contains copyrighted material or uses a trademark without permission.

**Broken or unusable**

> This pattern is broken or does not display correctly.

**Other**

I have no idea for this one, not sure what it would be used for in reality.

angelasjin:
Thanks @ryelle! I think this looks good, I just have some minor edits. 

> WordPress.org has removed your pattern from the directory for the following reason:

The reasoning for **Rude** and **Copyrighted** look good to me.

For **Broken or unusable**, does "does not display correctly" cover "unusable" well enough, or does any further elaboration help? 

For **Other**, what is being flagged here? We could use something generic like:

> This pattern has been flagged for additional review. If you have more questions, please contact _____

ryelle:
> For Broken or unusable, does "does not display correctly" cover "unusable" well enough, or does any further elaboration help?

I thought so, but maybe not? Do you think it needs more explanation?

> For Other, what is being flagged here?

Really anything else, I don't know — I'm going off the flagging reasons when reporting a pattern.

> <i>This pattern has been flagged for additional review. If you have more questions, please contact _____</i>

Who should be contacted?

angelasjin:
For **Broken or unusable**, what you suggested works!

I was thinking through whether we needed the **Other** category, as seems to be a catch all for patterns we don't expect. When a pattern is flagged for **Other**, can a reason be submitted for why it was flagged?

In any event, I like having a catch-all, just in case. Let's keep the explanation real simple, how about: 

> Additional review has been requested for this pattern.

Thanks @ryelle!

[...]

PR #453, opened 2022-03-22
My Pattern: Display the reason on unlisted patterns

### LINKED ISSUE #437: Declined patterns always show default reason

ryelle:
This updates the unlisted modal to display the "reason" description on unlisted patterns. I've copied the existing strings into the terms on w.org/patterns and my local environment:

Screen Shot 2022-03-22 at 1 44 50 PM

Fixes #437.

I also removed the meta registration for `wpop_unlisted_reason`, as this is not used anymore.

### Screenshots

Not in English:

Spam:

No reason set:

0. Add the reason strings to the term descriptions
1. Make sure you have some unlisted patterns with the tax term set — this command is helpful: `yarn wp-env run cli "post term set [POST_ID] wporg-pattern-flag-reason 2-copyright"`
2. View an unlisted pattern that you're author of
3. See the red notice, click the Learn more button
4. The modal should contain 3 paragraphs, the middle italicized one should be the term description

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -229,23 +229,6 @@ function register_post_type_data() {
 			),
 		)
 	);
-
-	register_post_meta(
-		POST_TYPE,
-		'wpop_unlisted_reason',
-		array(
-			'type'              => 'string',
-			'description'       => 'The ID of a flag reason, used to indicate why a pattern was unlisted.',
-			'single'            => true,
-			'sanitize_callback' => 'sanitize_text_field',
-			'auth_callback'     => __NAMESPACE__ . '\can_edit_this_pattern',
-			'show_in_rest'      => array(
-				'schema' => array(
-					'type'     => 'string',
-				),
-			),
-		)
-	);
 }
 
 /**
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -388,6 +388,44 @@ function register_rest_fields() {
 			),
 		)
 	);
+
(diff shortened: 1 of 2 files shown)


ryelle:
> Additional review has been requested for this pattern.

This was the agreed-on fallback for "other" (see #371), but I don't think that's clear - an unlisted pattern won't get more review, it's already unlisted. Maybe this?

> This pattern doesn't meet the guidelines for the pattern directory.

coreymckrill (approved):
👍 Works well locally.

> an unlisted pattern won't get more review, it's already unlisted.
>
>> This pattern doesn't meet the guidelines for the pattern directory.

Yeah, I think this would be better as the hard-coded fallback for when an unlisted pattern doesn't have a reason term. We'd want to maintain the old wording for the "Other" reason term, though, since it's also used when submitting a flag on a pattern.

ryelle:
> We'd want to maintain the old wording for the "Other" reason term, though, since it's also used when submitting a flag on a pattern.

Ah true, and that's used in the email to users now. I'll leave that alone for now, then.

ryelle:
@mrfoxtalbot That's a good point, it's not clear what's happened in that case. I've updated the message, so now you get this:

Screen Shot 2022-07-18 at 5 38 44 PM
 
> Additionally, I would include a link to My Patterns just to make it easier for users to locate the pattern they created.

Where do you think that link is needed?

PR #454, opened 2022-03-22
Stats: Add spam statuses to the stats records

ryelle:
This adds a count of the "possibly spam" and "unlisted for spam" patterns to the stats snapshot records.

I haven't added stats for the other unlisted reasons, though we could. I'm not sure if that would be useful, or too much information.

1. View the table in `/wp-admin/edit.php?post_type=wporg-pattern&page=wporg-pattern-stats`
2. It should show `count-patterns_possible-spam` and `count-patterns_unlisted-spam`, and these values should be accurate
3. Download the CSV — you might need to generate it with `yarn wp-env run cli "cron event run wporg-pattern_record_snapshot"`
4. It should have values for `count-patterns_possible-spam` and `count-patterns_unlisted-spam`, and the `version` should be `2`. Earlier rows in the sheet should have empty spaces for these new properties.


coreymckrill (approved):
👍 Looks good on the wporg sandbox.

> I haven't added stats for the other unlisted reasons, though we could. I'm not sure if that would be useful, or too much information.

I could imagine specific situations where we'd want to know, say, how many patterns have been taken down for copyright violations. But I think we can cross that bridge when we come to it.

PR #455, opened 2022-03-22
Validation: Avoid status validation if the user is a moderator or admin

### LINKED ISSUE #450: Unlisted patterns: Can't re-publish from the WP Admin post edit screen
While working on #448 I noticed that (in local dev at least) if you try to publish a pattern that is currently unlisted you get some weird behavior:

* First it will update the post status to `pending-review` i.e. Spam. But the "Are you ready to publish?" sidebar stays open.
* If you then click the Publish button again, you can an error saying the status must be `pending-review`.

It looks like the error message part can be resolved by changing [this permission check](https://github.com/WordPress/pattern-directory/blob/trunk/public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php#L212). `edit_others_patterns` is not actually a valid raw cap for the pattern post type, because it doesn't register any custom caps. So it should be

```php
! current_user_can( $post_type->cap->edit_others_posts )
```

or perhaps better yet would be

```php
! current_user_can( 'edit_post', $prepared_post->ID )
```

For the first problem, though, I can't figure out why hitting the Publish button is switching the post status to Spam...

ryelle:
Fixes #450 — Skip the status validation entirely if the user can edit others' posts (patterns). This will allow moderators to publish patterns regardless of the setting for "Default status of new patterns".

With an editor or admin user…

1. Set the default status to "pending"
2. Try to publish an unlisted pattern from within wp-admin
3. The pattern should correctly be set to "publish" with no error

With a subscriber…

1. Submit a new pattern
2. It should submit as the expected default status
3. Try to change it via an API request, it should fail with the validation error

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
@@ -190,6 +190,11 @@ function validate_status( $prepared_post, $request ) {
 		return $prepared_post;
 	}
 
+	// Skip validation if the user is a moderator.
+	if ( current_user_can( $post_type->cap->edit_others_posts ) ) {
+		return $prepared_post;
+	}
+
 	$default_status = get_option( 'wporg-pattern-default_status', 'publish' );
 	$valid_states   = array_unique( array( 'pending', SPAM_STATUS, $default_status ) );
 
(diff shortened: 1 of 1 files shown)


coreymckrill (approved):
👍

PR #462, opened 2022-03-24
Fix the missing locale name in title

### LINKED ISSUE #404: Title management should not remove locale name
Patterns URLs like https://fr.wordpress.org/patterns/ (correctly) append a localization string to the `<title>` elem, but this is removed via JavaScript.

This behaviour should be removed; the title should not be altered.

outdoor2kode:
This PR
1. Restructure `wporgLocale` to include more locale info:
    a. locale base id
    b. locale display name
2. Fix the logic affected by the restruction
3. Add functions to get the rosetta name (locale display name)

Fixes #404 

### Screencasts
Before:
https://d.pr/v/466P0M
After:
https://d.pr/v/2SsuBI

1. Go to any rosetta site
2. Open Chrome devTools
3. Search for <title>
4. Check if the locale name is appended

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -4,6 +4,7 @@ namespace WordPressdotorg\Pattern_Directory\Theme;
 
 use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\POST_TYPE;
 use const WordPressdotorg\Pattern_Directory\Pattern_Flag_Post_Type\POST_TYPE as FLAG_POST_TYPE;
+use function WordPressdotorg\MU_Plugins\Global_Header_Footer\{ is_rosetta_site, get_rosetta_name };
 
 require_once __DIR__ . '/includes/inline-styles.php';
 
@@ -80,7 +81,13 @@ function enqueue_assets() {
 
 		wp_add_inline_script(
 			'wporg-pattern-script',
-			sprintf( "var wporgLocale = '%s';", wp_json_encode( get_locale() ) ),
+			sprintf(
+				"var wporgLocale = JSON.parse( decodeURIComponent( '%s' ) )",
+				rawurlencode( wp_json_encode( array(
+					'id' => get_locale(),
+					'displayName' => is_rosetta_site() ? get_rosetta_name() : '',
+				) ) ),
+			),
 			'before'
 		);
 
--- a/public_html/wp-content/themes/pattern-directory/src/components/document-title-monitor/index.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/document-title-monitor/index.js
@@ -48,7 +48,7 @@ const DocumentTitleMonitor = () => {
 			);
 		}
 
-		parts.push( __( 'WordPress.org', 'wporg-patterns' ) );
+		parts.push( __( 'WordPress.org', 'wporg-patterns' ) + ` ${ wporgLocale.displayName }` );
 		return parts.join( ' | ' );
 	} );
 
--- a/public_html/wp-content/themes/pattern-directory/src/store/resolvers.js
+++ b/public_html/wp-content/themes/pattern-directory/src/store/resolvers.js
@@ -46,7 +46,7 @@ export function* getPatternsByQuery( query ) {
 	try {
 		yield fetchPatterns( queryString );
 		const response = yield apiFetch( {
-			path: addQueryArgs( '/wp/v2/wporg-pattern', { ...query, locale: JSON.parse( wporgLocale ) } ),
+			path: addQueryArgs( '/wp/v2/wporg-pattern', { ...query, locale: wporgLocale.id } ),
 			parse: false,
 		} );
 		const { total, totalPages, results } = yield __unstableAwaitPromise( parseResponse( response ) );
--- a/public_html/wp-content/themes/pattern-directory/src/store/test/resolvers.js
+++ b/public_html/wp-content/themes/pattern-directory/src/store/test/resolvers.js
@@ -7,7 +7,7 @@ import apiPatternFlagReasons from './fixtures/pattern-flag-reasons';
 import { getCategories, getFavorites, getPattern, getPatternFlagReasons, getPatternsByQuery } from '../resolvers';
 
 // Set up the global.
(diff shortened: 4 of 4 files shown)


public_html/wp-content/themes/pattern-directory/functions.php
```diff
@@ -80,7 +106,13 @@ function enqueue_assets() {
 
 		wp_add_inline_script(
 			'wporg-pattern-script',
-			sprintf( "var wporgLocale = '%s';", wp_json_encode( get_locale() ) ),
+			sprintf(
+				"var wporgLocale = JSON.parse( decodeURIComponent( '%s' ) )",
+				rawurlencode( wp_json_encode( array(
+					'baseId' => get_locale(),
```
StevenDufresne: I think JavaScript API's call this the `baseName`. I'm not bothered by `baseId` though.

public_html/wp-content/themes/pattern-directory/functions.php
```diff
@@ -36,6 +37,31 @@ function setup() {
...
 }
 
+/**
+ * Determine if the current site is a Rosetta site (e.g., `es-mx.wordpress.org`).
+ *
+ * This returns `false` for `translate.wordpress.org`; it's part of the Rosetta network, but isn't a Rosetta site.
+ *
+ * @return bool
+ */
+function is_rosetta_site() : bool {
```
StevenDufresne: I wonder if we could move these rosetta helper functions into a `mu-plugin`?... maybe in the new [`wporg-mu-plugins` repo](https://github.com/WordPress/wporg-mu-plugins). If we needed them in the pattern directory, we are bound to need them elsewhere.

StevenDufresne (commented):
Nice! I left a few comments.

I logged a few issues that I found while testing these changes:
https://github.com/WordPress/pattern-directory/issues/464
https://github.com/WordPress/pattern-directory/issues/465

These issues don't have to be solved by this ticket, but it may be worth thinking about them so you don't have to come back?

public_html/wp-content/themes/pattern-directory/functions.php
outdoor2kode: Good point, it's definitely better to put this kind of function in one place and reuse them. I'll see how to put it together.

ryelle: These functions already exist [in the global blocks,](https://github.com/WordPress/wporg-mu-plugins/blob/b1ac446e2305dbd70c1d10950f1dc3f8680461af/mu-plugins/blocks/global-header-footer/blocks.php#L453) so you should be able to use them by pulling in `use function WordPressdotorg\MU_Plugins\Global_Header_Footer\{ get_rosetta_name, is_rosetta_site }`.

A possible improvement (to the mu-plugins project) could be to move the functions out of the blocks folder so it's clear they're not tied to the blocks.

public_html/wp-content/themes/pattern-directory/functions.php
outdoor2kode: I do like `baseName` more, but after looking into this [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale), I found that in its Parameters and Examples sections, it suggests that `ko-Kore-KR` is different from `ko-KR`. 
The former one is `baseName` and the latter one is `locale identifier` (see the description of tag under [Parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/Locale#parameters) and this [Example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/Locale#basic_usage)). 
So I'll change it to just `id` then. Thanks for bringing this up!

public_html/wp-content/themes/pattern-directory/functions.php
outdoor2kode: Didn't know those two functions can be pulled in that way here. Thanks!

> A possible improvement (to the mu-plugins project) could be to move the functions out of the blocks folder so it's clear they're not tied to the blocks.

I'll take a look at it. 👀

public_html/wp-content/themes/pattern-directory/functions.php
outdoor2kode: > So I'll change it to just id then. Thanks for bringing this up!

Done changing.

public_html/wp-content/themes/pattern-directory/functions.php
outdoor2kode: > These functions already exist [in the global blocks,](https://github.com/WordPress/wporg-mu-plugins/blob/b1ac446e2305dbd70c1d10950f1dc3f8680461af/mu-plugins/blocks/global-header-footer/blocks.php#L453) so you should be able to use them by pulling in use function WordPressdotorg\MU_Plugins\Global_Header_Footer\{ get_rosetta_name, is_rosetta_site }.

Done changing.

> A possible improvement (to the mu-plugins project) could be to move the functions out of the blocks folder so it's clear they're not tied to the blocks.
>> I'll take a look at it. 👀

I've created a new [issue](https://github.com/WordPress/wporg-mu-plugins/issues/191) in the wporg-mu-plugins repo, and I'll keep track of it.

renintw:
> Nice! I left a few comments.
> 
> I logged a few issues that I found while testing these changes: #464 #465
> 
> These issues don't have to be solved by this ticket, but it may be worth thinking about them so you don't have to come back?

I've taken a look at these two issues, and I've assigned them to myself. I'll create another PR for them.


@ryelle @StevenDufresne, please help review the PR again for what I've changed based on your comments, thanks!

outdoor2kode:
> Nice! I left a few comments.
> 
> I logged a few issues that I found while testing these changes: #464 #465
> 
> These issues don't have to be solved by this ticket, but it may be worth thinking about them so you don't have to come back?

I've taken a look at these two issues and have assigned them to myself. I'll create another PR for them.


@ryelle @StevenDufresne, I've made changes based on your comments, please help review them if you got time, thanks!

StevenDufresne:
This LGTM. I'll let @ryelle get the final thought seeing that she also had comments.

ryelle (approved):
Looks good! The title is updating correctly, and the correct locale is still used when loading patterns 👍🏻

PR #463, opened 2022-03-25
Use alternate source for embedding author data in internal notes

coreymckrill:
Currently the Internal Notes plugin retrieves embedded data about note authors using the standard `wp/v2/users` endpoint. However, this endpoint returns nothing if the specified user isn't a member of the site (which is most users on the Pattern Directory). This replaces that embeddable link with a wporg-specific endpoint that can get the data on any wporg user, regardless of site membership.

Depends on:
* https://github.com/WordPress/wporg-mu-plugins/pull/184
* https://github.com/WordPress/wporg-internal-notes/pull/12

### Screenshots

| Before | After |
| --- | --- |
| before | after |

1. Upload this change to your sandbox
2. Try viewing the Internal Notes sidebar for any pattern

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/logging.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/logging.php
@@ -23,8 +23,8 @@ add_action( 'transition_post_status', __NAMESPACE__ . '\flag_status_change', 10,
  * @return \WP_REST_Response
  */
 function replace_rest_controller_author_link( $response ) {
-	$resonse_data = $response->get_data();
-	$author = get_user_by( 'id', $resonse_data['author'] ?? 0 );
+	$response_data = $response->get_data();
+	$author = get_user_by( 'id', $response_data['author'] ?? 0 );
 
 	if ( ! $author ) {
 		return $response;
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/pattern-directory/includes/logging.php
```diff
@@ -9,8 +9,39 @@
 /**
  * Actions and filters.
  */
+add_filter( 'wporg_internal_notes_rest_prepare_response', __NAMESPACE__ . '\replace_rest_controller_author_link' );
```
ryelle, line 12: Why not make this change in the internal notes plugin itself?

coreymckrill: Good question. I originally thought of putting it there, but decided it would be best to keep the functionality in the plugin as environment-neutral as possible (which was inspired by @dd32 mentioning that he had tried using the plugin elsewhere on an external site). That said, all uses of the Internal Notes plugin I can think of within wporg would probably need this same modification, so perhaps this would be better somewhere in mu-plugins?

ryelle: Oh, I misread the endpoint namespace change in https://github.com/WordPress/wporg-internal-notes/pull/12, and thought it was already using wporg endpoints. Since it's not designed for wporg-only, I take back my comment :)

> all uses of the Internal Notes plugin I can think of within wporg would probably need this same modification, so perhaps this would be better somewhere in mu-plugins?

Yeah, that makes sense to me.

coreymckrill:
Closing this in favor of https://github.com/WordPress/wporg-mu-plugins/pull/192

Issue #449, opened 2022-03-21
Indicitive check for non-patterns at submission


tellyworth:
We've seen several submissions that - while not spam or objectionable in any way - aren't really patterns. For example:

* A list of text (think "John's favourite movies")
* A blog post
* An entire web page with lots of copywritten content
* Patterns with almost nothing in them, like just a single heading block with some text

These aren't suitable for the directory because they're not really reusable by others. And also because of translation issues: we don't want polyglots having to translate a pattern full of content that might never be used.

We could have some simple heuristic checks to detect this:

* Check the ratio of text content to other stuff
* Check for > n blocks, or > n different types of blocks
* Make sure those blocks have some attributes present, and aren't just default styles
* Look for common types of non-patterns ("unstyled list", "just paragaphs", etc)

Depending on the type and degree of match, it could take various actions:

* Warn the user before submitting
* Send the pattern to pending instead of published
* Publish it but alert a moderator to review
* Use a special status (publish but don't translate or promote?)

This needs some discussion before implementing. Suggestions welcome.

This definitely doesn't need to be done for launch. We'll learn more about non-patterns when we have more submissions.

This might also tie into the idea of better placeholder text for translators.

dd32:
> Check for > n blocks, or > n different types of blocks

Worth noting that I included a "If there's only paragraph blocks, nothing else, send it to spam" in the spam checks. I've since seen that 'bypassed' because they use a list or heading block in there too.

Perhaps that could be extended to be "If you're only using heading/paragraph/list blocks, goto spam" ie. non-content blocks.

My bigger question here though, is, what was the submitters intent? Were they spamming? Were they thinking it was going to be re-usable by others?

> Check the ratio of text content to other stuff

A high number of extracted strings would be a good thing to pause on, and go to pending rather than published/spam. A Pattern shouldn't have more than say, 25 strings, right? 25 oughta be enough for anyone? Maybe that number can be tweaked based on a comparison to other Patterns (This pattern contains more strings than 90% of other published blocks)

ryelle:
I've seen two patterns submitted over the weekend that copy existing patterns - one was a complete copy, and the other copied a pattern, then added a single image to the end.

Some more data points for "non-patterns" I've seen:

- Single image + paragraph
- Single image + quote
- Cover + single child (paragraph)
- Single HTML block

>  what was the submitters intent? Were they spamming? Were they thinking it was going to be re-usable by others?

[...]

PR #477, opened 2022-04-13
Pattern Validation: Try checking block count, flagged title words

ryelle:
See #449 — Adds some automated checks based on the comments in this issue.

1. Check if there are more than 3 non-empty blocks
2. Check if there are less than 75 non-empty blocks
3. Check if the title contains any flagged words, like "test" or "my pattern"

This uses the same validation format as the other content validation, where submitting a pattern is totally blocked.

I also updated the tests a bit to clean up some of the repeated pattern-saving code (TIL about phpunit's `dataProvider` feature).

### Screenshots

When invalid content is submitted, it shows up in the submission modal and as a banner above the title, so it's clear how to fix the issue even when the modal is closed.

error-notice

1. Create or edit a pattern
2. Use some combination of invalid content
3. You should see the error both in the modal and above the title
4. Fix the issue & resubmit, it should work

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
@@ -175,8 +175,10 @@ function validate_title( $prepared_post, $request ) {
 		return $prepared_post;
 	}
 
+	$title = isset( $request['title'] ) ? $request['title'] : get_the_title( $prepared_post->ID );
+
 	// A title exists, but is empty -- invalid.
-	if ( isset( $request['title'] ) && empty( trim( $request['title'] ) ) ) {
+	if ( isset( $title ) && empty( trim( $title ) ) ) {
 		return new \WP_Error(
 			'rest_pattern_empty_title',
 			__( 'A pattern title is required.', 'wporg-patterns' ),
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
```diff
@@ -375,3 +407,27 @@ function check_for_spam( $post ) {
...
+/**
+ * Helper function to check for a valid pattern title.
+ *
+ * @param string $title
+ * @return boolean
+ */
+function is_title_valid( $title ) {
+	// Check title against a list of disallowed words.
+	// Note the space after `test ` to avoid matching "testimonial".
+	$disallow_list = array( 'test ', 'testing', 'my pattern', 'wordpress', 'example' );
```
StevenDufresne, line 420: Should we add `pattern` to the list?

StevenDufresne (commented):
This is working well. I couldn't submit a _useless_ pattern.

I find the red notice in the background detracts from the modal. Have you tried using one or the other?

public_html/wp-content/plugins/pattern-directory/includes/pattern-validation.php
ryelle: I though about that, but I think `pattern` could refer to a patterned background, like "Headline over a bright pattern"

ryelle:
> I find the red notice in the background detracts from the modal. Have you tried using one or the other?

The red notice shows up as a result of saving the post. If we got rid of it, we would also be getting rid of other error feedback. It's not ideal, but I think it would be more complicated to remove.

I suppose we could hide the modal on error so the user just sees the red notice, but I'd worry the red banner is easy to overlook and it might seem like the pattern was saved. At least with the final modal screen, there's the big "There's an issue" headline.

StevenDufresne:
Following the theme of fewer modals is a good thing ... :)

Could we run the `< 3` & `> 75` checks on clicking `Submit` and show the red notice as is? That would make the creator align more with the way Gutenberg handles content/API errors. The title validation could be done inline where the input is in the modal... before going through the hassle of writing a description and choosing categories just to have it fail at something that was technically step 1.

ryelle:
> Could we run the < 3 & > 75 checks on clicking Submit and show the red notice as is?

Currently these checks are run as part of the saving process, so no, we can't just run those before going through the full modal flow. There would need to be a JS validation process, and the PHP checks would need to be rewritten in JS.

StevenDufresne (approved):

PR #478, opened 2022-04-14
Notifications: Trigger emails after post update is finished

### LINKED ISSUE #459: Pattern moderation: Email notif when unlisting a pattern doesn't show the correct reason
Pulling this from #451 

For some reason, when a pattern gets unlisted from the Block Editor screen, and a reason term gets assigned to it, the term is not included when `transition_post_status` is fired and the email notification to the pattern author is sent. Instead, the default fallback reason gets included in the email.

My suspicion is that when `transition_post_status` is fired, the term has not yet been attached to the post, although it seems like it should be, given the order of operations in `wp_insert_post`. Maybe there's a race condition happening somewhere?

### LINKED ISSUE #468: Moderation: Making an edit to a published pattern triggers a "spam" email
The notification code uses ["published -> pending"](https://github.com/WordPress/pattern-directory/blob/6a58d9d9ccd6581a4832b24c2cbbc18b437244aa/public_html/wp-content/plugins/pattern-directory/includes/notifications.php#L38) as a shortcut for "passed the flag threshold" and if there are no flags attached, it sends the "spam" email. But anyone editing a published pattern right now submits the pattern back into a pending state, so this is not a correct assumption.

To reproduce:

- Have a published pattern
- Edit it, it should set back to "pending" for reapproval
- You should get an email saying your pattern was flagged for spam, but it wasn't

ryelle:
This updates the logic for when to send a notification email to run on the `wp_after_insert_post` hook. This hook also gets the before & after post object, so we can compare post status like on the transition hook, but at this point the flag reasons have been saved.

I've added a custom action which fires when a pattern is unlisted, so we can attach the `notify_pattern_flagged` call to that action specifically, rather than checking for the publish -> pending switch (which fixes #468).

The logic for `notify_pattern_flagged` was changed slightly, now it checks for the spam status first, and sends the spam reason, rather than checking for flags first. It's also using `array_merge` instead of the array union operator (`+`), because the union op overwrites values at the same index, so if there were multiple reasons, each `term[0]` was overwriting the previous value.

Fixes #468, fixes #459.

Try to generate each email, and catch them with mailcatcher or similar (I used [the email log plugin](https://wordpress.org/plugins/email-log/) for my testing).

In Settings -> Block Patterns, make sure submitted patterns are saved as pending, and the flag threshold should be 2 for testing. Have two users to test with.

1. Submit a new pattern, it should save as pending
2. In wp-admin, publish the pattern
3. 📧 You should see a "Pattern published" email
4. View the pattern, and report it with both your users
5. The pattern should automatically unlist
6. 📧 You should see the "Pattern being reviewed" email, with the reason(s) chosen
7. Create another pattern
8. In wp-admin, mark it as spam
9. 📧 You should see the "Pattern being reviewed" email, with spam as the reason
10. Create another pattern
11. Go into wp-admin and manually unlist it
12.  📧 You should see the "Pattern unlisted" email, with the reason chosen

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/notifications.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/notifications.php
@@ -11,32 +11,39 @@ defined( 'WPINC' ) || die();
 /**
  * Actions and filters.
  */
-add_action( 'transition_post_status', __NAMESPACE__ . '\monitor_post_status_transitions', 20, 3 );
+add_action( 'wp_after_insert_post', __NAMESPACE__ . '\trigger_notifications', 20, 4 );
+add_action( 'wporg_unlist_pattern', __NAMESPACE__ . '\notify_pattern_flagged' );
 
 /**
- * Hook into post status transitions to detect changes that warrant a notification.
+ * Fire off relevant notification when a post is finished updating.
  *
- * @param string   $new_status
- * @param string   $old_status
- * @param \WP_Post $post
+ * @param int           $post_id     Post ID.
+ * @param \WP_Post      $post        Post object.
+ * @param bool          $update      Whether this is an existing post being updated.
+ * @param null|\WP_Post $post_before Null for new posts, the WP_Post object prior
+ *                                  to the update for updated posts.
  *
  * @return void
  */
-function monitor_post_status_transitions( $new_status, $old_status, $post ) {
+function trigger_notifications( $post_id, $post, $update, $post_before ) {
 	if ( PATTERN !== get_post_type( $post ) ) {
 		return;
 	}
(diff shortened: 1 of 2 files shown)


coreymckrill (approved):
👍 This looks great! Using `wp_after_insert_post` seems way better. We should probably do that for [logging flags](https://github.com/WordPress/pattern-directory/blob/c761277102efef93316aaffd4ecb7f811a1b2732/public_html/wp-content/plugins/pattern-directory/includes/logging.php#L12) and the logging in the [Internal Notes plugin](https://github.com/WordPress/wporg-internal-notes/issues/13) as well.

PR #514, opened 2022-08-18
Theme: Fix layout style bleed from previews into main theme

### LINKED ISSUE #513: Preview styles are affecting the admin bar
After an update to GB 13.9, the admin bar now has a max-width of 800px.

image

The offending style is the `body > div > :where…` from here, which is generated as part of "Core styles: block-supports".

```css
.entry-content > :where(:not(.alignleft):not(.alignright):not(.alignfull)),
body > div > :where(:not(.alignleft):not(.alignright):not(.alignfull)) {
  max-width: 800px;
  margin-left: auto !important;
  margin-right: auto !important;
}
```

This style is set up [in the theme here](https://github.com/WordPress/pattern-directory/blob/b6b320cc3b119cfaa6786d9bd4d400466361c9bf/public_html/wp-content/themes/pattern-directory/functions.php#L157), but in previous GB versions, this only output as the `window.__editorStyles` value passed into the pattern previews. Somehow it's being cached(?) and also output along with the regular theme layout styles.

This is probably the issue, `WP_Style_Engine::store_css_rule(…`: https://github.com/WordPress/gutenberg/blob/29d6801902ad97f749ccadd789a23b2d97369e3e/packages/style-engine/class-wp-style-engine.php#L613

ryelle:
Fixes #513 — Any styles added via `gutenberg_get_layout_style` will be added to the global `block-supports` store in the style engine (definitely new in GB 13.9, possibly from https://github.com/WordPress/gutenberg/pull/42880). The pattern previews used to need a workaround to support layout width (see #428), which was generating the styles injected via `window.__editorStyles`. This function now also has a side effect of adding that rule to the main site too, which is breaking the admin bar.

The fix added here is to use `entry-content` class on the preview & thumbnail container divs, so that the theme & default styles are correctly applied. Then we can remove the call to `gutenberg_get_layout_style`. Removing old workarounds 👍🏻 

Props kafleg, dd32

### Screenshots

| Before | After |
|--------|-------|
| list-before | list-after |
| preview-before | preview-after |

This also adds the block spacing between the groups ^ which matches the editor UI, so I think that's overall a good thing.

1. View the front end, the admin bar should not appear centered
2. Create or view some patterns with varying widths
3. The default width should be 800px wide, wide should be 1000px wide, and full width should be edge-to-edge
4. Block spacing matches the editor
5. There should be no regressions in style

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/functions.php
+++ b/public_html/wp-content/themes/pattern-directory/functions.php
@@ -150,13 +150,6 @@ function generate_block_editor_styles_html() {
 	wp_styles()->do_items( $handles );
 	wp_styles()->done = $done;
 
-	// Build up the alignment styles to match the layout set in theme.json.
-	// See https://github.com/WordPress/gutenberg/blob/9d4b83cbbafcd6c6cbd20c86b572f458fc65ff16/lib/block-supports/layout.php#L38
-	$block_gap = wp_get_global_styles( array( 'spacing', 'blockGap' ) );
-	$layout = wp_get_global_settings( array( 'layout' ) );
-	$style = gutenberg_get_layout_style( 'body > div', $layout, true, $block_gap );
-	echo '<style>' . $style . '</style>'; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-
 	wp_add_inline_script(
 		'wporg-pattern-script',
 		sprintf(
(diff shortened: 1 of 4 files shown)


adamwoodnz:
Didn't quite get finished with my review this week @ryelle 

The admin bar LGTM but haven't tested with many patterns as my test data is a bit limited. Maybe @dd32 has a bit more context on the old workaround and can approve?

ryelle:
I'm going to merge this then, I haven't see any issues & would like to fix the admin bar.

PR #521, opened 2022-09-23
Add a new endpoint for pattern previews

### LINKED ISSUE #519: Social Icons are misaligned in the site footer
See https://meta.trac.wordpress.org/ticket/6503

Issue

This is caused by the local `inline-styles.php` code, which injects the computed block styles into the patterns. As a side-effect, it removes the core/GB handling that would apply the `is-layout-flex` class.

ryelle:
This pulls out one change from #518, where I introduced a new endpoint for pattern previews — `wp.org/patterns/pattern/slug/view/`. Now the iframe can use that URL directly, and since the page has both the header & footer hooks in it, the theme doesn't need to do any workarounds for `__editorStyles` or generating Gutenberg styles.

Additionally, since the theme's styles are applied by WordPress rather than by loading a single theme stylesheet, the layout of the patterns looks better 🎉 

This PR also removes the `Iframe` component, since we no longer need all the special handling of injecting head & body content into an empty iframe.

Fixes #519 because we can remove the Gutenberg workaround causing that bug.

### Screenshots

|   | Before | After |
|---|--------|------|
| Grid view | grid-trunk | grid-after |
| Single pattern | pattern-trunk | pattern-after |
| Footer alignment is fixed | footer-trunk | footer-after |

1. Apply the PR, build files `yarn workspaces run build`
2. `composer update` if needed to get the latest Gutenberg
3. Refresh permalinks by going to Settings > Permalinks (this will cause a change in `.htaccess`, discard it)
4. View the front page, navigate through some patterns
5. There shouldn't be any drastic visual changes
6. In the single pattern preview, you should still be able to resize the preview

Things to check: wide/full alignments & child block alignments, duotone, content spacing & colors.


public_html/wp-content/plugins/pattern-directory/views/view.php
```diff
@@ -0,0 +1,38 @@
...
+			width:100%;
+			pointer-events: none;
+		}
+	</style>
+</head>
+
+<body <?php body_class(); ?>>
+<?php wp_body_open(); ?>
+
+<div class="wp-site-blocks">
```
StevenDufresne, line 30: Any chance that Gutenberg CSS styles can conflict and cause the content to not be centered or introduce other side effects?

public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/canvas.js
```diff
@@ -55,28 +54,19 @@ function PatternThumbnail( { className, html } ) {
...
 			ref={ wrapperRef }
 			style={ {
 				height: frameHeight,
+				overflow: 'hidden',
 			} }
-			tabIndex="-1"
 		>
-			<Iframe
-				className="pattern-grid__preview-iframe"
+			<iframe
```
StevenDufresne, line 63: I think we should aim at using an [image snapshot](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/blocks/screenshot-preview) in a future PR.

StevenDufresne (approved):
This is a great update! It definitely makes pattern previewing more predictable and easier to debug. 👍 

I imported some content from wp.org to test this PR, but I can't say that my local configuration of those patterns is a good test environment. Some patterns looked a bit off.

I think this is ultimately a more reliable approach and I would rather iterate on these changes regardless of the minor differences I noticed. If you would like a more thorough examination from someone else, let me know. I can test in my sandbox. But for now, I don't feel there is a substantial risk to _spin_ everything up :)

public_html/wp-content/plugins/pattern-directory/views/view.php
ryelle: Theoretically no, if anything this should fix assorted alignment issues since this more accurately reflects how the patterns are used in real sites. It's definitely more stable than the workaround we're using now (which you can see in the before screenshots, is not working right now).

public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/canvas.js
ryelle: Good idea for these thumbnail views  👍🏻

PR #522, opened 2022-09-26
Fix 2 add/remove event listener typos.

StevenDufresne:
This PR updates 2 _typos_ that lead to significantly more `resize` event listeners on the `window`.

1. Load the pattern homepage in the Chrome Developer Tools console, run `getEventListeners(window).resize`.
2. Choose a different category (will the same number of patterns)
3. run `getEventListeners(window).resize`, expect to see the same number of resize events.


ryelle (approved):
Good catch 👍🏻

PR #528, opened 2022-10-05
Use `mshots` service to generate pattern screenshots

### LINKED ISSUE #349: Pattern Browsing: Very large images are used in grid view
When browsing images, the thumbnails load the entire image used in the pattern, which is the full-size image. These images are all very large, over 1500x1500. Loading the `/patterns/` homepage is 7.8MB just of images.

These were all optimized before uploading, so further compressing them would probably affect quality.

The large images are good when viewing the full-sized pattern, and when using the pattern locally. When viewing the patterns in grid-view, though, it would be nice to use a smaller image. The pattern scale is usually around 33%, but in some responsive views (500px wide) can be 50%, so I think it would be best (visually) to scale the images down about 50%.

Maybe we could filter the rendered content when in an archive view? We would need a dynamic way to get half-sized images from the originals though, maybe the WP-generated thumbnails. I think the `medium_large` size (768px wide, scaled) could work?

ryelle:
See https://github.com/WordPress/pattern-directory/pull/521#discussion_r979500622 — This swaps out the iframes in the pattern grid in favor of an mshots request. This does improve the number and overall size of requests, but it doesn't work on local sites.

**Stats** for loading the homepage (and scrolling down to load all pattern previews)

`trunk`: 416 requests, 30.75 MB / 9.35 MB transferred.
`try/mshots`: 177 requests, 5.91 MB / 3.28 MB transferred.

I forked [the ScreenShotImg component from `wporg-mu-plugins`](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/screenshot-preview/src/screenshot.js) so it could be used here, and added a few more props.

Fixes #349

### Screenshots

| Before | After |
|--------|------|
| <video src="https://user-images.githubusercontent.com/541093/194179167-04a45eed-9ecd-4daa-bacb-b8d4058762eb.mp4"> | <video src="https://user-images.githubusercontent.com/541093/194179174-d3c48a65-8580-4d06-bcd2-616d8b64d68f.mp4"> |

On local sites it just returns this 403 image. This is not great, but I'm not sure if this should be a blocker.

Screen Shot 2022-10-05 at 19 08 47

1. Apply this to a sandbox
2. View assorted pattern grids — categories, my patterns, etc
3. The small pattern previews should look correct & match the larger preview on the pattern detail page

### DIFF
--- a/public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/canvas.js
+++ b/public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/canvas.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { __, isRTL } from '@wordpress/i18n';
+import { __ } from '@wordpress/i18n';
 import { useEffect, useRef, useState } from '@wordpress/element';
 
 /**
@@ -9,14 +9,14 @@ import { useEffect, useRef, useState } from '@wordpress/element';
  */
 import getCardFrameHeight from '../../utils/get-card-frame-height';
 import useInView from '../../hooks/in-view';
+import Screenshot from './screenshot';
 
 const VIEWPORT_WIDTH = 1200;
 
-export default function ( { url } ) {
+export default function ( { alt, url } ) {
 	const wrapperRef = useRef();
-	const [ frameHeight, setFrameHeight ] = useState( '1px' );
-	const [ frameScale, setFrameScale ] = useState( 0.3125 );
 	const isVisible = useInView( { element: wrapperRef } );
+	const [ frameHeight, setFrameHeight ] = useState( '1px' );
 	const [ shouldLoad, setShouldLoad ] = useState( false );
 
 	useEffect( () => {
@@ -29,7 +29,6 @@ export default function ( { url } ) {
 		const handleOnResize = () => {
 			try {
 				setFrameHeight( getCardFrameHeight( wrapperRef.current.clientWidth ) );
-				setFrameScale( wrapperRef.current.clientWidth / VIEWPORT_WIDTH );
 			} catch ( err ) {}
 		};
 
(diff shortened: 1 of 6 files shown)


public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/canvas.js
```diff
@@ -40,32 +39,27 @@ export default function ( { url } ) {
...
+			<Screenshot
 				className="pattern-grid__preview"
-				title={ __( 'Pattern Preview', 'wporg-patterns' ) }
-				tabIndex="-1"
+				alt={ alt || __( 'Pattern Preview', 'wporg-patterns' ) }
 				style={ style }
-				src={ shouldLoad ? url : '' }
+				size={ { width: VIEWPORT_WIDTH, height: getCardFrameHeight( VIEWPORT_WIDTH ) } }
+				isReady={ shouldLoad }
+				src={ url }
```
StevenDufresne: Can patterns be updated? 

If so you'll want to introduce a way to break the cache so a new mShot is generated.

public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/screenshot.js
```diff
@@ -0,0 +1,117 @@
...
+	);
+
+	if ( ! isReady ) {
+		return null;
+	}
+
+	if ( isLoading ) {
+		return (
+			<div className={ `${ className } is-loading` } style={ style }>
+				<Spinner style={ { width: '33%', height: '33%' } } />
```
StevenDufresne: The spinner feels a little bit large, but that's personal preference.

ryelle: Fair enough - the default size was very tiny, and this wasn't really chosen with any intention. Swapped it out for 2x default size.

Screen Shot 2022-10-06 at 12 27 49 PM

public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/canvas.js
ryelle: Yeah, they can, good catch — I added the pattern's modified date as part of the URL.

StevenDufresne:
LGTM 👍.

public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/index.js
```diff
@@ -35,9 +34,7 @@ function PatternThumbnail( { pattern, showAvatar, showOptions } ) {
 			<div className="pattern-grid__pattern-frame">
 				<a href={ pattern.link } rel="bookmark">
 					<span className="screen-reader-text">{ decodeEntities( pattern.title.rendered ) }</span>
-					<Disabled>
-						<Canvas url={ addQueryArgs( pattern.link, { view: true } ) } />
-					</Disabled>
+					<Canvas url={ addQueryArgs( pattern.link, { view: true, modified: pattern.modified_gmt } ) } />
```
ryelle: `modified` will break the cache when a pattern is modified, but there should probably also be a site-wide cachebuster in case we update styles in the pattern preview endpoint.

StevenDufresne: I wouldn't be opposed to a manual cache break if it's really needed in short order.

PR #537, opened 2022-10-31
Use `REST_REQUEST` over `wp_is_json_request()`

dd32:
When accessing REST API's via the browser, `wp_is_json_request()` is false which causes the ES modification args not to kick in.

By changing to the `REST_REQUEST` constant it should affect it, while also catching the same scenario's.

The one case that this won't catch, is when a PHP function is making a rest-api call from within a non-rest-api request. However. The existing check also doesn't work in that case, unless it was an XHR request with the appropriate json headers.

I think this change is safe, but PR'ing incase there's something else preventing this.

Fixes #, See #

Props <username>, <username>

1.
2.
3.


ryelle (approved):
This change looks good to me, this seems more like the constant we should be calling anyway given the comment here.

PR #540, opened 2022-11-16
Add meta and query parameter to fetch patterns matching a list of allowed blocks

ryelle:
See https://github.com/WordPress/gutenberg/pull/45237

> Another really important issue is that if we make any request to the PD, the eligible patterns to be inserted are subject to the `allowedBlocks` in the editor or a block. What this means is that we could fetch 10 patterns for example where none of them could be inserted.

This PR aims to add an API parameter `allowed_blocks` to the local endpoint<sup>[1]</sup> so that sites with restrictions can pass through the list of valid blocks, and the pattern directory will return only patterns with blocks on the list.

To do this, a new meta value is used to track the blocks in a pattern, `wpop_contains_block_types`. There's also a migration script which can be run on a sandbox to update all existing patterns, a hook to update the meta when a pattern is updated, and the meta is set on translated patterns when those are updated.

cc @ntsekouras 

-----

**[1]** Once merged, it should be available in `api.w.org`, but will need another Gutenberg PR before it can be used in the editor.

**Test the migration script**

1. You can run the script on your docker wp-env using the `cli` container.
2. To dry-run over all posts, use:
  `yarn wp-env run cli "php wp-content/plugins/pattern-directory/bin/update-contains-block-types.php --all --verbose"`
3. To apply the changes, use:
  `yarn wp-env run cli "php wp-content/plugins/pattern-directory/bin/update-contains-block-types.php --all --verbose --apply"`
4. You can also run on just one post by using `--post=[id]` instead of `--all`
5. The updated meta should be visible in the API, or via wp-cli `yarn wp-env run cli "wp post meta list [ID]"`
6. Spot check some posts that the meta value reflects the post content

**Test the pattern update hook**

1. Create or update a pattern, change the block types used
2. Check the post meta
3. The meta value should exist, and reflect the blocks used

**Test the API endpoint**

This requires some patterns having the meta set, which is why it's last.

Make a request to your site using a list of block types, ex:

```
http://localhost:8888/wp-json/wp/v2/wporg-pattern?locale=en_US&allowed_blocks[]=core/cover&allowed_blocks[]=core/heading&allowed_blocks[]=core/spacer&allowed_blocks[]=core/paragraph
```
This requests only patterns with `core/cover, core/heading, core/spacer, core/paragraph`. Any pattern with other blocks should not be returned in the list.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/bin/update-contains-block-types.php
+++ b/public_html/wp-content/plugins/pattern-directory/bin/update-contains-block-types.php
@@ -1,0 +1,102 @@
+<?php
+/**
+ * Update all patterns with the "contains block types" meta field.
+ *
+ * To run locally, use `wp-env`, ex:
+ * yarn wp-env run cli "php wp-content/plugins/pattern-directory/bin/update-contains-block-types.php --all --per_page=100 --apply"
+ *
+ * To run in a sandbox, use php directly, ex:
+ * php ./bin/update-contains-block-types.php --all --apply
+ */
+
+namespace WordPressdotorg\Pattern_Directory;
+
+use const WordPressdotorg\Pattern_Directory\Pattern_Post_Type\{ POST_TYPE };
+
+// This script should only be called in a CLI environment.
+if ( 'cli' != php_sapi_name() ) {
+	die();
+}
+
+$opts = getopt( '', array( 'post:', 'url:', 'abspath:', 'per_page:', 'all', 'apply', 'verbose' ) );
+
+if ( empty( $opts['url'] ) ) {
+	$opts['url'] = 'https://wordpress.org/patterns/';
+}
+
+if ( empty( $opts['abspath'] ) && false !== strpos( __DIR__, 'wp-content' ) ) {
+	$opts['abspath'] = substr( __DIR__, 0, strpos( __DIR__, 'wp-content' ) );
+}
+
+$opts['apply']   = isset( $opts['apply'] );
+$opts['verbose'] = isset( $opts['verbose'] );
+$opts['all']     = isset( $opts['all'] );
+
+// Bootstrap WordPress
+$_SERVER['HTTP_HOST']   = parse_url( $opts['url'], PHP_URL_HOST );
+$_SERVER['REQUEST_URI'] = parse_url( $opts['url'], PHP_URL_PATH );
+
+require rtrim( $opts['abspath'], '/' ) . '/wp-load.php';
+
+if ( ! $opts['all'] && ! isset( $opts['post'] ) ) {
+	fwrite( STDERR, "Error! Either specify a post ID with --post=<ID> or explicitly run over --all.\n" );
+	die();
+}
+
+if ( ! $opts['apply'] ) {
+	echo "Dry run, will not update any patterns.\n";
+}
+
+$args = array(
+	'post_type'      => POST_TYPE,
+	'post_status'    => array( 'publish', 'pending' ),
+	'posts_per_page' => isset( $opts['per_page'] ) ? $opts['per_page'] : -1,
+	'post_parent'    => 0,
+	'orderby'        => 'date',
+	'order'          => 'DESC',
+	'meta_query'     => array(
+		// Only update patterns without this meta.
+		array(
+			'key'     => 'wpop_contains_block_types',
+			'compare' => 'NOT EXISTS',
+		),
+	),
+);
+if ( isset( $opts['post'] ) ) {
+	$args = array(
+		'post_type' => POST_TYPE,
+		'p' => absint( $opts['post'] ),
+	);
+}
+
+$query = new \WP_Query( $args );
+$meta_updated = 0;
+
+while ( $query->have_posts() ) {
+	$query->the_post();
+	$pattern    = get_post();
+	$pattern_id = $pattern->ID;
+	$blocks     = parse_blocks( $pattern->post_content );
+	$all_blocks = _flatten_blocks( $blocks );
+
+	// Get the list of block names and convert it to a single string.
+	$block_names = wp_list_pluck( $all_blocks, 'blockName' );
+	$block_names = array_filter( $block_names );
+	$block_names = array_unique( $block_names );
+	sort( $block_names );
+	$used_blocks = implode( ',', $block_names );
+
+	if ( $opts['apply'] ) {
+		$result = update_post_meta( $pattern_id, 'wpop_contains_block_types', $used_blocks );
+		if ( $result ) {
+			$meta_updated++;
+		} else if ( $opts['verbose'] ) {
+			echo "Error updating {$pattern_id}.\n"; // phpcs:ignore
+		}
+	} else if ( $opts['verbose'] ) {
+		echo "Will update {$pattern_id} with '{$used_blocks}'.\n"; // phpcs:ignore
+	}
+}
+
+echo "Updated {$meta_updated} patterns.\n"; // phpcs:ignore
+echo "Done.\n\n"; // phpcs:ignore
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -653,6 +653,14 @@ function filter_patterns_collection_params( $query_params ) {
 		'type'        => 'string',
 	);
 
+	$query_params['allowed_blocks'] = array(
+		'description' => __( 'Filter the request to only return patterns with blocks on this list.', 'wporg-patterns' ),
+		'type'        => 'array',
+		'items'       => array(
+			'type' => 'string',
+		),
+	);
+
 	return $query_params;
 }
 
@@ -715,6 +723,16 @@ function filter_patterns_rest_query( $args, $request ) {
 				'key'     => 'wpop_wp_version',
 				'compare' => 'NOT EXISTS',
 			),
+		);
+	}
+
+	$allowed_blocks = $request->get_param( 'allowed_blocks' );
+	if ( $allowed_blocks ) {
+		// Only return a pattern if all contained blocks are in the allowed blocks list.
+		$args['meta_query']['allowed_blocks'] = array(
+			'key'     => 'wpop_contains_block_types',
+			'compare' => 'REGEXP',
+			'value'   => '^((' . implode( '|', $allowed_blocks ) . '),?)+$',
 		);
 	}
 
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -15,6 +15,7 @@ add_action( 'init', __NAMESPACE__ . '\register_post_type_data' );
 add_action( 'rest_api_init', __NAMESPACE__ . '\register_rest_fields' );
 add_action( 'init', __NAMESPACE__ . '\register_post_statuses' );
 add_action( 'transition_post_status', __NAMESPACE__ . '\status_transitions', 10, 3 );
+add_action( 'post_updated', __NAMESPACE__ . '\update_contains_block_types_meta' );
 add_action( 'enqueue_block_editor_assets', __NAMESPACE__ . '\enqueue_editor_assets' );
 add_filter( 'allowed_block_types_all', __NAMESPACE__ . '\remove_disallowed_blocks', 10, 2 );
 add_action( 'enqueue_block_editor_assets', __NAMESPACE__ . '\disable_block_directory', 0 );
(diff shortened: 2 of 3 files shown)


public_html/wp-content/plugins/pattern-directory/bin/update-contains-block-types.php
```diff
@@ -0,0 +1,102 @@
...
+	$pattern    = get_post();
+	$pattern_id = $pattern->ID;
+	$blocks     = parse_blocks( $pattern->post_content );
+	$all_blocks = _flatten_blocks( $blocks );
+
+	// Get the list of block names and convert it to a single string.
+	$block_names = wp_list_pluck( $all_blocks, 'blockName' );
+	$block_names = array_filter( $block_names );
+	$block_names = array_unique( $block_names );
+	sort( $block_names );
```
ntsekouras, line 86: Do you think it will help performance sorting the block names?

public_html/wp-content/plugins/pattern-directory/bin/update-contains-block-types.php
```diff
@@ -0,0 +1,102 @@
...
+
+while ( $query->have_posts() ) {
+	$query->the_post();
+	$pattern    = get_post();
+	$pattern_id = $pattern->ID;
+	$blocks     = parse_blocks( $pattern->post_content );
+	$all_blocks = _flatten_blocks( $blocks );
+
+	// Get the list of block names and convert it to a single string.
+	$block_names = wp_list_pluck( $all_blocks, 'blockName' );
```
ntsekouras, line 83: I think it's better to remove the `core/` prefix from the blocks and the reason is that we can end up with a quite big `GET` request that could exceed the limit. If every core block is allowed it seems we have `60` of them.

If in the future we add support in pattern directory for third party blocks, we can revisit and I don't think a solution would be that hard.

For now in my GB PR, I'm stripping the `core/` prefix: https://github.com/WordPress/gutenberg/pull/45237/commits/0b3d6e088343db6f525197e8552507986fcf8a06

ntsekouras (commented):
Thanks so much for working on this @ryelle! It might be the last big missing piece for the integration with GB, so I'm really excited!

In the [GB PR](https://github.com/WordPress/gutenberg/pull/45237) I have added support for this. How can we test this though and land this PR?

public_html/wp-content/plugins/pattern-directory/bin/update-contains-block-types.php
ryelle: > If every core block is allowed it seems we have 60 of them.

If every block is allowed, you could just not send the parameter, and we won't need to do any filtering.

> If in the future we add support in pattern directory for third party blocks

Yeah, this is trying to be future-friendly, so the full block name is necessary in the meta (otherwise we'll need another script to update it later…).

> we can end up with a quite big GET request that could exceed the limit. 

I think we're safe for now with just core blocks - are you running into the limit in your PR?

public_html/wp-content/plugins/pattern-directory/bin/update-contains-block-types.php
ryelle: Honestly this is mostly for devex, so it's easier to check the value against the expected value.

public_html/wp-content/plugins/pattern-directory/bin/update-contains-block-types.php
ntsekouras: >If every block is allowed, you could just not send the parameter, and we won't need to do any filtering.

That would be more convoluted as we would need to have operations in sync with the Blocks API(like `parent, ancestor, etc..`), so for now let's not do that. Even with having the above though, he problem remains if for example we have just one `disallowed` block. 

>I think we're safe for now with just core blocks - are you running into the limit in your PR?

With the prefix the length of `allowed_blocks` is `962` characters. It's quite long but I think we will be in the limit for now at least. Let's try that.

>Yeah, this is trying to be future-friendly, so the full block name is necessary in the meta (otherwise we'll need another script to update it later…).

We could just modify the script to add the prefix in `non-core` blocks, but as I said above we can try with the prefix for now and let's see if we need something different in the future.

I have [updated the GB PR ](https://github.com/WordPress/gutenberg/pull/45237/commits/3bcfa1de8313209958605714317cfdec9f65d56a)to send the prefix.

ntsekouras: >That would be more convoluted as we would need to have operations in sync with the Blocks API(like parent, ancestor, etc..), so for now let's not do that.

Hm... I just realized that there are some nuances in this. What I'm sending you right now doesn't include the blocks for example that have a specific parent(ex `List item and List`), but your parsing of the content will probably have the `List item` as well, correct?

ryelle: > your parsing of the content will probably have the List item as well, correct?

Yes, it collapses down all the blocks used in a pattern, for example: `core/column,core/columns,core/group,core/heading,core/list,core/list-item,core/paragraph,core/separator,core/social-link,core/social-links`

ntsekouras: I [updated GB PR](https://github.com/WordPress/gutenberg/pull/45237/commits/b16f8a8ca0dadb8483cdb4bb552ae7ce9ebc0997) to also send blocks that have a `parent` or `ancestor` relationship. 

Still I have the feeling that the `core/prefix` adds more overhead right now for some reasons:
1. A basic request now is `~1263` characters and `~898` without the prefix
2. I have the feeling that the `regex` will execute slower with the prefix as for every block it will always match everything for the first 5 characters. I haven't tested/measured anything though..
3. If in the future we add support for third party blocks somehow, they can have prefix and we will not need a migration script, just a small change when you insert the pattern script. I'm not sure there will be such support in the near future, but 🤷 

Having said that, we can go with the prefix if you have strong opinion and the important thing for me is how we can test this and hopefully land the integration soon. It seems really close to me 😄

ntsekouras: I have added the logic to include the parent/ancestor relationship and if all available core blocks are allowed to be inserted, I'm sending the `all` value as a `string`.

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
```diff
@@ -700,6 +747,16 @@ function filter_patterns_rest_query( $args, $request ) {
 		);
 	}
 
+	$allowed_blocks = $request->get_param( 'allowed_blocks' );
+	if ( $allowed_blocks ) {
+		// Only return a pattern if all contained blocks are in the allowed blocks list.
+		$args['meta_query']['allowed_blocks'] = array(
+			'key'     => 'wpop_contains_block_types',
+			'compare' => 'REGEXP',
+			'value'   => '^((' . implode( '|', $allowed_blocks ) . '),?)+$',
```
dd32, line 756: This isn't going to be amazing performance wise, but I'm not sure I can see a better way to achieve this, other than filtering post-query.

public_html/wp-content/plugins/pattern-directory/bin/update-contains-block-types.php
dd32: > If every core block is allowed it seems we have 60 of them.

This is going to affect the cacheability of the API significantly if every API request sends a list of their installed blocks that we then filter by. Due to the Pattern APIs using WordPress to serve the API requests with nginx caching in-front, this could be a longer-term performance issue.

We should require/suggest strongly, at a minimum, that the allowed_blocks parameter should be sorted consistently to avoid two sites with different block orders getting different cache keys.

I can't see that clients could reasonably send a "I don't support this block" flag though, as it doesn't know what it doesn't know about.

dd32 (commented):
As noted above, this is going to have some potentially significant performance bottlenecks.

At a high-level, this feels like something that potentially should be being filtered client-side, noting that would play havoc with the pagination.

The other aspect of this is that currently a Pattern that uses a WordPress 6.2 block might be returned to a 6.0 site, it would make sense to have the API differ in response in that case, but the client should still validate the pattern can be used by it.

ryelle:
> At a high-level, this feels like something that potentially should be being filtered client-side, noting that would play havoc with the pagination.

That's the issue— the editor could request X patterns, but if a site only supports a few blocks, once they're filtered out it will look like nothing's possible. The editor could continue to request additional pages of patterns to fill out the original query, but then it's running multiple API requests, and I think the experience of that would be slower than this approach.

> The other aspect of this is that currently a Pattern that uses a WordPress 6.2 block might be returned to a 6.0 site, it would make sense to have the API differ in response in that case, but the client should still validate the pattern can be used by it.

Having the client validate blocks was suggested as part of [the solution for compatibility issues here](https://github.com/WordPress/gutenberg/issues/45179#issuecomment-1287156526), the other part is manually setting a meta value on patterns for a minimum wordpress version (https://github.com/WordPress/pattern-directory/pull/536).

dd32:
> > At a high-level, this feels like something that potentially should be being filtered client-side, noting that would play havoc with the pagination.
> 
> That's the issue— the editor could request X patterns, but if a site only supports a few blocks, once they're filtered out it will look like nothing's possible. The editor could continue to request additional pages of patterns to fill out the original query, but then it's running multiple API requests, and I think the experience of that would be slower than this approach.

Yeah, it's a balancing act of where to have the performance impact.
Multiple cached responses would likely be more efficient than a single uncached response though.
The other aspect to consider is how many people does this actually affect? Multiple API requests for 0.5% of users is better than the other 99.5% getting a 50% cache hit rate or something. I don't personally see it being an issue if only 1 or 2 results can be used/displayed by the client before the next API request is fired.
There's also a limit in characters in a GET request - the path + query must be less than ~2,000 characters by spec, sure, that's 200+ blocks depending upon the implementation, but it's a limit that may eventually be run into. (I believe ~8k chars is the limit for api.wordpress.org before a `HTTP 414 Request too large` is hit, but you still have to consider Proxy servers/etc)

The other option is for Gutenberg to only include that parameter when it **knows** that a core block has been disabled. That could be done by having a static list of blocks that _"should"_ exist in that Gutenberg version, and instead sending a `exclude_blocks` parameter.

ntsekouras:
>The other aspect to consider is how many people does this actually affect? Multiple API requests for 0.5% of users is better than the other 99.5% getting a 50% cache hit rate or something. I don't personally see it being an issue if only 1 or 2 results can be used/displayed by the client before the next API request is fired.

I actually had a conversation with @mcsf about this yesterday and he was asking the same thing and suggesting this. Besides the pagination handling, that in my mind seems 'chaotic' 😄 but could be done, I don't think we have this info. I'd say that most people have all blocks activated, but 🤷 . If that's the usual case though to support all blocks, I'm not sending the param in the request, so no problems there..

All the patterns operations are expensive(parsing, rendering) and adding the recursive validation and eligibility for insertion checks might take a lot of time - we might end up needing to implement and test these though..

>The other option is for Gutenberg to only include that parameter when it knows that a core block has been disabled. That could be done by having a static list of blocks that "should" exist in that Gutenberg version, and instead sending a exclude_blocks parameter.

The workaround for this would be to support both `include` and `exclude` params and only send the one that contains less blocks. For example if only a couple of blocks are disabled, send them.

I think what we are missing right now is actually test the current approach against https://github.com/WordPress/gutenberg/pull/45237 and see the performance impact. How can we do that?

tellyworth:
By way of challenging an assumption: is the client side definitely doing the right thing by not displaying disabled blocks to the user at all? Would it be a better experience all round if it included them in the list, but showed them greyed out/disabled/similar? That way the user won't go hunting through x pages for a block they know exists but can't find, and there's an opportunity for the UI to explain to the user why that block isn't there.

That would obviously simplify the server side by eliminating the need for a query parameter at all.

ntsekouras:
>is the client side definitely doing the right thing by not displaying disabled blocks to the user at all?

Yes, the Gutenberg side shouldn't change, as it would add more noise and be way more confusing to show any unavailable blocks and explain why. The current alternative we might explore is doing all the work client side and handle pagination and possible multi-requests there. We should test the current approach first though, since it's implemented in both projects - but not merged to neither yet.

tellyworth:
Thanks Nik.

Let's merge this so we can test it out. If performance turns out to be an issue we can go with client-side handling.

ryelle:
@ntsekouras This has been merged, so you can now query api.wordpress.org/patterns with `allowed_blocks`, ex: 

```
https://api.wordpress.org/patterns/1.0/?allowed_blocks[]=core/heading&allowed_blocks[]=core/gallery
```

ntsekouras:
Thanks @ryelle! 🎉 

I'm testing the GB PR and it seems can fetch patterns with more blocks, if we also provide the `search` param, ex: 
```
https://api.wordpress.org/patterns/1.0/?search=image&allowed_blocks[]=core/heading&allowed_blocks[]=core/gallery
```

Any insights about this?

ryelle:
Huh… I'm guessing it has to do with the fact that we query ElasticSearch for search results, and it probably doesn't know to pay attention to the `allowed_blocks`. Tracking this in https://github.com/WordPress/pattern-directory/issues/547.

PR #541, opened 2022-11-21
The code requires PHP 8, so tell wp-env to use it.

tellyworth:
The codebase includes a couple of calls to `str_contains()`, which is PHP 8.0+ only. It's possible to get a fatal error when running the local environment on 7.4, but that will be the default for some users.

This one-liner tells wp-env to use PHP 8.0.


dd32:
WordPress contains a compat version of the `str_*` functions for older PHP releases.

Any reason this shouldn't be bumped to PHP 8.1 though? (The next version we'll use on WordPress.org)

ocean90:
Is that code outside of WordPress? Asking because of https://github.com/WordPress/wordpress-develop/blob/d2f420493d87465c323190892d4e7a6e4c6070a7/src/wp-includes/compat.php#L440.

dd32:
Might as well bump it to PHP 8.1 to match production.  (I don't think the fatal described in the OP is possible, but this is for unrelated reasons)

PR #548, opened 2023-01-09
Search: Filter the search results based on block types

### LINKED ISSUE #547: Pattern API: Search requests ignore other API parameters
See https://github.com/WordPress/pattern-directory/pull/540#issuecomment-1375377118

> I'm testing the GB PR and it seems can fetch patterns with more blocks, if we also provide the search param, ex:
> ```
> https://api.wordpress.org/patterns/1.0/?search=image&allowed_blocks[]=core/heading&allowed_blocks[]=core/gallery
> ```
> Any insights about this?

Probably due to using ElasticSearch for search results.

Sync the `wpop_contains_block_types` meta and use it for a filter like https://github.com/WordPress/pattern-directory/commit/29f7b31a8c4142f068a2076935fd3583584b27eb.

ryelle:
If the `allowed_blocks` meta filter is set, we should also pass that through to the ElasticSearch filters. Currently, if you make a request via the API with the `allowed_blocks` parameter, it will correctly return only patterns with those blocks. If you add a search term, it skips the filtering and returns all matching patterns. Searches should also return only patterns with matching blocks.

I've already added the `wpop_contains_block_types` meta to the synced data in https://github.com/WordPress/pattern-directory/commit/8ac046deba60d42feeba728e06a4a117743a3df2. This PR adds the filter to the ES query so that it should do the same regex matching to return only patterns with a subset of blocks in the given `allowed_blocks`.

See https://github.com/WordPress/pattern-directory/pull/540#issuecomment-1375377118. Fixes #547.

No idea 😓  this is my best guess based on the ES docs. @iandunn you've worked on the search code here before, how can I test this either locally or on a sandbox?

Note - search queries without `allowed_blocks` should still work, as the pattern directory itself doesn't supply `allowed_blocks`.

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/search.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/search.php
@@ -163,12 +163,16 @@ function modify_es_query_args( $es_query_args, $wp_query ) {
 	// If there is an allowed_blocks meta_query, use it to filter the ES query.
 	if ( isset( $meta_query['allowed_blocks'] ) && ! empty( $meta_query['allowed_blocks']['value'] ) ) {
 		// Parse it out of the regex format.
-		$regex       = $meta_query['allowed_blocks']['value'];
-		$list        = substr( $regex, 3, -6 );
-		$block_types = explode( '|', $list );
+		$regex = $meta_query['allowed_blocks']['value'];
+		$regex = substr( $regex, 1, -1 ); // Strips the ^$, which are not supported in ES regex (and not needed).
 
 		$filter['bool']['must'][] = [
-			'terms' => [ 'meta.wpop_contains_block_types.value.raw' => $block_types ],
+			'regexp' => [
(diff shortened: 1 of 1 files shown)


iandunn:
It's been awhile, but it looks like uploading the code to a sandbox is all that's needed to affect the output of the API response (albeit not in the desired way). I don't think it's possible to test locally, since the server needs to authenticate as w.org to get access to the custom ES index. Adding a PHPUnit test in the api.w.org repo is another way to test, and I'd recommend that too for all the normal reasons.

I'd probably start by using Sense to test the raw ES query; I found that helpful before moving it to PHP code, since I ran into a lot of problems just getting the query right. There's info on accessing Sense in the A8C FG. I think you can get the raw query from `Automattic\Jetpack\Search\Classic_Search\search()`. The `#elasticsearch` channel on A8C is a good place to get input from folks who know it well, but I'm also happy to help where I can.

PR #571, opened 2023-03-29
Fix the screenshot requesting when it doesn't need to.

StevenDufresne:
See: https://github.com/WordPress/wporg-mu-plugins/pull/357 and https://github.com/WordPress/wporg-mu-plugins/commit/1e4975600cb0e617e687b624123cbe0b1d391e4d

This is a copy/paste of the code updates made in the `wordpress\wporg-mu-plugins`. It isn't ideal that we have code in both places, it would be nice to address that, hopefully when we redesign this directory.

See https://github.com/WordPress/wporg-mu-plugins/pull/357 for testing details.


ryelle (approved):
Looks good — I haven't been seeing the same errors that were reported, so I can't confirm the fix, but I don't see anything wrong here.

> It isn't ideal that we have code in both places, it would be nice to address that, hopefully when we redesign this directory.

I agree 💯

Issue #580, opened 2023-05-17
Add select for filtering the pattern directory


richtabor:
Consolidating next steps for https://github.com/WordPress/pattern-directory/issues/395 and https://github.com/WordPress/pattern-directory/issues/542. Let's add a select control for filtering pattern results at a high level. 

This will introduce "Curated" and "Community" top-level filters, where "Curated" patterns are hand-picked to be surfaced front-and-center on the Patterns Directory, and eventually within the Pattern Explorer via https://github.com/WordPress/gutenberg/issues/44611. 

### Tasks 

- [ ] Add a filtering select with the options "Curated" and "Community" above the pattern categories.
- [ ] Move the sort select beside the new filter.
- [ ] Remove context bar in lieu of simpler pattern count to the right of the filters.
- [ ] Set "Curated" as the default value.
- [ ] Improve design metrics re the visuals below (could be a follow-up)

### Visuals

<img width="1361" alt="patterns" src="https://github.com/WordPress/pattern-directory/assets/1813435/8a25e137-e0b4-4d95-844d-3008f2ae3f31">
<img width="1361" alt="2" src="https://github.com/WordPress/pattern-directory/assets/1813435/0857b093-bad9-4b38-85c3-da72a44e9367">

richtabor:
Note that the existing "Featured" category is not the same thing as this new "Curated" filter. Curated patterns should be available via the Pattern Explorer—but not necessarily within the "Featured" category. 

@ryelle [detailed the conditions](https://github.com/WordPress/pattern-directory/issues/542#issuecomment-1476777370) how this currently works via a `featured` tag (I suppose the tag itself is a bit misleading).

richtabor:
I think we can close https://github.com/WordPress/pattern-directory/issues/542 and https://github.com/WordPress/pattern-directory/issues/395 in favor of this implementation issue.

ryelle:
> (I suppose the tag itself is a bit misleading)

The whole this is confusing 😣 So a few questions to try to clear up some of that… "Curated patterns should be available via the Pattern Explorer" — is that different from the patterns that are by-default loaded in core? With this question, I'm trying to decide if we actually need the new `featured` keyword, or if we can stick with labeling everything `core`.

What is the purpose of the Featured category after this change, is it just a subset of curated patterns?

>  Remove context bar in lieu of simpler pattern count to the right of the filters.

[...]

PR #583, opened 2023-06-05
Pattern API: Add `curation` parameter 

ryelle:
Add a new `curation` parameter to the API, which can be one of `all`, `core`, or `community`. It defaults to `all` in the API so that requests continue to work as expected.

See #580.

This parameter is also supported by the api.w.org endpoint, which will be used by the pattern explorer some time in the future (see https://github.com/WordPress/gutenberg/issues/44611)

This is an example request that will show "community" patterns.

`http://localhost:8888/wp-json/wp/v2/wporg-pattern?locale=en_US&curation=community`

Try the request with different `curation` values:

- `core` should return only patterns with the `core` keyword
- `community` should return only patterns without the `core` keyword (check `keyword_slugs` if you're not sure)
- `all` should return all patterns, no filtering
- any other value should trigger a validation error

You can also test on a sandbox, `https://api.wordpress.org/patterns/1.0/?locale=en_US&curation=core`

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -646,6 +646,7 @@ function disable_block_directory() {
  * Filter the collection parameters:
  * - set a new default for per_page.
  * - add a new parameter, `author_name`, for a user's nicename slug.
+ * - add a new parameter, `curation`, to filter between curated, community, and all patterns.
  *
  * @param array $query_params JSON Schema-formatted collection parameters.
  * @return array Filtered parameters.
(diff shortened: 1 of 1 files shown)


outdoor2kode (approved):
Tested on a sandbox and LGTM 👍

public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
```diff
@@ -716,6 +728,28 @@ function filter_patterns_rest_query( $args, $request ) {
...
+	// If `curation` is passed and either `core` or `community`, we should
+	// filter the result. If `curation=all`, no filtering is needed.
+	if ( isset( $request['curation'] ) ) {
+		if ( 'core' === $request['curation'] ) {
+			// Patterns with the core keyword.
+			$args['tax_query']['core_keyword'] = array(
+				'taxonomy' => 'wporg-pattern-keyword',
+				'field'    => 'slug',
+				'terms'    => 'core',
+				'operator' => 'IN',
```
StevenDufresne, line 740: This looks like the only difference between the queries? If so, is this worth simplifying?

ryelle: The operator is the only difference, but I like that this keeps the full sub-query together.

StevenDufresne: Yep. 👍

PR #584, opened 2023-06-05
Patterns: Add a curation filter to the UI

### LINKED ISSUE #580: Add select for filtering the pattern directory

ryelle:
This adds a new filter to the UI for "curation", as described in #580— "curated" patterns are those which have been picked to auto-load into core. The new filter is a dropdown with "curated" and "community", which lets the viewer switch between those  two views. It defaults to curated on home & category archives, and uses curated for the empty state recommendations.

The author archive, favorites, and my patterns don't have this filter, and default to showing "all" patterns.

Fixes #580

Note, this is based off #583, which adds the parameter to the API endpoint.

### Screenshots

I applied this to my sandbox to take screenshots, so this is using production patterns.

|  | Curated | Community |
|---|---|---|
| Home, default sort | home-curated | home-community |
| Home, sorted by popular | home-curated-popular | home-community-popular |
| Category, default sort | category-curated | category-community |

<details>
<summary>Favorites & My Patterns have no changes, but screenshots for reference</summary>

favorites

my-patterns

</details>

On medium-smaller screens, the filters wrap, then at 782px the menu collapses.

This new change means that the Call to Action menu item defaults to empty (https://wordpress.org/patterns/categories/call-to-action/), but you can see here the empty state also uses the "curated" patterns.

home-curated-cta-empty

1. Check out the branch and build the project
2. Make sure you have some patterns with the `core` keyword: `/wp-admin/edit.php?wporg-pattern-keyword=core&post_type=wporg-pattern&post_parent=0&post_status=publish` — if not, you should be able to add some via wp-admin
3. View the homepage, it should default to the core patterns
4. Try using the filters on the UI to switch to community
5. Navigate through categories, it should persist your chosen curation
6. My Patterns, My Favorites, and author archives should still work, without a curation filter

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
+++ b/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php
@@ -669,7 +669,7 @@ function filter_patterns_collection_params( $query_params ) {
 	$query_params['curation'] = array(
 		'description' => __( 'Limit result to either curated core, community, or all patterns.', 'wporg-patterns' ),
 		'type'        => 'string',
-		'default'     => 'core',
+		'default'     => 'all',
 		'enum'        => array(
 			'all',
 			'core',
--- a/public_html/wp-content/themes/pattern-directory/css/components/_components.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_components.scss
@@ -32,12 +32,12 @@
 @import "pattern-menu";
 @import "pattern-navigation-layout";
 @import "pattern-notice";
-@import "pattern-order-select";
 @import "pattern-pagination";
 @import "pattern-preview";
 @import "pattern-report-button";
 @import "pattern-report-modal";
 @import "pattern-search";
+@import "pattern-select-control";
 @import "pattern";
 @import "patterns-favorites";
 @import "site-content";
--- a/public_html/wp-content/themes/pattern-directory/css/components/_pattern-navigation-layout.scss
+++ b/public_html/wp-content/themes/pattern-directory/css/components/_pattern-navigation-layout.scss
@@ -1,6 +1,7 @@
 .pattern-navigation-layout {
 	margin: 0 auto $gutter-default;
 	max-width: $size__site-main;
+	gap: $gutter-default;
 	display: flex;
 	flex-direction: column;
 	justify-content: space-between;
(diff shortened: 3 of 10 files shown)


jasmussen:
Looks good to me!

fcoveram:
Same. It looks good 👍

outdoor2kode (approved):
Tested on sandbox and worked as described. Codes-wise look good to me as well 👍 

> This new change means that the Call to Action menu item defaults to empty (https://wordpress.org/patterns/categories/call-to-action/), but you can see here the empty state also uses the "curated" patterns.

Does it need to make any changes to it accordingly, maybe rephrase the text under `No results found`? Any better ways to handle this?

StevenDufresne:
I probably missed this in the [original issue](https://github.com/WordPress/pattern-directory/issues/580), but does this mean I can never actually see "all" the patterns? I have to either subsection them by "community/curated"? 

What if I don't care about whether they are community or curated and just want them all?

annezazu:
Good point. I can see where folks would want to see all and we could still prioritize by curated. What do you think @richtabor? This looks good as an initial iteration though.

jasmussen:
The missing "all" is very much intentional, and an attempt to bring a little structure to what is currently an unfortunate kitchen sink. I like to think that as we roll out potential features in the future like author patterns, pattern bundles, clearer guidelines for each category, related patterns etc, we can see what that does to the curation and then revisit.

ryelle:
Technically, we do have the ability to show `all`, just no UI to trigger it. It's not prioritized by curated (because that introduces a new sort order variable), but it does exist. In case we want to bring it back (now or later).

PR #620, opened 2024-02-01
Change the Translated pattern cron task to run many smaller tasks

dd32:
Recently the cron task 'pattern_import_translations_to_directory' has been running into memory limits.

While this shouldn't happen, it makes sense to just split this task up into many smaller jobs.


pkevan:
seems this change caused some fatal errors, which I thought were due to `glotpress` not being activated, but doing so caused the crons to fail.

@akirk committed the following to hopefully fix it: https://github.com/WordPress/wordpress.org/commit/78e808b5dfeb396d8ecc3e2ae9eafda29adc82db, and the plugin is now not activated on the site.

Issue #642, opened 2024-03-22
Block theme: Add drag handles to pattern previewer


ryelle:
The pattern previewer should have drag handles for the preview width.

<img width="307" alt="Screenshot 2024-03-22 at 1 20 52 PM" src="https://github.com/WordPress/pattern-directory/assets/541093/e9660967-d70d-4a39-8b6a-e5e2e825655e">

ryelle:
Added with #650

PR #650, opened 2024-04-01
Block theme: Add drag handles to resize pattern preview

ryelle:
See #642. This updates the `wporg/pattern-view-control` block to include drag handles for resizing the preview. Since the new version of the block uses the Interactivity API, not React components, we can't "just port over" the old functionality, and we can't use the gesture library. I've made it work fairly well with vanilla JS + interactivity directives, but it still seems janky to me.

- When the drag handle is clicked (mousedown), it triggers the "dragging" state, and tracks the current x-position & which handle was clicked
- On any mousemove, if we're in dragging state, check the difference in x-position to see which way we should resize and update the previewWidth accordingly
    - On large screens, the iframe width doesn't update correctly, so the preview is pushed to the side
    - There's a throttle timeout, but it's just 10ms — any higher and the resizing was jumpy
- On any mouseup, it resets the dragging state back to the initial values

For keyboard users, they can focus on the drag handles, and use the left/right arrow keys to change the preview width. This matches the current behavior.

I did not use the HTML `draggable` API because it created a ghost-button while dragging, and it seemed to not work on the `button` element.

In the design, the toggles are only 4px wide. When testing, I had a lot of trouble targeting that size, so I've changed them to 8px.

⚠️  **Known issues, would appreciate opinions:**

- I had to give the whole preview a fixed height, otherwise the middle-aligned drag buttons jump around, so now the patterns are cut off on the wide/medium sizes (or have extra whitespace, if the pattern is short).
- The mousemove event over iframe only works when `pointer-events: none` or there's an overlay, but then we can't scroll inside the window.
- The aforementioned width issue on large screens (shown in the video below)

### Screenshots

Screen Shot 2024-04-01 at 16 28 13

On a large screen, resize up and then back down.

https://github.com/WordPress/pattern-directory/assets/541093/245961d7-4f9f-4b42-9048-18dbc6d5a4f3

On small screen, resize from the "mobile toggle" up to large screen, which scales down the preview (but not the frame height).

https://github.com/WordPress/pattern-directory/assets/541093/169c3af9-f570-4956-a6b9-468418b89903

1. Check out the branch and built the project
2. View any pattern
3. You should see drag handles, and be able to drag them to change the preview size
4. Clicking the set buttons should still work
5. Try again with a pattern you own

### DIFF
--- a/public_html/wp-content/themes/wporg-pattern-directory-2024/src/blocks/pattern-preview/controls/render.php
+++ b/public_html/wp-content/themes/wporg-pattern-directory-2024/src/blocks/pattern-preview/controls/render.php
@@ -23,12 +23,16 @@ $p->remove_attribute( 'data-wp-interactive' );
 $p->remove_attribute( 'data-wp-context' );
 $content = $p->get_updated_html();
 
+$html_id = wp_unique_id( 'pattern-preview-help-' );
+
 ?>
 <div
 	<?php echo get_block_wrapper_attributes(); // phpcs:ignore ?>
 	data-wp-interactive="wporg/patterns/preview"
 	data-wp-context="<?php echo esc_attr( $encoded_state ); ?>"
 	data-wp-class--is-mobile-view="state.isWidthNarrow"
+	data-wp-on-window--mousemove="actions.onDrag"
+	data-wp-on-window--mouseup="actions.onDragEnd"
 >
 	<section class="wporg-pattern-view-control__controls wp-block-buttons" aria-label="<?php esc_attr_e( 'Preview width', 'wporg-patterns' ); ?>">
 		<div class="wp-block-button is-style-toggle is-small">
@@ -66,8 +70,37 @@ $content = $p->get_updated_html();
 		</div>
 	</section>
 
-	<?php
-		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Content from child blocks.
-		echo $content;
-	?>
+	<div class="wporg-pattern-preview__drag-container">
+		<div class="wporg-pattern-preview__drag-handle">
+			<button
+				class="wporg-pattern-view-control__drag-handle is-left"
+				aria-label="<?php esc_attr( 'Drag to resize', 'wporg-patterns' ); ?>"
+				aria-describedby="<?php echo esc_attr( $html_id ); ?>-left"
+				data-direction="left"
+				data-wp-on--keydown="actions.onLeftKeyDown"
+				data-wp-on--mousedown="actions.onDragStart"
+			></button>
+		</div>
+		<?php
+			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Content from child blocks.
+			echo $content;
+		?>
+		<div class="wporg-pattern-preview__drag-handle">
+			<button
+				class="wporg-pattern-view-control__drag-handle is-right"
+				aria-label="<?php esc_attr( 'Drag to resize', 'wporg-patterns' ); ?>"
+				aria-describedby="<?php echo esc_attr( $html_id ); ?>-right"
+				data-direction="right"
+				data-wp-on--keydown="actions.onRightKeyDown"
+				data-wp-on--mousedown="actions.onDragStart"
+			></button>
+		</div>
+	</div>
+
+	<span id="<?php echo esc_attr( $html_id ); ?>-left" class="screen-reader-text">
+		<?php esc_attr_e( 'Drag or use arrow keys to resize the pattern preview. Left to make larger, right to make smaller.', 'wporg-patterns' ); ?>
+	</span>
+	<span id="<?php echo esc_attr( $html_id ); ?>-right" class="screen-reader-text">
+		<?php esc_attr_e( 'Drag or use arrow keys to resize the pattern preview. Right to make larger, left to make smaller.', 'wporg-patterns' ); ?>
+	</span>
 </div>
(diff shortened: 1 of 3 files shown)


StevenDufresne:
I checked out this PR and it is working pretty well.

> I had to give the whole preview a fixed height, otherwise the middle-aligned drag buttons jump around, so now the patterns are cut off on the wide/medium sizes (or have extra whitespace, if the pattern is short).

I like the fixed height and feel like this should probably be the default behavior.

>The mousemove event over iframe only works when pointer-events: none or there's an overlay, but then we can't scroll inside the window.

I don't understand.

> On large screens, the iframe width doesn't update correctly, so the preview is pushed to the side

Does this depend on how the pattern was made? Seems to be working for me.

Screen Capture on 2024-04-02 at 13-15-42

On this front, I think we can have an outer bound for the width. I don't see a use case for extending the width indefinitely.

A couple of small points:
- The handle target area is pretty small
- An active and inactive state for the handle would improve UX.
  - We should update the cursor appropriately.
  - We could change the color and make the handle larger. 

<hr>

I appreciate the effort on this PR. 🙇 

I do however question the value of this feature and wonder if provides enough _extra_ context for users to find it valuable and use it. What if we shipped without it and added it if requested?

jasmussen:
Nice work. Is there an easy way to test this? Visually this looks close. Charcoal 5, pill shaped, 4x52px:

Screenshot 2024-04-02 at 10 50 13

I would keep that _optical_ handle size, though naturally we want the hit area to be big enough to be comfortable. So if you can make it bigger in practice, that would be fine. E.g. something like:
Screenshot 2024-04-02 at 10 52 33

That's 14x52 just as an example. Can be even bigger if you can find the space for it.

Here's how big the hit area is in the block editor, which is one of the patterns we're following for this:
handle

ryelle:
To @StevenDufresne's comment:

> I like the fixed height and feel like this should probably be the default behavior.

Yeah, I think I'm outvoted 🤷🏻  I like being able to see the whole pattern at once.

> > The mousemove event over iframe only works when pointer-events: none or there's an overlay, but then we can't scroll inside the window.
>
> I don't understand.

For example, this pattern is taller than 600px, so it's cut off. But I can't scroll, it only scrolls the full page. If I disable `pointer-events: none`, then I can scroll to see the full pattern. But now trying to scale the pattern, it stops when I'm over the pattern preview (in fact, I released the click while on the pattern, but that did not trigger the drag state to end).

https://github.com/WordPress/pattern-directory/assets/541093/2d3ff6f4-85e8-40fe-ab79-e8615110a746

However, this might just be a FF issue, since scrolling seems to work fine with `pointer-events: none` on Chrome.

> Does this depend on how the pattern was made? Seems to be working for me.

Ah, I accidentally missed rebasing the PR on the layout updates. In your version the preview is constrained by the wideSize (1160), as you see by the container not actually expanding. If you try on the rebased PR (pushed now), you'll see it like the screen recording.

> On this front, I think we can have an outer bound for the width. I don't see a use case for extending the width indefinitely.

We do, it's 2400px. Happy to change that though, I just went with an arbitrary large size. Anything over 1200 will cause the issue, though.

> I do however question the value of this feature and wonder if provides enough extra context for users to find it valuable and use it. What if we shipped without it and added it if requested?

I agree that it's not very valuable, and I was trying to simplify the preview by leaving it out. IMO the 3 size toggles provide enough preview capability.

To @jasmussen's comment

> Is there an easy way to test this?

You can use wp-env like all the other redesign repos, [the instructions are in the project readme](https://github.com/WordPress/pattern-directory/blob/trunk/readme.md).

> I would keep that optical handle size, though naturally we want the hit area to be big enough to be comfortable. 

I can try this, but it's strange to me that the click target will be larger— it's still going to feel like you need to hit a tiny line.

ryelle:
So I made a fuss on the call about how hard fixing this up would be, but I think coming back to this with fresh eyes and no pressure was the solution 😅 

I spent some time noodling on it and I think I've fixed all the previous weird edge cases.

On desktop sizes, this works well — grabbing the drag handles can expand up to 1400, down to 320, and the switching between drag and toggles works.

https://github.com/WordPress/pattern-directory/assets/541093/46dedb5c-3d1a-4532-b26d-eead9d20a8b7

Even with keyboard, the toggles work and left/right arrow keys when the handles are focused will resize the view (same as production).

https://github.com/WordPress/pattern-directory/assets/541093/cd6bf66c-3a90-4909-bb31-245684327894

It does get awkward on small screens/mobile, since you can "drag" wider than the screen, and shrinking down from a big preview only changes the preview's scale. So the handle & cursor no longer match. The production site gets around this by not allowing a larger preview than the current screen size, but I think that's a silly restriction— a user on a phone should be able to preview desktop size. And with the toggles this works fine. Maybe we should hide the handles at some screen size?

Also, I haven't tested this with an actual phone or tablet, so I'm not totally sure how the touch interface will work (trying production on my phone though, it doesn't work well, another point to hiding it at small sizes).

https://github.com/WordPress/pattern-directory/assets/541093/da91b2f8-2b8a-49c3-89ec-2c4a7088ed18

So I'd actually be comfortable merging this before launch after all, provided it passes others' QA checks, and then thinking through the small screen behavior once it's on the preview site — should be enough time before the proposed launch next week.

StevenDufresne:
> Maybe we should hide the handles at some screen size?

I think this would be fine. We can cycle back if/when this feature is requested.

jasmussen:
Nice work! I'd be happy to launch with this, and even hide those handles at small screen sizes.

fcoveram:
It looks very nice ✨ And agree with Joen, we could hide the handlers.

PR #655, opened 2024-04-02
Block: Pattern Thumbnail: Fix loading and error state

### LINKED ISSUE #652: Block Theme: Broken image links.
**Describe the bug**
I'm getting broken images for some of the patterns at a pretty high rate.

We should investigate why and consider a better error view.

**Screenshots**

Screenshot 2024-04-02 at 11 51 32 AM

 - OS: MacOs
 - Browser: Chrome

ryelle:
Fixes #652 — This updates the pattern preview block to use context for the initial block state (ex, `attempts`). Previously the state.attempts value was global, so each pattern on the page incremented it. By the time the last patterns on the page are rendered, the `attempts` value would already be >10, so it would only try the initial fetch and if the mshots image did not exist, it would error - no retries.

This updates the `attempts` to behave as expected, 10 attempts per pattern. This should cut down on the number of errors per page.

The error display has also been updated, in the case where there are issues. Happy to tweak this if there are any suggestions @StevenDufresne @WordPress/meta-design (screenshot below)

### Screenshots

Screenshot 2024-04-02 at 5 01 34 PM

1. View the pattern archives
2. The previews should load successfully
3. If there are any errors, it should not show the broken `` tag

To check that this is loading correctly from scratch, you could update `$cache_key` to something random, which will force mshots to re-take the screenshots, and take longer for loading.

To intentionally trigger an error state, update the `$cache_key` (render.php) and change `MAX_ATTEMPTS` (view.js) to 1 or 2.

Remember to rebuild if you make those changes.

### DIFF
--- a/public_html/wp-content/themes/wporg-pattern-directory-2024/src/blocks/pattern-thumbnail/render.php
+++ b/public_html/wp-content/themes/wporg-pattern-directory-2024/src/blocks/pattern-thumbnail/render.php
@@ -36,6 +36,9 @@ $init_state = [
 	'base64Image' => '',
 	'src' => esc_url( $url ),
 	'alt' => the_title_attribute( array( 'echo' => false ) ),
+	'attempts' => 0,
+	'shouldRetry' => true,
+	'hasError' => false,
 ];
 $encoded_state = wp_json_encode( $init_state );
 
--- a/public_html/wp-content/themes/wporg-pattern-directory-2024/src/blocks/pattern-thumbnail/render.php
+++ b/public_html/wp-content/themes/wporg-pattern-directory-2024/src/blocks/pattern-thumbnail/render.php
@@ -72,6 +72,7 @@ if ( $has_link ) {
 		/>
 		<span
 			data-wp-bind--hidden="state.base64Image"
+			data-wp-class--screen-reader-text="!state.base64Image"
 			data-wp-text="context.alt"
 		></span>
 	</div>
--- a/public_html/wp-content/themes/wporg-pattern-directory-2024/src/blocks/pattern-thumbnail/style.scss
+++ b/public_html/wp-content/themes/wporg-pattern-directory-2024/src/blocks/pattern-thumbnail/style.scss
@@ -41,8 +41,7 @@
 		justify-content: center;
 		height: 100%;
 
-		img,
-		span {
+		img {
 			display: none;
 		}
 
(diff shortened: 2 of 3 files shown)


jasmussen:
Is this the error state?

screenshot of supposed error state

If yes, then I wonder, what might cause that error state to happen? Is this if the thumbnail doesn't load properly? I forgot what error state we use for the Showcase thumbnails, but we should probably use the same here, and omit the duplicated link-text. If the error state for the Showcase is not appropriate, we should use a broken image icon, something like:

Screenshot 2024-04-03 at 08 58 26

SVG:

```
<svg width="24" height="24" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M5 3a2 2 0 0 0-2 2v14a2 2 0 0 0 2 2h5l-.121-.121 1.379-1.379H5a.5.5 0 0 1-.5-.5v-2.436l4.096-2.987 2.996 1.941a.75.75 0 0 0 .93-.091L16 12.046l1.375 1.337 1.06-1.06-1.912-1.86a.75.75 0 0 0-1.046 0l-3.571 3.471-2.927-1.897a.75.75 0 0 0-.85.024L4.5 14.707V5a.5.5 0 0 1 .5-.5h14a.5.5 0 0 1 .5.5v6.258l1.5-1.5V5a2 2 0 0 0-2-2H5Zm16 11-1.5 1.5V19a.5.5 0 0 1-.5.5h-3.5L14 21h5a2 2 0 0 0 2-2v-5Z" fill="#1E1E1E"/></svg>
```

Shown at 24x24px.

ryelle:
> If yes, then I wonder, what might cause that error state to happen? Is this if the thumbnail doesn't load properly?

Yes - though it's pretty rare since we retry fetching the image 10 times.

>  I forgot what error state we use for the Showcase thumbnails

I'm pretty sure the error state for Showcase is just the site title text (so like this, but without the icon). Since all the sites got custom screenshots (which would never show this), we never got back to requesting design for an error state.

I'll update the image, and see what i can do about the text— we need some text there for accessibility (since it's a link), but perhaps I can hide it better.

ryelle:
Updated the text & icon, it now looks like this if an image fails to load:

Screenshot 2024-04-03 at 2 31 43 PM

jasmussen:
Nice!

StevenDufresne:
I'm getting 403's from mShots making it hard to test but since they are from mShots it's probably a fair representation of this branch. The current sequence shows `broken image > loader > image`. It becomes more pronounced when I add throttling.

The code looks correct from a logic standpoint. 

Screen Capture on 2024-04-04 at 09-18-36

ryelle:
I think you're getting the broken image before the Interactivity API kicks in, so I've tweaked the CSS a bit to hide the image by default. So now you should see the empty grey box -> loader -> image.

I'm going to merge this PR, we can iterate if there are edge cases once it's on production.

StevenDufresne:
I'm still getting this on production:

Screen Capture on 2024-04-15 at 13-43-08

PR #656, opened 2024-04-03
Block theme: Update filters with new single-select behavior

### LINKED ISSUE #648: Block Theme: Filters: Default state and clear action in Sort
**Describe the bug**
I noticed an odd interaction while sorting on the homepage. The default sort is by popularity, but when clicking on it, I'm able to clear the action and set an "empty" sort. This flow was thought to be for applying filters, and it seems this case inherits the same logic.

I am drawn to always keep one sort enabled, as in radio, where users can not deselect. "Apply" would still exist as it confirms the change, but "clear" is no longer needed.

**To Reproduce**
1. Go to the homepage
2. Click on "Sort: Popular"
3. Click on "Clear" and then on "Apply" to confirm the change
4. See how the label shows "Sort" only.

**Video recording**

https://github.com/WordPress/pattern-directory/assets/895819/4d831032-4e9b-4ea1-8e40-dce2d09659b8

ryelle:
Fixes #648, along with https://github.com/WordPress/wporg-mu-plugins/pull/596.

The Query Filter block was updated in https://github.com/WordPress/wporg-mu-plugins/pull/596, and this PR updates the configuration to work with the new expected behavior.

- The label prefix ("Sort"/"Filter") is removed from the buttons, and is now visible in the dropdown
- Sort now has a visible default value, so "Newest" is selected when no filters are applied (this is the WP_Query default)
- Filter now has an "All" option so that something can be selected on the "All patterns" page

### Screenshots

|  | Screenshot | 
|---|---|
| Home | home-default |
| "Filter" open | home-filter-open |
| "Sort" open | home-sort-open |
| All patterns | all-patterns-default |
| All patterns, with filters applied | all-patterns-filtered |

1. Apply https://github.com/WordPress/wporg-mu-plugins/pull/596 to wporg-mu-plugins
2. View various grid pages, the filters should show up and work as expected
3. You can no longer clear a filter, so something should always appear selected

### DIFF
--- a/public_html/wp-content/themes/wporg-pattern-directory-2024/inc/block-config.php
+++ b/public_html/wp-content/themes/wporg-pattern-directory-2024/inc/block-config.php
@@ -132,6 +132,9 @@ function get_curation_options( $options ) {
 		case 'core':
 			$label = __( 'Curated', 'wporg' );
 			break;
+		default:
+			$label = __( 'All', 'wporg' );
+			break;
 	}
 
 	// Show the correct filters on the front page.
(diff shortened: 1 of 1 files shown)


jasmussen:
Looks good.

PR #679, opened 2024-04-18
Update block-config.php

pkevan:
Add check for string values, otherwise a php warning occurs.


pkevan:
Example:

```
E_WARNING: strtolower() expects parameter 1 to be string, array given in [..]/wp-content/themes/wporg-pattern-directory-2024/inc/block-config.php:152
strtolower() expects parameter 1 to be string, array given
Source: GET https://ps.wordpress.org/patterns/page/48/?curation[]=all&s=1
```

ryelle (approved):
Looks good, thanks for the fix!

PR #682, opened 2024-04-22
Add results check for URL field

pkevan:
`get_term_link` can return WP_Error if there is no matching term (https://developer.wordpress.org/reference/functions/get_term_link/) - this additional check protects against this and the warnings produced from attempting to escape this.

if we want to produce a standard URL i.e. home_url this could also be possible, although unsure what the implications are for doing so i.e. indexing the URLs etc

### DIFF
--- a/public_html/wp-content/themes/wporg-pattern-directory-2024/functions.php
+++ b/public_html/wp-content/themes/wporg-pattern-directory-2024/functions.php
@@ -475,12 +475,18 @@ function add_social_meta_tags() {
 			'og:image'       => esc_url( $default_image ),
 		];
 	} else if ( is_tax() ) {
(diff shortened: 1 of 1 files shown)


ryelle:
I think this will always be a "no results" page, so I would rather not output any meta tags — I think @dd32's solution in #681 would be better (checking before we're in the conditional), what do you think of that PR?

pkevan:
> what do you think of that PR?

I should have checked for existing PRs first 😄 

@dd32's solution is neater, just not sure on the implication of having them not there - I'd guess it's probably fine, but that was my concern.

ryelle:
Thanks, I've merged #681, so I'll close this one.

PR #697, opened 2024-07-02
Remove favorite button block, use shared block

ryelle:
This removes the favorite block in favor of the new shared favorite button. Unlike https://github.com/WordPress/wporg-theme-directory/pull/148, I did not remove the favorite endpoints here — they're not bundled with the theme, and while I don't think they're used anywhere else, I wanted a second opinion before just pulling them out.

See https://github.com/WordPress/wporg-mu-plugins/discussions/617

**To test**

- Use the branch here https://github.com/WordPress/wporg-mu-plugins/pull/635
- Make sure to build both projects
- View the pattern directory
- There should be small heart icons on each pattern in the grid, with counts
- If you are logged in, those are clickable; if not, they're just text
- View a single pattern
- You should see a button in the theme header with a count
- If you are logged in, it should be are clickable; if not, it's just text
- You should be able to click or trigger with a keyboard to favorite/unfavorite items, and the button should update with the interaction
- Reload the page or view /favorites to ensure the favorite saved correctly


adamwoodnz:
Updated my local env to test this, but was initially blocked because the parent theme wasn't installed. I think we need to add it to composer.json 🙂

colorful-tones:
> Updated my local env to test this, but was initially blocked because the parent theme wasn't installed. I think we need to add it to composer.json 🙂

I opened #701 to try and address this.

PR #700, opened 2024-07-08
Update string on private APIs

pkevan:
Changed in https://github.com/WordPress/gutenberg/pull/62635

Broke pattern creator, as reported in: https://wordpress.slack.com/archives/C02QB8GMM/p1720180116195859


ryelle:
Thanks for fixing this!

PR #702, opened 2024-07-14
Add parent theme to local env dependencies

### LINKED ISSUE #701: Missing parent theme: wporg-parent-2021
**Describe the bug**
Attempting to get spun up on a local development environment. However, it seems that the parent theme is missing. Do we need the `wporg-parent-2021` missing dependency? Perhaps there is a newer parent theme we should be referencing for `Template:`?

https://github.com/WordPress/pattern-directory/blob/99d71f9b55ab923cef70c263641e5686f1a16305/public_html/wp-content/themes/wporg-pattern-directory-2024/style.css#L10

**To Reproduce**
1. `yarn run create`
2. Visit http://localhost:8888/wp-admin/themes.php

**Expected behavior**
The front end output does not match the production Pattern Directory, which certainly reinforces the issue of missing parent theme. 

**Screenshots**
Screenshot 2024-07-12 at 2 13 57 PM

**Desktop:**
 - OS: macOS 14.5 (Sonoma)
 - Browser Chrome 126
 - Docker Desktop 4.32.0

adamwoodnz:
See https://github.com/WordPress/pattern-directory/pull/697#issuecomment-2209622721

Fixes #701

Props @colorful-tones

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,6 +58,10 @@
 		},
 		{
 			"type": "vcs",
+			"url": "git@github.com:WordPress/wporg-parent-2021.git"
+		},
+		{
+			"type": "vcs",
 			"url": "git@github.com:WordPress/wporg-internal-notes.git"
 		}
 	],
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -78,7 +82,8 @@
 		"rmccue/requests": "^1.8.1",
 		"wp-coding-standards/wpcs": "2.*",
 		"wp-phpunit/wp-phpunit": "~6.0",
-		"yoast/phpunit-polyfills": "^1.1"
+		"yoast/phpunit-polyfills": "^1.1",
+		"wporg/wporg-parent-2021": "dev-build"
```
ryelle: These are sorted alphabetically (or they should be, I see the `require` is not _quite_ right).

```suggestion
		"wporg/wporg-parent-2021": "dev-build",
		"yoast/phpunit-polyfills": "^1.1"
```

ryelle (approved):
Thanks for fixing this!

PR #708, opened 2024-09-10
Navigation: Add "Log in" link to local nav

ryelle:
This updates the Pattern Directory to include the "Log in" link in the local navigation across the site. The link only appears when you're logged out, once logged in, you can manage your account/login status in the admin bar. Once logged in, the "My patterns" link replaces the "Log in" link.

This site already has no admin bar when logged out.

See https://github.com/WordPress/wporg-mu-plugins/issues/647

### Screenshots

| Logged out | Logged in |
|--------|-------|
| Screen Shot 2024-09-10 at 15 44 32 | Screen Shot 2024-09-10 at 15 46 22 |

1. Be logged out
2. There should be a "log in" link
3. Clicking it and signing in should redirect you back to the page you were just on
4. The "log in" link should not appear
5. You should see the admin bar

cc @WordPress/meta-design

### DIFF
--- a/public_html/wp-content/themes/wporg-pattern-directory-2024/inc/block-config.php
+++ b/public_html/wp-content/themes/wporg-pattern-directory-2024/inc/block-config.php
@@ -358,6 +358,13 @@ function add_site_navigation_menus( $menus ) {
 			'label' => __( 'My patterns', 'wporg-patterns' ),
 			'url' => '/my-patterns/',
 		);
+	} else {
+		global $wp;
+		$redirect_url = home_url( $wp->request );
+		$menu[] = array(
+			'label' => __( 'Log in', 'wporg-patterns' ),
+			'url' => wp_login_url( $redirect_url ),
+		);
 	}
 
 	$current_status = isset( $wp_query->query['status'] ) ? $wp_query->query['status'] : false;


StevenDufresne:
I don't want to go against the momentum, but it's weird to me that the login link disappears and isn't replaced with my user account image or link to my account.

Apart from that, the code looks good. The way we have built our subnav will force duplicate code across all the themes. We may want to consider adding a filter or creating a block for this longer term.

adamwoodnz (approved):
Works for me 💯 

> The way we have built our subnav will force duplicate code across all the themes. We may want to consider adding a filter or creating a block for this longer term.

Feels like the local nav could probably add this link by default, maybe with a block attribute to disable.

jasmussen:
> I don't want to go against the momentum, but it's weird to me that the login link disappears and isn't replaced with my user account image or link to my account.

I wouldn't be opposed to future explorations that add profile management to that toolbar. We might even be able to hide the adminbar for most user roles other than editor and admin. In that vein, I think of this as the first step, which will fix an important layout shift that happens currently to every visitor to the site that isn't logged in, and never will be, which is arguably the majority. I believe @fcoveram even has some mockups for further profile management; just important to not block this important first step.

ryelle:
> Feels like the local nav could probably add this link by default, maybe with a block attribute to disable.

I thought about trying something like that, but all the themes use different menu slugs, and it's not requested for all the sites, so I couldn't auto-add it. An attribute could work, but some links need the `has-separator` class and some don't… overall it seemed easier to drop it in to each theme.

If we do come back to add in profile management (which, IMO, would need a lot of design and UX research to _replace_ the admin bar, otherwise we'll have the same "layout jump" issue as today), we can iterate on this. For example, we might want a separate block for those user controls, not part of the nav menu.

PR #711, opened 2024-10-01
Patterns API: Fix the `category_slugs` format for featured patterns

ParhamG:
This fixes a regression caused via https://github.com/WordPress/pattern-directory/pull/704 that returns `category_slugs` as an object instead of an array for patterns that has the `featured` category.

This is breaking the Patterns tab in the block editor for all WP sites.

The modified array returned by `array_diff()` can have non-consecutive indexes, causing it to be JSON-encoded as an object rather than array which is the expected format. Applying `array_values()` to the response fixes it.

1. View the response for https://api.wordpress.org/patterns/1.0/?page=1&per_page=100&order=desc&orderby=date&locale=en_US&wp-version=6.6.1&pattern-keywords=11
2. Lookup the pattern with `"id": 199` 
3. The value for its `"category_slugs"` attribute must be an array and not an object


ryelle (approved):
Thanks for the fix and explanation!

PR #712, opened 2024-10-07
Pattern directory: Add pattern endpoint test matching schema

ryelle:
We recently had an issue causing errors on client WP sites because the Patterns API returned malformed category data — see #711. We did not have any tests to catch that change. This adds a very simple test to check the pattern API response against its own schema. This is not a perfect check (since the validation normalizes arrays), so I've also added some extra checks to make sure the pattern content is never empty, and the two [array values used in core](https://github.com/WordPress/WordPress/blob/master/wp-includes/rest-api/endpoints/class-wp-rest-pattern-directory-controller.php#L191-L200) (`category_slugs` & `meta.wpop_block_types`) are sequential arrays.

See related tests in https://github.com/WordPress/wordpress.org/pull/393

1. Start up the wp-env
2. Run `yarn test:php` or `yarn test:php -- --group rest-api` for just this test
3. It should pass ✅ 
4. Comment out the `array_values` fix in [plugins/pattern-directory/includes/pattern-post-type.php](https://github.com/WordPress/pattern-directory/blob/a47c46cbbcc18731da5963bb3f8ccd9f38b6a4be/public_html/wp-content/plugins/pattern-directory/includes/pattern-post-type.php#L298)
5. Run the test again
6. It should fail ❌

### DIFF
--- a/public_html/wp-content/plugins/pattern-directory/tests/phpunit/endpoint-wporg-pattern-test.php
+++ b/public_html/wp-content/plugins/pattern-directory/tests/phpunit/endpoint-wporg-pattern-test.php
@@ -42,7 +42,12 @@ class Endpoint_Wporg_Pattern_Test extends WP_UnitTestCase {
 	}
 
 	/**
-	 * Verify the pattern response matches the schema.
+	 * Verify the pattern response matches the schema, plus strict type checking
+	 * for the array values.
+	 *
+	 * `rest_validate_value_from_schema` will check most values, but it also
+	 * "normalizes" array values to associative arrays, which does not happen
+	 * in practice, so we need to manually test those values.
 	 */
 	public function test_pattern_directory_api() {
 		$request = new WP_REST_Request( 'GET', '/wp/v2/wporg-pattern/' . self::$pattern_id );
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/pattern-directory/tests/phpunit/endpoint-wporg-pattern-test.php
```diff
@@ -0,0 +1,74 @@
...
+		// New request to get schema, so that `rest_api_init` is called (to register the custom endpoint fields).
+		$request = new WP_REST_Request( 'OPTIONS', '/wp/v2/wporg-pattern/' . self::$pattern_id );
+		$response = rest_do_request( $request );
+		$schema = $response->get_data();
+		$schema = $schema['schema'];
+
+		$result = rest_validate_value_from_schema( $pattern, $schema );
+		$this->assertTrue( $result );
+
+		// Pattern content should always exist.
```
StevenDufresne, line 67: Should these comments be passed as messages in the assert functions?

public_html/wp-content/plugins/pattern-directory/tests/phpunit/endpoint-wporg-pattern-test.php
```diff
@@ -0,0 +1,74 @@
...
+		$this->assertNotEmpty( $pattern['pattern_content'] );
+
+		// Check that these arrays are sequential, not associative arrays.
```
StevenDufresne, line 70: Assuming yes, but can't these be empty? Do we need to test for that?

public_html/wp-content/plugins/pattern-directory/tests/phpunit/endpoint-wporg-pattern-test.php
ryelle: Sure, done in https://github.com/WordPress/pattern-directory/pull/712/commits/61c8e979417cf030b7c94390c1901f6d474b971e

public_html/wp-content/plugins/pattern-directory/tests/phpunit/endpoint-wporg-pattern-test.php
ryelle: They can be, and as long as they're arrays it shouldn't cause any issues in the core API, so I don't think we need to test that.

PR #725, opened 2025-07-07
Fix 'core/interface' registration error

### LINKED ISSUE #723: WP Pattern Directory Pattern Editor not working properly due to store core/interface' already registered error
**Describe the bug**
When attempting to create a new pattern using the WordPress Pattern Directory editor, an error "Store 'core/interface' is already registered. i@registry.js:220" appears in the browser's developer console. This error prevents blocks added to the editor from rendering and styling correctly, leading to a broken visual experience where blocks do not appear as expected or lose their intended styling.

**To Reproduce**
1. Go to 'https://wordpress.org/patterns/new-pattern/'
2. Use the editor to add any blocks (e.g., Paragraph, Image, Columns).
3. Observe that the added blocks are not working as expected; their styling is incorrect or they fail to render properly.
4. Open the browser's developer console (usually by pressing F12 or right-clicking and selecting "Inspect" then navigating to the "Console" tab).
5. See the error message: Store "core/interface" is already registered. i@registry.js:220

**Expected behavior**
When using the WordPress Pattern Directory editor, blocks should render and apply their styling correctly without any console errors. The editing experience should be smooth, allowing users to accurately preview the pattern they are creating.

**Screenshots**
Windows - Edge Browser: https://share.zight.com/5zuk0xWP
Windows - Chrome: https://share.zight.com/eDupNq95
MacOS - Chrome: https://share.zight.com/7KulYOdz

- OS: Windows 10/11, macOS (Latest versions)
- Browser: Chrome, Firefox, Edge
- Version: Latest stable versions (e.g., Chrome v138.0.7204.63 , Edge v138 as of July 2025)

gigitux:
Fixes #723.

This PR fixes the `core/interface` registration error, but it doesn't fix the other errors mentioned in the issue. 

This error happens because `core/interface` is a package included in the bundle and not available on the document object. 
Currently, on the `/new-pattern/` route are loaded the `site-editor` scripts and the bundle of the pattern creator editor.
Both of them, under the hood, load the `@wordpress/editor` package and this causes that the `core/interface` store is registered twice. 

With this PR, the `@wordpress/editor` is bundled in the pattern creator editor and the `site-editor` scripts aren't loaded. In this way, the `core/interface` store is registered once. 

At the same time, while this PR can be a good temporary solution, I think that we need to explore https://github.com/WordPress/pattern-directory/pull/719. As mentioned in #719, this will reduce the maintenance cost and fix other issues too.

Props <username>, <username>

1. Visit `/new-pattern/`
2. Open the dev tools.
3. Ensure that `Store "core/interface" is already registered.` isn't visible.

### DIFF
--- a/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
+++ b/public_html/wp-content/plugins/pattern-creator/pattern-creator.php
@@ -182,7 +182,6 @@ function pattern_creator_init() {
 		'after'
 	);
 
-	wp_enqueue_script( 'wp-edit-site' );
 	wp_enqueue_script( 'wp-format-library' );
 	wp_enqueue_style( 'wp-edit-site' );
 	wp_enqueue_style( 'wp-format-library' );
--- a/public_html/wp-content/plugins/pattern-creator/webpack.config.js
+++ b/public_html/wp-content/plugins/pattern-creator/webpack.config.js
@@ -1,4 +1,5 @@
 const defaultConfig = require( '@wordpress/scripts/config/webpack.config' );
+const DependencyExtractionWebpackPlugin = require( '@wordpress/dependency-extraction-webpack-plugin' );
 
 const config = {
 	...defaultConfig,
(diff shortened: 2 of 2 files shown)


ryelle (approved):
Looks good, fixes the error as described.

The failing test is also failing on trunk, and it's two parts— first, the call to `npm run test:unit --workspace=wporg-pattern-directory-theme` can be removed, since the old theme is not used anymore. But then the creator tests also fail with "SyntaxError: Unexpected token 'export'". Out of scope for this PR, and maybe irrelevant with #719 since it removes the separate creator plugin.

gigitux:
Thanks for the review! Should I merge the PR?

@ryelle let me know how I can assist you to make #719 ready to be reviewed!

PR #773, opened 2026-09-14
Pattern directory: extend validation test coverage

obenland:
Follow-up test coverage for the content validation helpers in `pattern-validation.php`.

`content_has_block_directives()` scans both the submitted markup and its sanitised
form, since some elements hide their contents from `WP_HTML_Tag_Processor` while KSES
drops the wrapper on save and keeps what it held. The data providers covered that for
raw-text elements in the HTML namespace, but not for the foreign-content case, which
reaches the same end state by a different route and so is worth pinning separately.

- `pattern-content-validation-test.php` — two cases covering the foreign-content shape,
  closed and unclosed.
- `pattern-translated-content-test.php` — the same shape on the translated-content path,
  which shares the helper but is a separate entry point.

Both cases fail if the sanitised-markup pass is removed from the helper, so they pin the
current behaviour rather than restating it.

Tests only; no change to validation behaviour. Full PHPUnit suite (226 tests) and `phpcs`
both pass locally.


bor0:
Late to this one (already merged), but a note for whoever touches these providers next.

The rows do what the description says — they fail if the sanitised-markup pass is removed — so the coverage is real. The stated reason isn't, though: `<svg>` isn't what hides the anchor. `WP_HTML_Tag_Processor` walks straight into `<svg>`, `<desc>` and `<annotation-xml>`; a directive on a tag inside plain `<svg>` is found on the first pass. What hides it is `<script>` being a raw-text element, which is the row immediately above in the same provider. So these are the existing case with a wrapper that doesn't change the outcome, rather than a second route — and the new comments in both files now assert the opposite.

While looking at it I found the shape that genuinely isn't covered, which I'd rather not spell out here. Short version: the two-pass approach works because the sanitiser removes the element that did the hiding, and that isn't true for every such element — for the ones it keeps, both passes miss. Sent the detail and a suggested fix through the usual channel; happy to open a follow-up PR if that's easier.

obenland:
@bor0 Please feel free to just fix it

bor0:
Done — #777.

It turned out the cleanest fix was to stop reading the content as markup at all. The rule is that a pattern may not carry `data-wp-` anywhere, and that is a substring test; the helper already ran exactly that as its fast-path negative, so promoting it to the decision covers every hiding place and drops the sanitise-and-tokenize pass. The shape that wasn't covered was an element whose contents are text *and* which the sanitiser keeps, so neither pass ever reached the tag.

Your rows are kept as regression coverage, with the explanations corrected. CI is green.

One thing worth your call in the PR: it widens the rule, so prose that merely mentions `data-wp-` is refused too. Zero of the 600 published patterns I sampled and zero of the 1,858 seed patterns carry the marker in the fields this reads, so it looks free — but there's a provider row pinning that case, and it's easy to drop if you'd rather leave prose alone.

PR #775, opened 2026-09-16
Pattern directory: validate the blocks that will be stored and make preview mode one decision

mcliwanow:
## Why

Three correctness problems in how a submitted pattern is validated and how a pattern preview is resolved.

**Validation and storage could disagree about where blocks begin and end.** The REST validators parse `post_content` as it was submitted, but `wp_insert_post()` passes it through the save-time filters first, and those rewrite markup. Content can therefore be stored as a different set of blocks from the set the validators checked, so the block allow-list, the nesting check and the block counts were all applied to a representation that is not the saved one.

**Preview mode was decided two different ways.** The theme switch at `setup_theme` matched `#[?&]view=[1|true]#`. That bracket expression is a character class: it accepts `view=1` but also `view=t`, `view=u`, and misses `view=2`. The template selector and the theme's query tweaks, meanwhile, only checked whether the `view` query var was set, which `?view=`, `?view=2` and `/view/anything/` all do. A request could get the preview template while keeping the ordinary directory theme instead of the plain preview theme.

**The "Revert to draft" control changed state over GET.** It was a link built by a shortcode (`action=draft` plus a nonce), so anything that made the browser fetch that URL ran the status change.

## What changed

- **Content that would change on save is refused.** Two `rest_pre_insert` filters ahead of the validators. The first refuses a submission with a control character in a rendered field (`rest_pattern_control_characters`); no legitimate pattern contains one and the save filters delete them. The second computes the bytes `wp_insert_post()` will store and refuses the submission if that parses to a different block tree from the submitted bytes, in names, order or nesting (`rest_pattern_unstable_blocks`). Attributes are not compared: the save filters entity-normalise them, which would refuse ordinary content, and they only ever remove from a value. The validators keep parsing the submitted bytes, which are now known to describe the same blocks as the row. Checked against 300 live patterns from the directory API and the repo's seed exports (2,158 in all): no false positives with the tree comparison either.
- **One predicate decides preview mode.** `is_preview_request()` accepts the pretty `/view/` endpoint and `?view=1` or `?view=true`, nothing else. The theme switch asks it, and a `request` filter drops the `view` query var whenever it says no. Visible effect: `?view=2` and other non-`1` values now render the ordinary single template instead of the preview canvas; the site's own preview links (`?view=1`) are unaffected. The endpoint is matched against the URL path rather than the whole request URI, so `/view/?foo=bar` is a preview too.
- **The draft action is a POST.** The `[pattern_draft_link]` shortcode is removed and replaced by a server-rendered `wporg/draft-button` block that emits a form carrying the nonce, for a pattern its viewer can edit. `do_pattern_actions()` now returns unless the request is a POST, and reads `$_POST` for both the draft and report actions. Removing the shortcode is deliberate: it is what stops the status change being reachable as a link, so it should not be restored as a template tag.

## Testing

- Full PHPUnit suite green. Added coverage: submissions the save filters would rewrite are refused, with the error each case gets; the draft action does nothing on GET; the draft button renders a nonce-bearing POST form; and the preview predicate across the URL forms (`view=1`, `view=true`, the pretty endpoint, `view=2`, `view=t`, `view=`, an endpoint value).
- Checked over HTTP against this branch and against trunk: `?view=2` renders the preview canvas on trunk and the ordinary single template on the branch, while `?view=1` renders the preview on both. In a browser, as a pattern's author, the draft button submits and drafts the pattern. The existing valid-content fixtures (nested groups, a `mailto:` button, a query loop) are unaffected.


bor0:
Took a careful pass over this. The preview predicate and the draft-over-POST change both look right to me — I tried to fault them and couldn't.

The first change I'd hold on. Its stated premise is that once this one class of byte is refused, the validators and the stored row parse the same content. That premise doesn't quite hold: there's at least one other save-time transform that can move where the block boundaries land, and it involves none of the refused bytes. So the check is narrower than the invariant the docblock claims for it, and `validate_content()`'s new comment overstates what it can rely on.

Rather than widening the refusal, I'd suggest validating the stored representation directly — compute the content the save filters will actually produce and run the parse-based checks over that as well as over the raw bytes. Then the property holds regardless of which transforms we've thought of, and the raw parse is still available for the checks that need un-stripped text.

Sent the detail through the usual channel.

mcliwanow:
Pushed a new version, still one commit (bffdbbe).

The content check now compares the blocks the submitted markup parses to with the blocks the stored markup parses to, and refuses the submission when they differ. Ran it over 2,158 patterns (300 from the API, the rest from the seed exports) with no false positives. Dropped the canary test since the comparison covers it. Also picked up the smaller items: raw URI in is_preview_request(), endpoint match back to /view/ with the slash, post type check in the draft button. I'd appreciate another look 🙏

bor0:
Much better — comparing the two parses instead of listing the transforms is the right shape, and it holds up against everything I threw at it.

One thing left, and it's small: `block_names()` flattens the tree, so the comparison sees names in document order but not nesting. A rewrite that only re-parents blocks — same names, same order, different depth — compares equal and gets through, which leaves `validate_block_context()` checking a tree the row doesn't hold.

Suggest comparing the shape instead:

```php
function block_shape( $blocks ) {
	$shape = array();
	foreach ( $blocks as $block ) {
		if ( null === $block['blockName'] ) {
			continue;
		}
		$shape[] = array( $block['blockName'], block_shape( $block['innerBlocks'] ) );
	}
	return $shape;
}
```

`!==` on the nested arrays is recursive and order-sensitive, so it subsumes what you have now. I ran both over the 1,858 patterns in `.wp-env/data/exports/` and neither refuses any of them, so it looks free. Worth one more row in `data_content_rewritten_on_save()` for that case.

I'd stop at shape and not extend the comparison to attributes — I tried it and it refuses ordinary content (an image `alt` containing `&` normalises to `&amp;` on save), and it isn't needed.

Two non-blocking notes: `sanitize_post_field( …, 0, 'db' )` is correct because the `db` branch never reads the post ID, but that's worth half a sentence so the next reader doesn't have to go check; and `content_save_pre` now runs twice per insert, which is fine for core's callbacks but isn't obvious from the call site.

The preview predicate reading the raw URI and the endpoint match against the path are both better than what I suggested. Rest sent through the usual channel.

mcliwanow:
Hey @bor0 , thanks for the review. This is ready for another look.

bor0:
This looks good to me — no objection to merging.

`block_shape()` is a strict superset of what it replaces (equal shapes imply equal flattened name lists), so nothing the old comparison caught is lost, and it refuses the re-parenting case along with every variation of it I could build: the same trick one level deeper, a fabricated closer immediately followed by a fabricated opener, and an extra dash on a self-closing delimiter. Re-ran the seed corpus against the shipped code — 0 of 1,858 refused, so the tightening is free.

I also checked the two things the new docblock says it deliberately skips, and the reasoning holds: block attributes come through the save filters byte-identical (including the entity-encoded forms, which `attribute_has_unsafe_scheme()` decodes before its scheme check), so what the attribute validators saw is what gets stored. Good call putting all three of those notes in the docblock rather than leaving them for the next reader to rediscover.

One cosmetic thing, no action needed: the control-character check uses all of `RENDERED_FIELDS` while this one is `post_content` only. That's correct — the excerpt isn't block-parsed anywhere and the save filters only remove from it — but it may read as an oversight later, so it might be worth half a sentence.

CI is green on `4e1e7f9`.

bor0 (approved):

PR #776, opened 2026-09-16
Pattern translations: store translated patterns with the bytes the job assembled

mcliwanow:
## Why

`create_or_update_translated_pattern()` hands `wp_insert_post()` an unslashed array. `wp_insert_post()` expects slashed input and strips one level of backslashes from every field before it writes, `meta_input` included, so a translation is stored one backslash short. A translator's literal backslash goes missing, and so does the backslash in the `--` the block serialiser writes for `--` inside attribute JSON, which turns `var(--wp--preset--color--x)` into `var(u002du002dwpu002du002d...)`.

It is visible on production today. Reading `api.wordpress.org/patterns/1.0/?locale=<L>` for 15 locales, 500 patterns each (2026-09-16):

| locale | patterns with a stripped `\uXXXX` escape |
|---|---|
| en_US | 0 of 500 |
| de_DE, fr_FR, es_ES, it_IT, nl_NL, ru_RU, pl_PL, ko_KR | 6 of 500 each |
| tr_TR, ja, ar, pt_BR | 5, 4, 3, 2 |
| all non-en_US | 68 of 7000 |

Same handful of parents in every locale (a gradient and a spacer height with `--` in them, and one `"Read More u0026gt;"`). `decode_pattern_content()` in the directory plugin already has a workaround for the `u0026amp;` case ("Sometimes the initial `\` is missing"), which is this bug seen from the other end. The count undercounts because that function repairs the one it knows about before the API returns.

## What changed

- The insert is `wp_insert_post( wp_slash( $args ), true )`. Everything in `$args` is unslashed on arrival (GlotPress strings, `get_post()` reads) and nothing is pre-escaped, so slashing the array once is the whole fix. `wp_slash()` only touches strings, so the boolean and the parent's array-valued meta pass through as before.
- The four `meta_input` lines that read from the parent post fall back to `''` like the post fields above them, instead of reading a property off `false` when there is no parent. No change when a parent exists.
- `PatternParser::decode_unicode_characters()` leaves `\u005c` encoded. It used to decode it to a bare backslash, which starts a new escape inside the attribute JSON, so a placeholder containing `C:\Temp` parsed as no attributes at all and was dropped on save. It also leaves alone any escape that decodes to a byte KSES deletes on save (`[\x00-\x08\x0B\x0C\x0E-\x1F]`), so the markup it hands on is the markup that ends up stored.
- `is_translated_content_allowed()` checks the bytes the row will hold (`sanitize_post_field( ..., 'db' )` of the assembled markup, the job runs without a user so the save-time filters apply) as well as the assembled form.
- Tests: `Pattern_Translation_Storage_Test` runs the parser and the insert with a backslash in the title, description, keywords and a heading placeholder and checks each stored field, plus one for a typed control-byte escape; the existing translated-content test gets a row for a raw control byte in a delimiter.

Existing rows don't need a migration. The twice-daily job rewrites every translation from its parent and the current strings, so they come right on the first run after deploy.

## Testing

- The three new cases with trunk's copies of `pattern-translations.php` and `class-patternparser.php` swapped in: all three fail (the placeholder comes back without its backslash, then without the attribute at all). With this branch they pass.
- Full PHPUnit suite on the branch: 263 tests, 692 assertions, green. `phpcs` clean.
- Legitimate path: the round trip is the test, and the parent's `wpop_viewport_width` still lands as `1200`.

Post-deploy check, after the next `pattern_import_translations_to_directory` run:

```
curl -sL 'https://api.wordpress.org/patterns/1.0/?locale=de_DE&per_page=100&page=1' | grep -o 'u002du002d' | wc -l
```

Expect 0. It was 6 patterns' worth on 2026-09-16.


bor0:
Reproduced the round trip from first principles and this is right — `wp_insert_post()` unslashes post fields and `meta_input` alike, so the unslashed call loses a backslash from content, title and meta, and one `wp_slash( $args )` at the insert covers all three. The stored form I get on trunk matches the production symptom in the description exactly (`var(u002du002dwpu002du002dpreset)`). `create_or_update_translated_pattern()` holds the plugin's only `wp_insert_post()`, and `cron.php` re-runs every pattern × locale unconditionally, so the "no migration needed" note checks out.

The parser changes look right too. Leaving `\` encoded is necessary — decoding it to a bare backslash opens a new escape in the attribute JSON, and I confirmed the attribute is dropped on trunk and survives here. Refusing to decode an escape whose value the save filters would delete is the right shape: it stops the producer emitting something that save will change, rather than trying to detect it afterwards. Nice touch that the two changes are independently sufficient — I checked each one with the other reverted and either alone is enough.

Two things, neither blocking:

**The new comment isn't quite accurate.** `content_save_pre` carries five callbacks, not one:

```
8   wp_strip_custom_css_from_blocks
9   wp_filter_global_styles_post
10  convert_invalid_entities
10  wp_filter_post_kses
50  balanceTags
```

so the row doesn't hold `wp_kses_post()` output — a block with a `style.css` attribute is stored without it, which `wp_kses_post()` keeps. It makes no difference to what this function checks (block names and shape came out identical in every case I tried, and all four of the others only remove), but since #775 settled on `wp_unslash( sanitize_post_field( 'post_content', wp_slash( $html ), 0, 'db' ) )` for "the bytes that will be stored", using the same call here would make the comment true and give both paths one idiom.

**A note for anyone reproducing this by hand:** WP-CLI doesn't attach the KSES filters, so a probe run under `wp eval` without `wp_set_current_user( 0 ); kses_init();` tests a save path with no KSES in it and everything looks stable. Cost me a round of confusing results. Might be worth a line in the testing steps.

CI is green on `e2f1d13`. Rest sent through the usual channel.

bor0 (approved):
Approving.

I reproduced the round trip against core independently of the test suite: `wp_insert_post()` unslashes post fields and `meta_input` alike, the unslashed call loses a backslash from content, title and meta, and one `wp_slash( $args )` covers all three. The stored form I get on trunk matches the production symptom in the description. `create_or_update_translated_pattern()` holds the plugin's only `wp_insert_post()`, and the cron re-runs every pattern × locale unconditionally, so existing rows heal on the first run as described.

The parser changes are right for the reasons in the description, and I checked each of the two with the other reverted — either alone is sufficient, which is a good property to have here.

Neither of my earlier notes blocks this: the `content_save_pre` comment is a comment, and the behaviour it describes has no effect on what the function actually checks. Still worth switching to `wp_unslash( sanitize_post_field( 'post_content', wp_slash( $html ), 0, 'db' ) )` so both paths share one idiom for "the bytes that will be stored" — happy for that to be an amend here or a follow-up, whichever is less disruptive.

One thing I could not do: my local env won't build (`npm install` fails on `@wordpress/primitives@4.50.0`), so I never ran the suite, and the trunk-swap control has only been run by you. CI is green and I verified the underlying behaviour by hand, so I'm satisfied — but that control is the part no one independent has reproduced, and it's worth a second pair of hands if anyone has a working checkout.

mcliwanow:
> Neither of my earlier notes blocks this: the content_save_pre comment is a comment, and the behaviour it describes has no effect on what the function actually checks. Still worth switching to wp_unslash( sanitize_post_field( 'post_content', wp_slash( $html ), 0, 'db' ) ) so both paths share one idiom for "the bytes that will be stored" — happy for that to be an amend here or a follow-up, whichever is less disruptive.

Already in as of f373c18 (line 54), which is the commit you approved. Thanks for the review.

